PR #13, opened 2024-12-17
Feat/teams component

### LINKED ISSUE #8: Create "team" page
We want to create a page that consumes the VMS API and display a team within Toronto JS.

The design should be based on: https://torontojs.com/volunteers

manvinderjit:
Add Teams Component to show TorontoJS' teams.
Add TeamMember Component to show team members.
Updated index.css file to fix open props import and variables.

Resolves: #8


public/members.json
```diff
@@ -0,0 +1,105 @@
+[
+  {
+    "id": "1",
+    "email": "john.doe@example.com",
+    "name": "John Doe",
+    "description": "A passionate volunteer.",
+    "happenedAt": "2024-12-09T12:00:00",
+    "insertedAt": "2024-12-01T10:00:00",
+    "avatar": "https://randomuser.me/api/portraits/men/1.jpg",
```
madcampos, line 5: **nitpick:** being super absurdly picky here, can we have some variation on the data, specially the avatars? 🙃

what about using [faker](https://fakerjs.dev/) to generate us some data?‏

public/teams.json
```diff
@@ -0,0 +1,34 @@
...
+        "id": "884b92f1-992b-46fd-a3ba-54251fe85daa",
+        "name": "Community Crafters",
+        "description": "Toronto JS' tournament wizards!",
+        "happenedAt": "2024-12-09T12:00:00",
+        "insertedAt": "2024-12-01T10:00:00",
+        "avatar": "https://randomuser.me/api/portraits/lego/5.jpg"
+    },
+    {
+        "id": "ae2379dc-f215-4c79-a942-45ccb308312a",
+        "name": "Omnipotent Organisers",
```
madcampos, line 28: **praise:** Love the team names! 💖‏

src/components/TeamMember/TeamMember.jsx
madcampos, line 1: **suggestion:** ‏I think it would be better to call this component "team member card" or "member card". It makes the name more explicit on what it is and how it is used.

The reasoning is: imagine you open a link to this file on github, without any other context, is it easy to figure out what this component do or how it relate to the application as a whole?

src/components/TeamMember/TeamMember.jsx
```diff
@@ -0,0 +1,21 @@
...
+            src={member.avatar || "/default-avatar.png"}
+            alt={`Team ${member.name} Avatar`}
+          />
+          <div className="team-member-card-text">
+            <h3 className="team-member-card-heading">{member.name}</h3>
+            <span>linkedin</span>
+          </div>
+        </div>
+      </a>
+    </div>
```
madcampos:
```suggestion
    <article className="team-member-card">
      <a
        href={member.socialLinks.linkedin}
        target="_blank"
        className="team-member-profile"
      >
        <picture>
          <img
            className="avatar"
            src={member.avatar || "/default-avatar.png"}
            alt={`Team ${member.name} Avatar`}
          />
        </picture>
        <header className="team-member-card-text">
          <h3 className="team-member-card-heading">{member.name}</h3>
        </header>
        <div className="social-media-links">
          <ul aria-label="Social Media Links">
            <li><a>linkedin</a></li>
          </ul>
        </div>
      </a>
    </article>
```

**Explanation:**

- Use semantic tags, those relate to [landmark roles](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/) and make navigation easier.
  - Specifically, `<article>` here means "an item on a list of items". It is not necessarily a "news _article_", it's meaning is more like an "_article_ of clothing" (i.e. an item on a shopping website). That makes the tag a good candidate for a "card".
  - `<header>` is also a landmark role and helps with organizing content inside of an `<article>`. Other tags that help are `<footer>` and `<aside>`.
- The `<div>` for social media links is just to help partition content, it has no special meaning and doesn't affect the overall structure, so it is okay to use a `<div>` here
- The list however, _does_ have a meaning.
  - Here we want to add a list of links, to having a `<ul>` conveys that.
  - When a screen reader user navigates to the list, the `aria-label` acts as an extra [affordance](https://en.wikipedia.org/wiki/Affordance) to help contextualize the user where they are in the page and what the expected content is.
  - For sighted users, the affordance[^1] here will be icons and text for the social media links.
- Lastly, wrapping the `` in a `<picture>` element helps with styling the image and giving a container to do whatever we want without removing semantic meaning.

Here is an extra reference, it doesn't include the landmark elements, but is a good read on cards and accessibility, albeit a little bit old: https://inclusive-components.design/cards/

[^1]: Cool video about UX but somewhat unrelated to this: https://youtu.be/yY96hTb8WgI?si=dELfNYcuG7GaQKgt

src/App.jsx
madcampos, line 1: **suggestion:** Let's use this PR as a base: https://github.com/torontojs/volunteer/pull/12

It changes a little bit the structure to add two different pages as separate "apps", with many quotes here. One for profile and other for team.

That way we can build them as separate pages and not worry about react router or other shenanigans like that, making the app closer to a multi-page application than a single page application‏

src/App.css
madcampos, line 1: **nitpick:** I'd rather we have a separate css file for each component with all the styles relative to it there.

To make things easier to load, we can add a `<link>` tag for each component css or have a `main.css` file that `@import` all of the components styles.

src/index.css
```diff
@@ -1,6 +1,10 @@
-@import url('https://cdn.jsdelivr.net/npm/open-props@1.0.0/open-props.min.css');
+@import "open-props/style";
+@import "open-props/open-props.min.css";
```
madcampos:
```suggestion
@import "open-props/open-props.min.css";
```

This is a redundant import, let's use the direct css file one.

src/components/Teams/Teams.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+const Teams = () => {
+  const [isLoadedTeamsData, setIsLoadedTeamsData] = useState(false);
+  const [teamsData, setTeamsData] = useState([]);
+  const [teamMembersData, setTeamMembersData] = useState([]);
+  const [teamsError, setTeamsError] = useState(null);
+
+  useEffect(() => {
+  
+    //TODO: Add abort controller when fetching from API
```
madcampos: **question:** Why would we need an abort controller here?‏

src/components/Teams/Teams.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+    setIsLoadedTeamsData(false);
+    // Fetch data from the JSON file in the public directory
+    async function fetchData () {
+      try {
+        const [responseTeams, responseMembers] =  await Promise.all([fetch("/teams.json"), fetch("/members.json")]);
```
madcampos: **suggestion:** Separate the two requests so they are independent.

We should first load the team data and then only if that loads we load the members data. It is okay to display only the teams information and not the members.

src/components/Teams/Teams.jsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+    fetchData();
+    
+    //TODO: Add cleanup function for abort controller when fetching from API
+    
+  }, []);
+
+  // If data not yet loaded
+  if (!isLoadedTeamsData) {
+    return <div>Loading...</div>;
```
madcampos:
```suggestion
    return <div aria-live="polite" role="status">Loading teams...</div>;
```

Here is the reference for [ARIA Live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).

The TLDR:
- Live region will announce what is going on to screen readers, we use the `polite` value here as it will not interrupt what the screen reader is reading. As the message about loading data is not urgent it is better to wait and not interrupt things.
- The `role="status"` identifies this `<div>` as a place where the status of the page is present.
- The added text helps give context of that is happening.

src/components/Teams/Teams.jsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  // If error encountered
+  if (teamsError) {
+    return <div>An error occurred!</div>;
```
madcampos:
```suggestion
    return <div>Unable to load teams. Please try refreshing the page.</div>;
```

As a UX improvement we give a more detailed error message, explaining what happened and also provide a potential fix for it.

[This heuristic](https://www.nngroup.com/articles/ten-usability-heuristics/#toc-9-help-users-recognize-diagnose-and-recover-from-errors-9) for UX design in specific should be present here.

src/components/Teams/Teams.jsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  // If team data is loaded but empty data returned
+  if (isLoadedTeamsData && !teamsError && (!teamsData || teamsData.length === 0)) {
+    return <div>No data found!</div>;
```
madcampos: Again here: provide a better error message and how to recover from that.

src/components/Teams/Teams.jsx
```diff
@@ -0,0 +1,85 @@
...
+                )}                
+                {teamMembersData.length > 0 &&
+                  teamMembersData.map((member, index) => (
+                    <TeamMember key={index} member={member} />
+                  ))}
+              </div>
+            </div>
+          );
+        })}
+      </section>
```
madcampos:
```suggestion
      <main>
        {teamsData.map((team, index) => {
          return (
            <article key={index} className="team">
              <header className="team-header">
                <h2>{team.name}</h2>
              </header>

              <div className="team-body">
                <p>{team.description}</p>
         
                {(!teamMembersData || teamMembersData.length <= 0) && (
                  <div>This team has no active members.</div>
                )}                
                {teamMembersData.length > 0 && (
                  <h3 id="team-members-label-{index}">Team members</h3>
                  <ul aria-labeled-by="team-members-label-{index}">
                  {teamMembersData.map((member, index) => (
                    <li><TeamMember key={index} member={member} /></li>
                  ))}
                  </ul>
                )}
              </div>
            </article>
          );
        })}
      </main>
```

Explanation:

- `<main>` represents the main content on the page, if we are rendering only this component, it should render as the main element. Please note there should only be **ONE** `<main>` element on each page.
- The team description can be part of the team body.
- The team members heading helps with content structure.
- The team members should be a list, it also has a label defined using the `aria-labeledby` attribute. The attribute has to reference an `id`, because of that, the id is constructed using the index of each team.
- Again, `<header>` and `<article>` are used as landmarks.

src/App.css
```diff
@@ -42,3 +42,127 @@
...
+  width: 0;
+  height: 1px;
+  background-color:var(--color-grey);  
+  transform: translateX(-50%);
+}
+
+.team-member-card:hover span::after {
+  width: 100%;
+  left: 50%;
+}
```
madcampos: **suggestion:** Transitioning `width` and `height` force the browser to do expensive layout and content flow calculations‏, even if the element doesn't affect the parent.

As the intended effect is to change the background color, using a `linear-gradient` and changing the `background-size` property will work, as well as the `transform` property with a `scale` value.

Those properties will only affect the paint process and are GPU accelerated.

Here is a reference of which property triggers what: https://css-triggers.com/

In order of cheap to expensive: paint > layout > composite

src/components/Teams/Teams.jsx
```diff
@@ -0,0 +1,85 @@
...
+                <p>{team.description}</p>
+              </div>
+
+              <div className="team-body">              
+                {(!teamMembersData || teamMembersData.length <= 0) && (
+                  <div>No Team Members Found!</div>
+                )}                
+                {teamMembersData.length > 0 &&
+                  teamMembersData.map((member, index) => (
+                    <TeamMember key={index} member={member} />
```
madcampos: **thought:** I don't think we want to load the team members on the team list page... It makes more sense to have a page listing teams have only the names and a short summary of the team, not much information there, and if the person want to see more, they go to the team page itself where all information is listed.

@kbventures @marianavrocha thought?

madcampos (changes requested):

src/components/Teams/Teams.jsx
manvinderjit: In case we want to abort the request to API if the component reloads without the previous request being fulfilled. It is not a necessity though.

src/App.jsx
manvinderjit: Do you want me to clone it right away or wait for the pull request to be merged?

src/components/TeamMember/TeamMember.jsx
manvinderjit: For now, I was following the semantics on the current volunteer page and the w3c design, but will adapt these changes from now on.

src/App.jsx
madcampos: Let's merge [that PR](https://github.com/torontojs/volunteer/pull/12) first and then you can rebase/merge on top of this.

I'm waiting for @kbventures review, but please feel free to also chime in on the PR

src/components/Teams/Teams.jsx
madcampos: You mean so re-rendering the component doesn't request the data twice?

To solve this problem, you can have two bits of information:
1. A flag telling if a request is in progress.
2. The data itself, it can be an object or nullish.

Then the `useEffect` can look like this:
1. Check if there is a request in progress and just `return`.
2. Check if the data exists (i.e. is not `null`/`undefined`) and just `return`.
3. Add the logic for the data request itself, taking care to assign both the data and the flag.

This way even if the component re-renders it will not send multiple requests. 😉

public/members.json
manvinderjit: Resolved: Added distinct names and avatars images for team members

src/App.css
manvinderjit: Separated css files for components

manvinderjit:
@madcampos Resolved all except the following:
1.  https://github.com/torontojs/volunteer/pull/13#discussion_r1888851747
Waiting on PR merge for https://github.com/torontojs/volunteer/pull/12 
2. [Flags for reload of useEffect](https://github.com/torontojs/volunteer/pull/13#discussion_r1888869002)
Will add when fetching data from API
3. Separating Team and Team Members pages https://github.com/torontojs/volunteer/pull/13#discussion_r1888926676
Waiting on confirmation

src/components/Profile/Profile.jsx
madcampos, line 1: Here we can use a similar structure as the team to create a card using `<article>` and friends.

The same patterns should be used.

manvinderjit: It was a component created by Ken when he setup the project. I removed it from the component tree in App.js for the moment but didn't delete the file. I don't know if it is to be included or not.

madcampos:
@manvinderjit I've merged the tooling PR, so you can merge `dev` into this branch.

If you would like a quick (15 ~ 30 min) call so we can go over the potential merge conflicts together, feel free to hit me up on slack and we can do this on the fly.

As for the flags on the `useEffect`, it is okay to do it differently. Not a big issue with me if you choose to do this another way.

And for separating teams list and individual team pages, we can do that as an iteration on this.

manvinderjit:
I will give merging a shot and in case I have any issues, I will ping on slack.

src/components/Profile/Profile.jsx
madcampos: I see, I thought it was the component used for displaying profiles.

We do want to display profiles and use the semantic html pattern for it. But if it is using this component or not, it is up to you 😉

manvinderjit:
@madcampos I have merged the two branches and pushed. Howver, there seems to be a CORS issue with the deployment. Please do look into it.

madcampos (approved):

madcampos (approved):

madcampos (approved):

PR #17, opened 2025-02-11
feat: Logo, HeaderTitle components

sounmind:


public/torontojs-logo.png
madcampos, line 1: **nitpick:** Please use the SVG logo instead: https://github.com/torontojs/tojs-styleguide/blob/main/Old/public/assets/logo.svg

Vector graphics look sharper than raster and are usually smaller. Our logo has relative simple shapes and colors, making it a very good candidate for vector graphics.‏

src/components/Logo/Logo.css
```diff
@@ -0,0 +1,4 @@
+.logo img {
+	height: 128px;
+	width: 128px;
```
madcampos, line 3: **praise:** Please use [relative units instead of absolute ones](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#lengths), they are less responsive!

Here is a video with one specific example: https://www.youtube.com/watch?v=xCSw6bPXZks

public/torontojs-logo.png
sounmind: Done it!

src/components/Logo/Logo.css
sounmind: Thanks, done it!

src/components/Logo/Logo.css
```diff
@@ -0,0 +1,14 @@
...
+}
+
+.sr-only:not(:focus):not(:active) {
+	clip: rect(0 0 0 0);
+	clip-path: inset(50%);
+	height: 1px;
+	overflow: hidden;
+	position: absolute;
+	white-space: nowrap;
+	width: 1px;
```
madcampos:
```suggestion
	clip: rect(0 0 0 0);
	clip-path: inset(50%);
	height: 1px;
	overflow: hidden;
	position: absolute;
	white-space: nowrap;
	width: 1px;
	color: transparent;
	pointer-events: none;
	user-select: none;
```

Those css properties actually disable any accidental clicks and text selection, as well as make the text transparent so there is no annoying pixel that may show up 😅

src/components/Logo/Logo.tsx
```diff
@@ -0,0 +1,99 @@
...
+					/>
+					<path
+						className='fill-fg-color'
+						d='M248.4,564.3c0,-0 41.5,1.5 66.1,-24.6c22,-22 19,-51.8 20,-68.7l0,-189.1l-55.9,-0l0,200.9c0,-0 1.5,32.3 -30.8,32.3c-29.8,-0 -38.4,-30.8 -38.4,-30.8l-45.6,28.2c0,0.1 19.5,50.8 84.6,51.8Z'
+						fill='#fff'
+					/>
+				</g>
+			</g>
+		</svg>
+		<span className='sr-only'>TorontoJS</span>
```
madcampos: Two things here:
1. It is better to use another pattern for svg names
2. Maybe we don't need a picture wrapper

### Pattern for svg names
Try this: https://github.com/madcampos/madcampos/blob/main/src/components/Logo/index.astro#L38-L41

I was [reading a while back](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/) that the best combo for screen readers to read an svg name is adding a `<title>` element and the `aria-describedby` attribute to the `<svg>` element itself.

We should still keep the `.sr-only` CSS class, but use it for other things ;)

### Picture wrapper

If the logo will be used standalone, for example, to represent an image in the middle of the text, then yes, we should add a `<picture>` element as a wrapper.

But if the logo will be used inside links, for example, as the first link on a navigation bar, then we should not add the `<picture>` wrapper.

My suggestion here would be to remove it from the logo element, and let the code using it deal with adding a wrapper or not. That way things are more flexible and simpler for us.

src/components/Logo/Logo.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const Logo = () => (
+	<picture className='logo'>
+		<svg
+			width='100%'
+			height='100%'
+			viewBox='0 0 613 613'
+			xmlns='http://www.w3.org/2000/svg'
+		>
+			<rect
```
sounmind: Memo

```suggestion
		<svg
			width='100%'
			height='100%'
			viewBox='0 0 613 613'
			xmlns='http://www.w3.org/2000/svg'
			aria-labeledby="id-for-title"
		>
		  <title id="id-for-title">svg title</title>
			<rect
```

madcampos (approved):
Approving the PR with one caveat:

We have a TODO comment to review in the future

PR #18, opened 2025-02-15
style: add global css

sounmind:

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -1,0 +1,153 @@
+:root {
+	--font-primary: 'Uai', 'Source Sans Pro', 'Montserrat', 'Roboto', Helvetica, Arial, sans-serif;
+	--text-base: 18px;
+	--text-scale-ratio: 1.067;
+	--spacing-base: 8px;
+	--rounded-base: 2px;
+	--leading-base: 1.6em;
+	--shadow-base: 2px;
+	--leading-h6: calc(var(--leading-base) - 0 * 0.08em);
+	--leading-h5: calc(var(--leading-base) - 1 * 0.08em);
+	--leading-h4: calc(var(--leading-base) - 2 * 0.08em);
+	--leading-h3: calc(var(--leading-base) - 3 * 0.08em);
+	--leading-h2: calc(var(--leading-base) - 4 * 0.08em);
+	--leading-h1: calc(var(--leading-base) - 5 * 0.08em);
+	--text-xs: calc(1em * pow(var(--text-scale-ratio), -2)); /* e.g.: helpers */
+	--text-sm: calc(var(--text-base) * pow(var(--text-scale-ratio), -1)); /* e.g.: input values, buttons */
+	--text-md: calc(var(--text-base) * pow(var(--text-scale-ratio), 1)); /* e.g.: h6 */
+	--text-lg: calc(var(--text-base) * pow(var(--text-scale-ratio), 2)); /* e.g.: h5 */
+	--text-xl: calc(var(--text-base) * pow(var(--text-scale-ratio), 3)); /* e.g.: h4 */
+	--text-2xl: calc(var(--text-base) * pow(var(--text-scale-ratio), 4)); /* e.g.: h3 */
+	--text-3xl: calc(var(--text-base) * pow(var(--text-scale-ratio), 5)); /* e.g.: h2 */
+	--text-4xl: calc(var(--text-base) * pow(var(--text-scale-ratio), 6)); /* e.g.: h1 */
+	--weight-thin: 100;
+	--weight-extralight: 200;
+	--weight-light: 300;
+	--weight-regular: 400;
+	--weight-medium: 500;
+	--weight-semibold: 600;
+	--weight-bold: 700;
+	--weight-extrabold: 800;
+	--weight-black: 900;
+	--color-foreground: hsla(0, 0%, 20%, 1);
+	--color-background: hsla(0, 0%, 100%, 1);
+	--color-border: hsla(0, 0%, 80%, 1);
+	--color-card: hsla(0, 0%, 90%, 1);
+	--color-helper: hsla(0, 0%, 60%, 1);
+	--color-accent: hsla(2, 84%, 56%, 1);
+	--color-error: hsla(0, 100%, 50%, 1);
+	--color-warning: hsla(44, 98%, 44%, 1);
+	--color-success: hsla(120, 100%, 30%, 1);
+	--color-info: hsla(224, 100%, 60%, 1);
+	--color-white: hsla(0, 0%, 100%, 1);
+	--color-black: hsla(0, 0%, 20%, 1);
+	--color-button: hsla(0, 0%, 0%, 0.15);
+	--color-shadow-light: hsla(0, 0%, 0%, 0.1);
+	--color-shadow-heavy: hsla(0, 0%, 0%, 0.125);
+	--spacing-xs: calc(1 * var(--spacing-base));
+	--spacing-sm: calc(2 * var(--spacing-base));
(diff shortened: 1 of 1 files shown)


src/index.css
```diff
@@ -0,0 +1,151 @@
+:root {
+	--font-primary: 'Uai', 'Source Sans Pro', 'Montserrat', 'Roboto', Helvetica, Arial, sans-serif;
```
madcampos, line 2: **question:** ‏Did we get the font files from Cadu?

src/index.css
```diff
@@ -0,0 +1,151 @@
...
+	--text-base: 18px;
+	--text-scale-ratio: 1.067;
+	--spacing-base: 8px;
+	--rounded-base: 2px;
+	--leading-base: 1.6em;
+	--shadow-base: 2px;
```
madcampos, line 8: **nitpick:** ‏Can we make those values relative to the text size?

Like this:

```css
:root {
  /* Here we need a unitless number so we can do ✨ math ✨ */
  --REFERENCE_BASE_SIZE: 18;

  /* This is a decent assumption for a browser default font setting. */
  /* (i.e. "medium font size") */
  /* Also a unitless value for ✨ math reasons ✨ */
  --BROWSER_DEFAULT: 16;

  /* Here we get the ratio between the size we want and the assumed browser default. */
  /* Then multiply by 1rem to force the browser to calculate a value out of it, */
  /* instead of having a unitless number. */
  /* Now: 1rem = 18px, for the default font size configuration in all major browsers. */
  --text-base: calc((var(--REFERENCE_BASE_SIZE) / var(--BROWSER_DEFAULT)) * 1rem);

  /* This is the single value everything will be based off and be relative to. */
  /* We set the root font size so it cascades to everything else. */
  /* And rem units work. */
  font-size: var(--text-base);

  /* ✨ MOAR MATH ✨ */
  /* But now normalized to the new rem value */
  --spacing-base: calc((8 / var(--REFERENCE_BASE_SIZE)) * 1rem); /* 8px / 18px = 0.444...rem */
  --rounded-base: calc((2 / var(--REFERENCE_BASE_SIZE)) * 1rem); /* 2px / 18px = 0.111...rem */
  --shadow-base: calc((2 / var(--REFERENCE_BASE_SIZE)) * 1rem); /* Same as --rounded-base */

  /* NOTE: we just need to do this for the base values 😅 */
}
```

Here is a minimal demo to showcase this difference: https://codepen.io/madcampos/pen/MYWKYjN

Go to your browser settings and change the font size and see the font change on all paragraphs but the one with a fixed pixel size. (This should be under "Appearance" or something.)

Here is it the setting in Edge:
image

src/index.css
```diff
@@ -0,0 +1,151 @@
...
+	--spacing-lg: calc(5 * var(--spacing-base));
+	--spacing-xl: calc(10 * var(--spacing-base));
+	--spacing-xl: calc(10 * var(--spacing-base));
+	--rounded-xs: calc(1 * var(--rounded-base));
+	--rounded-sm: calc(2 * var(--rounded-base));
+	--rounded-md: calc(4 * var(--rounded-base));
+	--rounded-lg: calc(8 * var(--rounded-base));
+	--rounded-xl: calc(12 * var(--rounded-base));
+	--rounded-full: 50%;
+	--border: 1px solid var(--color-border);
```
madcampos: **suggestion:** ‏There should be a `--border-base` value for the border thickness

src/index.css
```diff
@@ -0,0 +1,151 @@
...
+
+html,
+body {
+	color: var(--color-foreground);
+	font-family: var(--font-primary);
+	font-size: var(--text-base);
+	line-height: var(--leading-base);
+}
+
+a { color: var(--color-accent); }
```
madcampos, line 78: **suggestion:** ‏Also add [`accent-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color) to the `:root` declaration.

src/index.css
```diff
@@ -0,0 +1,151 @@
...
+h3,
+h4,
+h5,
+h6,
+.text-h1,
+.text-h2,
+.text-h3,
+.text-h4,
+.text-h5,
+.text-h6,
```
madcampos, line 91: **nitpick:** It is more readable this way:
‏
```suggestion
h1, h2, h3, h4, h5, h6,
.text-h1, .text-h2, .text-h3, .text-h4, .text-h5, .text-h6,
```

"Old man yelling at cloud" gif

And please use `dprint` to format the code instead of effing prettier.

Prettier is a complete and utter garbage, it messes up code in unintuitive ways for AST sake and not humab sake, and I  will die on this hill!

Add these files/lines to the project please:
- https://github.com/torontojs/vms/blob/feat/api-docs/.vscode/settings.json
- https://github.com/torontojs/vms/blob/feat/api-docs/dprint.json
- https://github.com/torontojs/vms/blob/feat/api-docs/package.json#L11
- https://github.com/torontojs/vms/blob/feat/api-docs/package.json#L26

src/index.css
manvinderjit: @madcampos Yes

src/index.css
manvinderjit: @madcampos Tagging @caducarvalho

src/index.css
caducarvalho: @manvinderjit @madcampos Which font format are you using? I've shared the OTF font files with the design team, but they're not proper for web browsers. I've attached the WOFF2 font files, with a link below.

[Uai-Variable.zip](https://github.com/user-attachments/files/18875138/Uai-Variable.zip)

caducarvalho: @manvinderjit @madcampos I've also added a small CSS font-face snippet, you just need to place the files and import in the global CSS.

manvinderjit: @madcampos I am not sure if there is an issue with the font or was it just a comment. @caducarvalho  Thank you for the files, we will use these just in case.

madcampos: Was just a comment to check if you got the files. I didn't see a css file with `@font-face` declarations so was just double checking.

Thank you @caducarvalho for providing the files! 😎

src/index.css
sounmind: @madcampos Thanks for comment! We might skip this for now and think about it later.

src/index.css
sounmind:
```css
:root {
    // ...
	accent-color: var(--color-accent);
}
```

Like this?
Can you tell me why we need this?

src/index.css
sounmind: It's weird. We already have dprint settings and its script(`npm run format` on pre-commit) is working well. I think there's no css config in dprint settings. Can you check this again? @madcampos

src/index.css
madcampos: Yes, exactly like that.

It is a recent addition to CSS that allows more control on the color of some things like checkboxes, progress bars, radio buttons, and range sliders. Basically it gives some customization for form controls without being too complex.

Along with that we can also set the scrollbar colors: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color

src/index.css
madcampos: Dprint [is configured to format CSS](https://github.com/torontojs/volunteer/blob/dev/dprint.json#L71-L86), but it is not annoying at that, like, it won't reformat this specific way of writing selectors.

src/index.css
sounmind: Cool! I didn't know we can even customize these parts. Thanks!

sounmind:
@madcampos Marco, are we okay to merge this PR?

madcampos (approved):
LGTM

PR #19, opened 2025-02-15
feat: Footer component

sounmind:


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+		<ul>
+			<li>
+				<a
+					target='_blank'
+					aria-label='External source: TorontoJS Website'
+					href='https://torontojs.com/'
+				>
+					<img src='/home.svg' alt='Home' />
+				</a>
```
manvinderjit: @madcampos , @sounmind and me wanted to confirm if we need to include aria label if we have text in the img tag. We have included it for now but thought we would confirm before removing it.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,57 @@
+import './Footer.css';
+
+const Footer = () => (
+	<footer>
+		<p className='text-helper'>TorontoJS</p>
+
+		<ul>
```
manvinderjit: @madcampos Do we need an aria-label for the `ul` tag?

madcampos: No. It is sufficient to have it use the default role here.

src/components/Footer/Footer.tsx
madcampos: No, we don't need it. But I would suggest adding the text as an alternative to the icons.

Something like this:
```html
<li>
  <a
    href="..."
    (1)
    rel="noopener noreferrer" (2)
  >
    <img src="..." alt="" /> (3)
    <span>YouTube</span> (4)
  </a>
</li>
```

Here is an explanation of the numbers:

1. Adding the `target` attribute can be seen as a violation of user control and freedom, so there are mixed feelings about adding it or not. I prefer not to as things don't break when users navigate away from the page, like, if they go back, they are back at the same page and haven't lost a some work.
2. Adding this `rel` attribute helps with privacy to avoid pesky trackers. It specifically tells the browser to **not** ser the [`Referer` header](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) and [`window.opener`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener) on the next page.
3. See this advice for keeping the `alt` attribute empty: https://html5accessibility.com/stuff/2024/11/23/old-alt-text-advice/#icon-images (I've learned about this recently as well)
4. Adding a text description in the html helps if the image doesn't load, if the icon is too small, and with automatic translation (`aria-*` attributes don't usually get translated). And if we want to hide it anyways, we can always use a [`.visually-hidden`](https://css-tricks.com/inclusively-hidden/) helper class to do so.

src/components/Footer/Footer.css
```diff
@@ -0,0 +1,25 @@
+footer {
+	align-items: center;
+	column-gap: 16px;
```
madcampos: **nitpick:** Please, use relative css units (like rems) instead of absolute ones (like pixels).‏

src/components/Footer/Footer.tsx
madcampos: **suggestion:** We should wrap this list in a `nav` element, as this would be a secondary navigation.

Like this:
```html
<nav aria-label="Secondary Navigation">
  ...
</nav>
```

Here using an `aria-label` is nice as it gives the user a way of knowing what navigation this is.

madcampos:
Forgot to add this to the review, but a further suggestion is: inlining the icon svgs in the html, so they can inherit the color and style of the sorrunding text.

Take a look here: https://iconify.design/

src/components/Footer/Footer.tsx
manvinderjit: @madcampos @sounmind Do we want to open social media links in the same window? Do you think a better experience for users will be if social media websites open in a new tab or in the same tab? I get that all the links for the vms itself should be in the same tab but not sure on the social links.

madcampos: > Do we want to open social media links in the same window? Do you think a better experience for users will be if social media websites open in a new tab or in the same tab? I get that all the links for the vms itself should be in the same tab but not sure on the social links.

So yeah, it is contentious... I prefer not to add the `target` attribute _personally_. But I also won't die on that hill, so do as you wish.

sounmind:
@madcampos Thank you for the comments, @manvinderjit and I applied your feedback to [this change](https://github.com/torontojs/volunteer/compare/157c20a5bc4b60f7b987b3c45303f9a1cad250a0..58006733f289a51706495ad18d4b9f93d28490b2). Can you review it?

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,179 @@
...
+								/>
+							</g>
+							<defs>
+								<clipPath id='clip0_468_1665'>
+									<rect width='16' height='16' fill='white' />
+								</clipPath>
+							</defs>
+						</svg>
+
+						<span id='TorontoJS' className='sr-only'>TorontoJS</span>
```
manvinderjit, line 48: @madcampos We are hiding the icon from the screen reader by using the aria hidden attribute and showing the <span>text</span> element in the screen ready only. Does this work from accessibility and design perspective? The figma design doesn't have text in the footer.
Screenshot from 2025-02-26 21-38-45

madcampos: If the design only provides icons, what you can do to make it accessible is something like this:
```html
<a href="#">
  <span class="visually-hidden">
    Home
  </span>
  <svg>
    ...
  </svg>
</a>
```

That `visually-hidden` class ensures it is accessible through a screen reader, but don't show up on the page.

Here is a reference for it: https://css-tricks.com/inclusively-hidden/

Also, on ARIA, avoid using it as much as possible and rely on the default semantics. ARIA is very powerful but is also very easy to shoot yourself in the foot and make it more confusing for users.

madcampos: Another suggestion is to try and use a screen reader. You can use [VoiceOver](https://support.apple.com/en-ca/guide/voiceover/welcome/mac) on Mac and iOS, [Narrator](https://support.microsoft.com/en-us/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1) or [NVDA](https://www.nvaccess.org/about-nvda/) on Windows, and [Orca](https://orca.gnome.org/) on Linux.

It is not that hard or weird and can help understand how things work. They all come with options to highlight what is being read on the screen

manvinderjit: > If the design only provides icons, what you can do to make it accessible is something like this:
> 
> ```
> <a href="#">
>   <span class="visually-hidden">
>     Home
>   </span>
>   <svg>
>     ...
>   </svg>
> </a>
> ```
> 
> That `visually-hidden` class ensures it is accessible through a screen reader, but don't show up on the page.
> 
> Here is a reference for it: https://css-tricks.com/inclusively-hidden/
> 
> Also, on ARIA, avoid using it as much as possible and rely on the default semantics. ARIA is very powerful but is also very easy to shoot yourself in the foot and make it more confusing for users.

@madcampos We are already implementing it this way and the `sr-only` class that we use for the `span` element is also from the link your provided.

Do you want us to remove the `id` attribute from the `span` tag because we are using to for `aria-labelledby` attribute of the `svg` tag.

If these are fine, can we merge the PR.

madcampos: So, adding the `aria-describedby` attribute to a `role=hidden` element won't do anything as it is already hidden and won't be announced at all. Then we should remove the `id`s and the `aria-labeledby` attributes. It will also avoid clashes as `id`s have to be unique within the page (and it is very annoying to do so).

I think this is a fine solution the way it is keeping the svg with `role=hidden` and the span with a `.sr-only` class.

Sidenote: There is [another solution](https://github.com/torontojs/volunteer/pull/17#discussion_r1994082504) if you want to explore, that is nicer for logos and all, for icons in a menu, the one here is good.

src/components/Footer/Footer.css
madcampos, line 1: I'm not sure how that will work, but I think those imports will be global. So a couple suggestions here:
1. Add a `@layer` so it won't clash with other styles and can propperly override everything that comes before.
2. Namespace the component with a class or id so the styles are contained.

For example, for this component:
```javascript
function Footer() {
  return <footer id="page-footer">...</footer>;
}
```

We would have this CSS:
```css
#footer {
  /* Every style is now contained to the element with an id of `footer`. */
}
```

madcampos (approved):
Approved with caveat of TODO comment

PR #22, opened 2025-03-02
feat: Button component with examples

sounmind:

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,0 +1,25 @@
+import type { ButtonHTMLAttributes } from 'react';
+import './Button.css';
+
+interface Props extends ButtonHTMLAttributes<HTMLButtonElement> {
+	hasOutline?: boolean;
+	isPrimary?: boolean;
+	isLarge?: boolean;
+}
+
+const Button = ({
+	hasOutline = false,
+	isPrimary = false,
+	isLarge = false,
+	children,
+	...rest
+}: Props) => (
+	<button
+		className={`button ${isPrimary ? 'primary' : ''} ${hasOutline ? 'outline' : ''} ${isLarge ? 'large' : 'small'}`}
+		{...rest}
+	>
+		{children}
+	</button>
+);
+
+export default Button;
--- a/src/pages/profile/profile.tsx
+++ b/src/pages/profile/profile.tsx
@@ -1,7 +1,7 @@
 import { StrictMode } from 'react';
 import { createRoot } from 'react-dom/client';
 
-import ProfileCard from '../../components/ProfileCard/ProfileCard.js';
+import Button from '../../components/Button/Button.js';
 
 import 'open-props';
 import './style.css';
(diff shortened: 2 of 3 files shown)


src/components/Button/Button.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const Button = ({
+	label,
+	disabled = false,
+	hasOutline = false,
+	isPrimary = false,
+	isLarge = false
+}: Props) => (
+	<button
+		disabled={disabled}
```
madcampos: **thought:** ‏Just something to think about: https://kittygiraudel.com/2024/03/29/on-disabled-and-aria-disabled-attributes/

Setting the `disabled` attribute on a form element makes it not focusable, thus invisible for accessibility. That has very valid use cases, but may not be the best for all scenarios.

E.g.: A login form has the button for logging in disabled before users fill in the email and password. That should use `aria-disabled` instead of `disabled` as the button will still be accessible by keyboard interaction while signaling that it _is_ disabled.

E.g.: A multi-step form, setting some parts of that as `disabled` make sense‏ as the user can and should only get to them after finishing other parts first.

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,26 @@
...
+		className={`button ${isPrimary ? 'primary' : ''} ${hasOutline ? 'outline' : ''} ${isLarge ? 'large' : 'small'}`}
```
madcampos, line 18: **thought:** Just an idea to think about.

For lists that get too large we could use some utility function (write it ourselves, please don't import a library for that) that turns an object into a string with the keys.‏

This can be as simple as:
```typescript
function htmlClasses(obj: Record<string, boolean>) {
  return Object.entries()
    .filter([, value]) => !Boolean(value))
    .map(([key]) => key)
    .join(' ');
}
```

Another option is to use [`data-*`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) properties instead of classes. So the HTML looks like this:

```html
<button data-is-primary>Click me!</button>
```

And the css looks like this:

```css
button[data-is-primary] {
  font-family: 'Comic Sans MS', fantasy;
}

button[data-is-primary] .icon {
  width: 2rem;
  height: 2rem;
}
```

That is useful for situations where there are a handful of options and they are well defined. Like using a `switch` instead of `if..else`.

It can also combine with other `data` attributes, or have a default value and then apply things based on what state it is in.

E.g.: A button that have options for primary, secondary, and error states, benefits from having that internally use a data attribute, as the states are well defined.

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface Props {
+	label: string;
+	disabled?: boolean;
+	hasOutline?: boolean;
+	isPrimary?: boolean;
+	isLarge?: boolean;
+}
+
+const Button = ({
+	label,
```
madcampos: **suggestion:** Use `children` instead.

I think here it makes more sense to give more control to consumers of the button. For example, imagine this scenario:
```html
<Button>
  <Icon />
  <span>Click <strong>ME!</strong></span>
</Button>
```

With only an attribute the component is less flexible, so it is beneficial here to just allow whatever so consumers are responsible for what is inisde the button while also having the freedom to do different designs.

src/pages/profile/profile.tsx
```diff
@@ -12,7 +14,33 @@ createRoot(root).render(
 	(
 		<StrictMode>
 			<div className='App'>
-				<ProfileList />
+				{/* NOTE: This is temp UI to test button component design */}
```
madcampos, line 16: **question:** Why not create a new page to test the button? 🙃😅🤣‏

src/pages/profile/profile.tsx
```diff
@@ -12,7 +14,33 @@ createRoot(root).render(
...
+				<h2>Render all the Button cases</h2>
+				<p>Default: Small, Not primary, Has not outline, Not disabled</p>
+
+				<div style={{ display: 'flex', flexDirection: 'column', gap: '1rem', width: 500 }}>
+					<Button label='Default Button' />
+					<Button label='Disabled Button' disabled />
+					<Button label='Outline Button' hasOutline />
+					<Button label='Outline Disabled Button' hasOutline disabled />
+
+					<Button label='Primary Small Button' isPrimary />
```
madcampos: **suggestion:** For grouping things, it is cool to use either [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) (with the `open` attribute set) or [`<fieldset>‏`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) (with some css to make it look nicer[^1]).

[^1]: This is a very old html element, so the default styles are a little bit ugly, but it is very usable and accessible by default. I like to just change it's border to `solid` and add some `border-radius` to make it nicer.

madcampos (commented):
Leaving a couple comments, they are nothing hard or ask for changes, it is mostly suggestions and food for thought on api design[^1] and ergonomics.

[^1]: "API design" here meaning the part that is exposed for other developers using the component library/design system. The interface people interact with _is_ an API and usually the design considerations are different than when designing an application.

src/pages/profile/profile.tsx
sounmind: There's no special reason. It's because it's really temp UI, I won't merge this changes to dev.

src/pages/profile/profile.tsx
sounmind: Good to know! But this UI will be removed when we merge it. So I think we don't need this.

src/components/Button/Button.tsx
sounmind: Applied it [here](https://github.com/torontojs/volunteer/compare/603e246b67a93a023f761b4336ea86004acb2d48..474d9421c8335efb96edad47620b6b5776311db7)!

src/components/Button/Button.tsx
sounmind: That's really great point!
We just made change like [this](https://github.com/torontojs/volunteer/compare/474d9421c8335efb96edad47620b6b5776311db7..6a34169d2011ad11ac908e3291438d433d093701).
We delegated disabled or aria-disabled attributes passing decision to user of Button component extending HtmlButtonAttributes.

src/components/Button/Button.tsx
sounmind: We really appreciate the opportunity to consider these fantastic points! While we believe it's not the right moment to implement a utility function for managing the three classes, it's definitely something we see doing in the future. As for the excellent feature of data-* usages, we're happy with how our class management is going, so we're planning to continue with our current approach for now.

manvinderjit (commented):
Looks good!

sounmind:
We're going to merge this PR to implement further components/pages.
Feel free to add comments here if you have any concerns!

src/components/Button/Button.tsx
evert: Very interesting share @madcampos !

PR #33, opened 2025-03-25
Basic css reset

### LINKED ISSUE #32: Basic CSS Reset

kbventures:
## Summary

Basic CSS reset


src/index.css
```diff
@@ -1,41 +1,81 @@
+/* --- General Reset and Box-Sizing --- */
+
+/* Ensure consistent box-sizing across all elements */
+*, *::before, *::after { box-sizing: border-box; /* Prevents padding and borders from affecting element size */ }
```
madcampos: **nitpick:** Add the comment before the css rules. It aids with readability by putting the explanation in one place close to the code, but not intertwined _with_ the code.

```suggestion
/* Prevents padding and borders from affecting element size */ 
*, *::before, *::after { box-sizing: border-box; }
```

src/index.css
```diff
@@ -1,41 +1,81 @@
...
+
+/* Reset margin and padding for all elements */
+* {
+	margin: 0;
+	padding: 0; /* Removes default spacing */
+}
```
madcampos: **thought:** ‏Removing margins and paddings can actually be detrimental in some places. If we do that, we would have to recreate the browser default stylesheet.

Having margins in places like paragraphs and headings actually helps.

And a blanket rule like that may cause unintended consequences because we can't foresee new html elements and their default styles (think dialog 5 years ago).

Please limit this rule to only places that are _**always**_ reseted, like the margin on `body`. If we want to reset or set a base style to things like headers, those should be done on rules specific to the element, so it is easier to find out where a rule is coming from.

src/index.css
```diff
@@ -1,41 +1,81 @@
...
+
+/* --- Remove Default Styling for Lists --- */
+ul, ol { list-style: none; /* Removes default bullet points and numbering */ }
```
madcampos:
```suggestion
nav :is(ul, ol) { list-style: none; /* Removes default bullet points and numbering */ }
```

We _**DON'T**_ want to remove list styles from everything. The only place where that is consistently done and makes sense is when a list is used as part of a navigation menu.

Like this:
```html
<nav>
  <ul>
    <li><a href="#">Link 1</a></li>
    <li><a href="#">Link 2</a></li>
    <li><a href="#">Link 3</a></li>
  </ul>
</nav>
```

For other common scenarios, like in a text body, we want to keep the list styles.

src/index.css
```diff
@@ -1,41 +1,81 @@
...
+
+/* --- Remove Default Button Styles --- */
+button {
+	all: unset; /* Removes all default styles applied by the browser */
```
madcampos:
```suggestion
```

A blanket rule like this have unintended consequences. Yes, it removes all styles that make a button look like a button. But it is better to be specific and overwrite only the styles that we need to make our buttons look more like we want.

src/index.css
```diff
@@ -1,41 +1,81 @@
...
+	cursor: pointer; /* Ensures buttons have a clickable cursor */
+}
+
+/* --- Ensure Images are Responsive --- */
+img {
```
madcampos:
```suggestion
:where(img, iframe, svg) {
```

We want to apply that to all "inline-ish" elements, in that case, those are images, iframes and inline SVGs[*].

The [`:where()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) selector here helps as it keeps the [selector specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Specificity) at 0, so every rule will overwrite it. This helps avoid unintended consequences like the use of `!important`.

[*]: We may carve out an exception for SVGs here like icons, where they have different styles applied.

src/index.css
```diff
@@ -1,41 +1,81 @@
...
+	display: block; /* Removes inline-block spacing issues */
+	height: auto; /* Ensures the aspect ratio is maintained */
+	max-width: 100%; /* Makes images responsive to their container width */
+}
+
+/* --- Inherit Font for Form Elements --- */
+input, textarea, select, button { font: inherit; /* Inherits the font-family from parent elements */ }
```
madcampos: **question:** ‏This will make _all_ font properties be inherited, that includes italics, bold, font size, and font family. Do we want that?

_I_ usually, only use `font-family` and `font-size`, as bold and italics of buttons don't change depending on the text surrounding them.

src/index.css
```diff
@@ -1,41 +1,81 @@
...
+
+/* --- Set Full Height for HTML and Body --- */
+html, body { height: 100%; /* Ensures full-height layouts work as expected */ }
+
+/* --- Link Styling --- */
+a {
+	color: inherit; /* Inherits text color from parent elements */
+	text-decoration: none; /* Removes the default underline */
```
madcampos: **issue:** Removing the text decoration of links is an accessibility issue‏, as it removes a visual indicator that that snippet of text is a link.

Only do this if you are replacing the underline with something else that conveys that the link is a link.

**_If you are doing that because you use link tags for things like creating a tab component or for "link buttons" I will personally hunt you down and make you suffer._**

References:
- https://adrianroselli.com/2016/01/links-buttons-submits-and-divs-oh-hell.html
- https://webaim.org/techniques/hypertext/link_text#appearance

src/index.css
```diff
@@ -1,41 +1,81 @@
...
+}
+
+/* --- Reset Scrollbars --- */
+html, body { overflow-x: hidden; /* Prevents horizontal scrolling */ }
```
madcampos: **issue:** ‏**DON'T!!!**

If the content overflows horizontally, it is a problem to fix in other ways rather than hiding overflow. This _cuts_ information for users by hiding it and making it inaccessible.

I rather have the UI break than make it inaccessible for users.

src/index.css
madcampos, line 1: Please move the new styles to a new file and `@import` it here.

madcampos (commented):
I CHOOSE VIOLENCE

image

src/index.css
kbventures: Recreating the browser default stylesheet was the intent.  Every browser has their own defaults  and therefore starting from a clean slate doesn't introduce quirks for specific browsers.

madcampos: I still don't like blanket rules like that.

Browser stylesheets are more common than they are different, that used to be an issue in the IE era, but now we are in a way better place.

The other point, and the main reason I don't like it, is not all elements have margins or paddings, and again, removing them can create unintended consequences.

To reiterate: if we want to reset margins and paddings, we should do so on a case-by-case basis and not as a one size fits all approach.

src/basic-css-reset.css
madcampos, line 1: Ain't we missing the styles here?

src/index.css
```diff
@@ -1,41 +1,35 @@
...
 	--leading-h2: calc(var(--leading-base) - 4 * 0.08em); /* Adjusted line height for h2 */
 	--leading-h1: calc(var(--leading-base) - 5 * 0.08em); /* Adjusted line height for h1 */
-	
 	/* Text size settings based on scale */
 	--text-xs: calc(1em * pow(var(--text-scale-ratio), -2)); /* Extra small text size */
 	--text-sm: calc(var(--text-base) * pow(var(--text-scale-ratio), -1)); /* Small text size */
 	--text-md: calc(var(--text-base) * pow(var(--text-scale-ratio), 1)); /* Medium text size */
 	--text-lg: calc(var(--text-base) * pow(var(--text-scale-ratio), 2)); /* Large text size */
-	--text-xl: calc(var(--text-base) * pow(var(--text-scale-ratio), 3)); /* Extra large text size */
+	--text-xl: calc(var(--text-base) * pow(var(--text-scale-ratio), 4)); /* Extra large text size */
```
lylaminju: Why did you change 3 to 4 ? It seems like a typo

src/basic-css-reset.css
lylaminju, line 1: Thanks for adding the reset.css !

```
ul,
ol {
	padding-left: 0;
}

a { text-decoration: none; }
```

I found these styles are needed for the pages I'm currently working on. Could you please add these too?

I made my own [reset.css](https://github.com/torontojs/communityhub-fe/pull/37/files#diff-17622d1494301e1cd7f0160335de0c15c8fe0b1784e0daa6aa63cd3c34d383ce) based on what I needed, but I'll remove it soon.

madcampos: @kbventures I think some of the content of the file got deleted by mistake, the previous version of the reset seemed longer than this... 👀

Is that the full reset file or am I mistaken?

src/index.css
```diff
@@ -1,41 +1,35 @@
-/* Font settings */
+@import 'basic-css-reset';
```
madcampos:
```suggestion
/**
 * Set all of the layers to be used by CSS.
 * It allows us to define styles that can be overriden by the next step without using !important everywhere.
 * And not have to be fighting the cascade.
 */
@layer reset, fonts, design-system, base, pages, components, theme, print, accessibility;

/**
 * Import the reset into the reset layer.
 */
@import './basic-css-reset.css' layer(reset);
```

The import is a URL, so I prefer having it as a relative URL here (starting with a `./`).

Also added [`@layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) so we can organize things into layers.

Here is a video on CSS layers: https://www.youtube.com/watch?v=ilrPpSQJb3U

src/index.css
```diff
@@ -1,41 +1,35 @@
...
+
+/* Design system */
 :root {
```
madcampos: **suggestion**: we should move it to another file and set the needed layers for it.

I particularly like to have two files: one for font definition and declarations and another with the rest of the variables that define a design system.

src/index.css
```diff
@@ -10,86 +10,4 @@
...
-	--shadow-xs: 0 calc(0.5 * var(--shadow-base)) calc(0 * var(--shadow-base)) var(--color-shadow-light); /* Extra small shadow */
-	--shadow-sm: 0 calc(0.5 * var(--shadow-base)) calc(1 * var(--shadow-base)) var(--color-shadow-light); /* Small shadow */
-	--shadow-md: 0 calc(1 * var(--shadow-base)) calc(2 * var(--shadow-base)) var(--color-shadow-light); /* Medium shadow */
-	--shadow-lg: 0 calc(2 * var(--shadow-base)) calc(4 * var(--shadow-base)) var(--color-shadow-heavy); /* Large shadow */
-	--shadow-xl: 0 calc(4 * var(--shadow-base)) calc(8 * var(--shadow-base)) var(--color-shadow-heavy); /* Extra large shadow */
-	--shadow-2xl: 0 calc(8 * var(--shadow-base)) calc(12 * var(--shadow-base)) var(--color-shadow-heavy); /* 2x extra large shadow */
-	/* Accent color */
-	accent-color: var(--color-accent); /* Accent color used throughout the site */
-}
+@import './design-system.css' layer(design-system);
```
madcampos, line 10: **issue**: The names have to match the first `@layer` definition, that is what helps set the order we want them to be applied. Or else the layer order will fallback to the order they were declared, thus defeating the purpose of having layers in the first place.

That gets worse with bundling as we (sorta, kinda) cannot guarantee the order the bundler will inline, split, or organize the CSS for us.

So, these names:
```css
@layer reset, fonts, design-system, base, pages, components, theme, print, accessibility;
```

Have to match these names:
```css
@import './basic-css-reset.css' layer(general-reset);
@import './basic-styling.css' layer(basic-styling);
@import './design-system.css' layer(design-system);
```

Pick the ones you think look nicer, this is CSS after all

madcampos (approved):

PR #35, opened 2025-03-30
style: complete profile page layout

### LINKED ISSUE #26: Implement profile pages from figma design
We need to implement the pages _13 - Complete profile_ and _14 - Profile completed_ from the Figma design.

Image

Image

## Notes
- Start with the html structure, css comes second
- After the structure, it should connect to the backend for requests
- If you need any components, for now, stick to plain html and just wrap it to a dummy component

## Sub tasks

- [ ] Validate user input
- [ ] Form submission handling
- [ ] Users should not be able to edit email
- Upload photo doesn't have a backend yet, so it is okay to not send it for now
  - [ ] ~~Photo upload should have a size limit~~
- [x] Create tag list for skills

sounmind:
## Summary

This is style changes for profile page layout based on this [Figma design](https://www.figma.com/design/62ZRtfpGdAz2uuWwVMAmQg/VMS-Project?node-id=219-5922&t=XwaPeRQtdE0lmi9v-0)

## Tests

We've coded with @madcampos, it's approved...
We did not test mobile version yet.

## Additional information


src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,241 @@
...
+						<div>
+							<label htmlFor='dev-to'>DEV.to</label>
+							<input id='dev-to' name='dev-to' type='url' />
+						</div>
+					</div>
+				</div>
+			</details>
+		</div>
+
+		<Button isPrimary id='submit-button' type='submit'>
```
kbventures: @sounmind @ImmaCheetah 

The button width isn't matchign Figma design on the mobile and full screen version:

image

See the Figma design:

image

The left and right spacing is not matching the mobile FIgma design.

kbventures:
@ImmaCheetah @madcampos 

Using padding instead of margin between elements can cause accessibility issues, especially for users with high zoom levels or screen magnifiers. This is because padding increases the clickable area inside an element, which can lead to overlapping interactive regions. Margin is better for spacing between elements to maintain clear separation.

madcampos:
@kbventures 
> Using padding instead of margin between elements can cause accessibility issues, especially for users with high zoom levels or screen magnifiers. This is because padding increases the clickable area inside an element, which can lead to overlapping interactive regions. Margin is better for spacing between elements to maintain clear separation.

[Um, actually](https://www.youtube.com/watch?v=Ft6N64pPll0)... An increased hit area is not bad for accessibility, specially on mobile (looking at you Instagram with it's tiny "x" to close things).

Now, padding _do indeed_ influence the element's size, either by _adding_ to the set size when `box-sizing` is set to the default value of `content-box` or by _subtracting_ from the content size when `box-sizing` is set to the way more sane `border-box`.

As a general rule of thumb, I use padding when I need to influence the element's children, and margin when the element influence it's relationship with sibling elements.

E.g.:

> When you want to add space between a container's border and the text, so the text is not hugging the border of the conteiner, using padding is a good solution.
>
> When you want to add spaces between multiple `<p>` blocks of text, adding a margin to them is a good solution.

My question is: where, specifically, you found the padding to be a problem?

References:
- https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events
- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_box_model/Mastering_margin_collapsing

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -56,23 +56,57 @@
...
 				/* grid: repeat(3, 80px)  */
 
 				/* TODO: move input styles to its own component */
-				input { width: 100%; }
+				.text-input { 
+					width: 100%;
+					height: 2.33rem;
+					box-sizing: border-box;
+					-webkit-box-sizing:border-box;
+					-moz-box-sizing: border-box;
```
madcampos: **issue:** Please remove vendor prefixed properties. There is only a handful of properties that still need vendor prefixes, and `box-sizing` is not one of them.

It has been widely available on browsers since July 2015‏.

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -56,23 +56,57 @@
...
+						background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='12' height='6' viewBox='0 0 12 6' fill='none'%3E%3Cpath d='M6 6L0 0H12L6 6Z' fill='%23999999'/%3E%3C/svg%3E");
+						background-position: center;
+						background-repeat: no-repeat;
+						background-size: contain;
+						content: '';
+						display: inline-block;
+						height: 1rem;
+						width: 1rem;
+					}
+				} */
```
madcampos: Unfortunately there is no cross browser way of styling the small arrow added by the datalist element on chromium browsers. Just leave it as is. Not the prettiest, but it works 🤷‍♂️

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -56,23 +56,57 @@
...
 			}
 
 			/* Media query for smaller screens */
-			@media (max-width: 768px) { .inputs-row-wrapper { grid-template-columns: 1fr; } }
+			@media (max-width: 768px) { .details-content-grid { grid-template-columns: 1fr; } }
 
-			.input-row-required {
+			.input-required {
+				& > span {
```
madcampos:
```suggestion
			.input-required span {
```

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -119,7 +128,61 @@
 			.slider-checkbox-row {
 				display: grid;
 				gap: var(--spacing-xs);
-				grid-template-columns: 1fr 8fr;
+				grid-template-columns: repeat(1fr, 2fr);
```
madcampos: I think there is a typo here... The `repeat` will be ignored.

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,256 @@
...
+					color: var(--color-accent);
+					font-size: 0.56rem;
+					font-weight: 400;
+					height: 1rem;
+					letter-spacing: 0;
+					line-height: 160%;
+				}
+				align-items: center;
+				display: flex;
+				justify-content: space-between;
```
madcampos:
```suggestion
			.input-required {
				align-items: center;
				display: flex;
				justify-content: space-between;
				
				& > span {
					color: var(--color-accent);
					font-size: 0.56rem;
					font-weight: 400;
					height: 1rem;
					letter-spacing: 0;
					line-height: 160%;
				}
```

**nitpick:** Please put the element styles before any nesting.‏

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,256 @@
...
+					border-radius: var(--rounded-xs);
+					box-sizing: border-box;
+					-webkit-box-sizing: border-box;
+					-moz-box-sizing: border-box;
+					height: 2.33rem;
+					padding: var(--spacing-xs);
+					width: 100%;
+				}
+
+				/* .text-input#pronouns {
```
madcampos: **nitpick:** Please let's keep the code tidy and remove unused and commented out pieces.‏

src/pages/complete-profile/style.css
```diff
@@ -0,0 +1 @@
+body { margin: 0; }
```
madcampos, line 1: **suggestion:** Let's centralize "CSS resets", please refer to: https://github.com/torontojs/communityhub-fe/pull/33‏

manvinderjit:
@madcampos @kbventures Please re-review this! Thanks!

src/pages/complete-profile/style.css
manvinderjit: Done!

madcampos:
Hey folks, I ran the code locally and used NVDA to test it out...

Here is how it is read by a screen reader:

https://github.com/user-attachments/assets/3a911e15-9119-4199-b9a7-8aca2fe57a33

Notice some things:

1. The upload avatar button is entirely skipped, it is not accessible through keyboard navigation
2. All of the social media buttons announce as "button", it doesn't say what they do
3. The fields marked as required have the "required" part read twice

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+				background-color: white;
+				border: var(--border);
+				border-color: var(--color-accent);
+				border-radius: 10px;
+				content: '';
+				height: 18px;
+				margin-right: 0.4rem;
+				min-width: 36px;
+				transition: 300ms ease-in-out;
+				width: 36px;
```
madcampos: **issue:** Please move the pixel values to use rem.‏

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+				height: 14px;
+				left: 3px;
+				position: absolute;
+				transition: 300ms ease-in-out;
+				width: 14px;
+			}
+
+			.slider-checkbox:checked + label::after {
+				background-color: white;
+				left: 20px;
```
madcampos: Same here

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+		align-items: center;
+		display: flex;
+		gap: var(--spacing-sm);
+		justify-content: center;
+
+		& picture {
+			max-height: 128px;
+			max-width: 128px;
+			min-height: 52px;
+			min-width: 52px;
```
madcampos: And here

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+
+			& img {
+				aspect-ratio: 1;
+				border-radius: 0.5rem;
+				height: 100%;
+				object-fit: cover;
```
madcampos: **suggestion:** we should use `contain` instead of `cover` here as it does not crop images that are not square. `cover` will make the image fill all the available space but crop it to fit the space, where `contain` will resize the image to fit inside the area.

On the screenshot the image is rectangular, where the preview is square and crops the image:
image
‏

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+
+		& .details-file-upload-buttons-wrapper {
+			display: flex;
+			flex-direction: column;
+			gap: var(--spacing-xs);
+			justify-content: start;
+			min-width: 244px;
+			width: 244px;
+
+			& input[type='file'] { display: none; }
```
madcampos: **issue:** This removes the input from the DOM, making it not accessible for keyboard users. A solution is do something similar to the checkbox, where it is visibly hidden, but not removed from the DOM.‏

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+					font-size: var(--text-sm);
+					letter-spacing: 0;
+					line-height: 160%;
+				}
+			}
+
+			.slider-checkbox {
+				opacity: 0;
+				position: absolute;
+			}
```
madcampos: **issue:** The checkbox doesn't have a focus state. So when keyboard users navigate to it, it does not show any indicator that the element is selected.‏

Here is a video recording of it:

https://github.com/user-attachments/assets/ff75584d-6ec0-4823-ae9c-10f0098e00b9

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+
+			& .custom-file-upload {
+				border-radius: var(--rounded-md);
+				color: var(--color-black);
+				cursor: pointer;
+				font-size: var(--text-xs);
+				font-weight: var(--weight-semibold);
+				letter-spacing: 0;
+				line-height: 160%;
```
madcampos: **nitpick:** messing with `line-height` can cause issue due to how it is calculated on CSS. In regular image/text editing software it is calculated based on the x-height, meaning, the height of lowercase characters, not counting ascenders, nor descenders (like the stick going up on "d" or down on "p").

On CSS the line height is calculated by drawing the whole text inside a box, and then adding equal space on top and bottom from it. So ascenders and descenders count for CSS line height, where for other tools they do not.

So, it is not a very good idea to change the line height as it may cause more troubles than it is worth.

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+				text-align: center;
+			}
+
+			label.custom-file-upload { background-color: var(--color-card); }
+
+			button.custom-file-upload {
+				background-color: var(--color-background);
+				border: var(--border);
+				border-color: var(--color-black);
+			}
```
madcampos: **issue:** The "remove photo" button is rendering very differently than the label with the same style. Maybe setting the button `appearance` and `font` properties may help.‏

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+		}
+	}
+
+	#details-information-grid {
+		display: grid;
+		gap: var(--spacing-sm);
+		grid-template-columns: repeat(3, 1fr);
+		justify-content: space-between;
+
+		label[for='linkedInProfile'], label[for='githubProfile'], label[for='portfolio'] {
```
madcampos:
```suggestion
		label:is([for='linkedInProfile'], [for='githubProfile'], [for='portfolio']) {
```

The `:is` selector works as a grouping selector.

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+			display: inline-block;
+			margin-bottom: var(--spacing-xs);
+		}
+
+		#details-information-skills {
+			display: flex;
+			flex-direction: column;
+			grid-column: 1 / span 3;
+
+			> label {
```
madcampos:
```suggestion
			& > label {
```

**nitpick:** To disambiguate any complex selector that is not the children selector (just a space between selectors), it is better to add the `&` here.‏

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+				border-radius: var(--rounded-base);
+				border-style: solid;
+				display: flex;
+				flex-wrap: wrap;
+				gap: var(--spacing-xs);
+				min-height: 4rem;
+				padding: var(--spacing-xs);
+				resize: none;
+
+				> span {
```
madcampos: Same here

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+					font-size: var(--text-xs);
+					font-weight: var(--weight-semibold);
+					gap: var(--spacing-xs);
+					letter-spacing: 0;
+					line-height: 160%;
+					padding-left: var(--spacing-xs);
+					padding-right: var(--spacing-xs);
+					transition: all ease-in-out 300ms;
+
+					> button {
```
madcampos: And here

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+				gap: var(--spacing-xs);
+				min-height: 4rem;
+				padding: var(--spacing-xs);
+				resize: none;
+
+				> span {
+					align-items: center;
+					background-color: var(--color-accent);
+					border: solid var(--border-base) var(--color-accent);
+					border-radius: 99px;
```
madcampos:
```suggestion
					border-radius: 100vmax;
```

Neat little trick to make things round 😉

It uses the largest viewport unit (`vmax`) and makes it as big as the whole viewport (100 of the unit).

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,401 @@
...
+						background-color: var(--color-white);
+						color: var(--color-accent);
+						> button { color: var(--color-accent); }
+					}
+				}
+
+				input#skill {
+					background: none;
+					border: solid var(--border-base) var(--color-grey);
+					border-radius: 99px;
```
madcampos:
```suggestion
```

Don't add `border-radius` here as it crops the input text.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,399 @@
...
+										className='text-input'
+										id='name'
+										name='name'
+										type='text'
+									/>
+								</div>
+								<div>
+									<label htmlFor='email' className='input-required'>
+										E-mail
+										<span>REQUIRED</span>
```
madcampos: **issue:** This "required" text gets read twice by screen readers, first from the input being marked as required, second from the `span` here. We can change it to be added with CSS pseudo elements (`::before` or `::after`).‏

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,399 @@
...
+										<option>They/them</option>
+									</datalist>
+								</div>
+								<div>
+									<span>Date of birth</span>
+									<br />
+									<div className='dob-wrapper'>
+										<label htmlFor='month'>Month</label>
+										<select id='month' name='month'>
+											<option value='01'>January</option>
```
madcampos, line 179:
```suggestion
											<option selected disabled hidden value=''>Select a month</option>
											<option value='01'>January</option>
```

This is a neat trick to add an option that doesn't show up but still adds a "placeholder" for the user.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,399 @@
...
+											<option value='07'>July</option>
+											<option value='08'>August</option>
+											<option value='09'>September</option>
+											<option value='10'>October</option>
+											<option value='11'>November</option>
+											<option value='12'>December</option>
+										</select>
+										<label htmlFor='day'>Day</label>
+										<select id='day' name='day'>
+											<option value='01'>1</option>
```
madcampos, line 195:
```suggestion
											<option selected disabled hidden value=''>Select a day</option>
											<option value='01'>1</option>
```

Same trick here.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,399 @@
...
+								<div className='slider-wrapper'>
+									<div className='slider-checkbox-row'>
+										<input
+											id='isBasedOnGTA'
+											name='isBasedOnGTA'
+											type='checkbox'
+											className='slider-checkbox'
+										/>
+										<label htmlFor='isBasedOnGTA'>
+											I'm based in Toronto or Greater Toronto Area
```
madcampos: **suggestion:** Here an option is to put the checkbox inside the label along with a span containing the text for the checkbox.‏

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,399 @@
...
+		<>
+			<StepBar
+				currentStep={3}
+				steps={[
+					{ label: 'Account confirmed' },
+					{ label: 'Check the conduct code' },
+					{ label: 'Complete your profile' }
+				]}
+			/>
+			<form action='' encType='multipart/form-data' id='complete-profile-form'>
```
madcampos, line 108:
```suggestion
			<form action='' encType='multipart/form-data' id='complete-profile-form' method='post'>
```

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,399 @@
...
+									<label htmlFor='slack-handle' className='input-required'>
+										Slack handle
+										<span>REQUIRED</span>
+									</label>
+									<input
+										className='text-input'
+										id='slack-handle'
+										name='slack-handle'
+										type='text'
+										aria-description='Your slack handle to TorontoJS'
```
madcampos:
```suggestion
```

This is being announced twice.

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: Done!

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: @madcampos The reason for using cover was so that the border radius applies to the image. Because if the image doesn't have a 1:1 aspect ratio, the border radius effect is virtually invisible with contain. See attached image for details.
Screenshot from 2025-05-06 15-06-07
Screenshot from 2025-05-06 15-06-15
Screenshot from 2025-05-06 15-06-25

manvinderjit: Do you recommend making the width fluid and height fixed with `contain`?

madcampos: Use this: https://github.com/madcampos/sdrlog/blob/main/src/components/SdrCard/style.css#L132-L152

Wrap the `img` in a `picture`, then the `picture` sets the size you want, the `img` can grow up to it, but is bound to the box size. It will keep the aspect ratio though.

manvinderjit: Got it, thanks.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Will add this. Just for future reference, we are using a separate branch and PR for implementing the form submission functionality in Complete Profile. 
https://github.com/torontojs/communityhub-fe/pull/54

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: They are in this order because I am adding the toggle sliders using CSS pseudo elements. The input checkbox is invisible here anyways.

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: @madcampos  Replaced label with button, should resolve now.

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: Implemented the change.

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: @madcampos  Change the logic and visibility, should resolve now.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @madcampos  Replaced aria-description with aria-label, should only read it once, please verify.

manvinderjit:
@madcampos I have addressed all issues. Please review and mark them resolved if you are satisfied. Let me know if any issues crop up. Thank you.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Also, is there an Ubuntu alternative for NVDA?

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: Removed most cases of line-height.

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: @manvinderjit yes, here is a list of options for screen readers:
- For linux there is [Orca](https://orca.gnome.org/).
- Android comes with [TalkBack](https://support.google.com/accessibility/android/answer/6007100?hl%3Den), you may need to download the app from the playstore though.
- Mac has [VoiceOver](https://support.apple.com/en-ca/guide/voiceover/welcome/mac) builtin.
- iOS also have [VoiceOver](https://support.apple.com/en-ca/guide/iphone/iph3e2e415f/ios) builtin.
- Windows has [Narrator](https://support.microsoft.com/en-us/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1) builtin.
- There is also [JAWS](https://www.freedomscientific.com/products/software/jaws/) for windows, but it is paid.

src/components/CompleteProfile/CompleteProfile.css
madcampos: There is still a typo here...

```suggestion
				grid-template-columns: repeat(2, 1fr);
```

See: https://developer.mozilla.org/en-US/docs/Web/CSS/repeat

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,413 @@
...
+				transition: 300ms ease-in-out;
+				width: 0.75rem;
+			}
+
+			.slider-checkbox:checked + label::after {
+				background-color: white;
+				left: 1.313rem;
+			}
+
+			.slider-checkbox:focus + label::before, .slider-checkbox:checked:focus + label::before { outline: auto Highlight; }
```
madcampos: **issue:** [`outline`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline) has a similar syntax to border‏.

Please either use the shorthand syntax specifying the type, thickness, and color. Or use the longhand syntax for changing only some of the properties.

Also: `Highlight` is a [system color](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color), I think it is not the best option to use here.

The two common scenarios for using system colors are:
- When using the `forced-colors` media query
- When trying to mimic an operating system design, like trying to make the page look like a Windows or MacOS application.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,419 @@
...
+										<label htmlFor='canJoinLocalEvents'>
+											I can join TorontoJS's local events
+										</label>
+									</div>
+								</div>
+							</div>
+						</div>
+					</details>
+
+					{/* TODO: Need to confirm adding green check icon on successful section completion on mobile view */}
```
madcampos, line 257: **question:** Was this confirmed?‏

src/components/CompleteProfile/CompleteProfile.css
madcampos: The focus style for the checkbox doesn't provide enough contrast and the checkbox is missalligned with the actual toggle.

image

Please position the input so it overlaps the visible toggle.

For the outline style, setting [`outline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset) helps here as it will move the outline outside of the toggle button, creating some space between the two thus increasing the contrast.

Outlines are drawn on top of the box and don't add to the box sizing, like border does, so adding an offset to the outline won't cause a [layout shift](https://developer.mozilla.org/en-US/docs/Glossary/CLS).

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -0,0 +1,413 @@
...
+
+			.input-required {
+				align-items: center;
+				display: flex;
+				justify-content: space-between;
+			}
+
+			.input-required::after {
+				color: var(--color-accent);
+				content: 'REQUIRED';
```
madcampos:
```suggestion
				content: 'Required' / '';
				text-transform: uppercase;
```

The second part of the `content` property is text added to accessibility technologies for generated content. Without it all modern browsers will announce the "required" text. 🥲

It is only recently supported however (since 2024), so for other projects add it with and without the alt text.

The `text-trasnform` property does auto transformation for us.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,419 @@
...
+										)}
+									</div>
+								</div>
+								<div id='details-social-inputs'>
+									{socialAccountInputs.map((socialInput) => (
+										<div>
+											<label htmlFor={socialInput}>
+												{socialInput}
+												<button
+													type='button'
```
madcampos, line 393:
```suggestion
													type='button'
																										aria-label={`Close ${socialInput} input`}
```

These button gets announced as only "button"

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: It still gets announced twice. The browser will read whatever text is in the placeholder (I didn't knew about that, before browsers didn't read it).

Let's remove the `aria-label`

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,419 @@
...
+										<img src={photoFile} />
+									</picture>
+								)}
+
+								<div className='details-file-upload-buttons-wrapper'>
+									{photoFile && (
+										<span className='file-upload-success'>
+											Avatar uploaded successfully
+										</span>
+									)}
```
madcampos:
```suggestion
									<span ref={avatarUploadStatusRef} aria-live='polite' aria-atomic='true' role='status' className='file-upload-success'>
										{photoFile ? 'Avatar uploaded successfully' : ''}
									</span>
```

This will make the upload be announced (or should make it so). And of course, we need to add a ref here because react is a piece of dumpster fire

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -0,0 +1,419 @@
...
+
+interface Skill {
+	id: string;
+	name: string;
+}
+
+const CompleteProfile = () => {
+	const fileInputRef = useRef<HTMLInputElement>(null);
+	const uploadPhotoButtonRef = useRef<HTMLButtonElement>(null);
+	const skillInputRef = useRef<HTMLInputElement>(null);
```
madcampos, line 27:
```suggestion
	const skillInputRef = useRef<HTMLInputElement>(null);
	const avatarUploadStatusRef = useRef<HTMLSpanElement>(null);
```

The other ref

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Yeah, it is a bit of mixed bag. I have read in multiple forums that `aria-label` supersedes the `placeholder` with only the former being announced but seems like it's not a universal thing.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Not yet, will bring it up with Cadu and Mariana again.

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: I didn't know this one. Thanks for sharing.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: I need to start using a screen reader for catching these.

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: It used to be the case that placeholders were not read. But maybe it is the combo Edge + NVDA

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: @madcampos Applied these changes.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @madcampos Removed the `aria-label`s

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: No update yet from design team, will implement this if and when confirmed.

manvinderjit:
@madcampos Hopefully should be resolved now. Let me know if something still needs work.

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: Let's leave it as an "outstanging item" as it is a non blocker and we can address later

madcampos (approved):

PR #37, opened 2025-04-01
Feat/pre profile completion

### LINKED ISSUE #34: Tabs Construct, Tab 1, Tab 2 and structure for create profile
Image

lylaminju:


src/components/Button/Button.css
```diff
@@ -30,11 +31,14 @@
...
+		&:disabled,
+		&[aria-disabled='true'] { color: var(--color-white); }
 	}
 
-	&:disabled, &[aria-disabled='true'] {
+	&:disabled,
+	&[aria-disabled='true'] {
 		color: var(--color-black);
+		cursor: not-allowed;
 		opacity: 50%;
```
madcampos: **issue:** ‏don't lower the opacity of disabled element. It is bad for accessibility as the contrast lowers due to the reduced opacity.

We should opt for another way to convey that a control is disabled.

src/components/CheckSteps/CheckSteps.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+button {
+	height: 59px;
+	padding-bottom: 12px;
+	padding-left: 16px;
+	padding-right: 16px;
+	padding-top: 12px;
+	width: 298px;
+}
```
madcampos: **issue:** This rule will affect all buttons. Please prefix it with a class or id, efectivly namespacing the rules to the component you want to apply these styles to.

E.g.:‏
```css
#check-steps button {
  /* ... */
}
```

src/components/CheckSteps/CheckSteps.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Header from '../Header/Header';
+import StepBar from '../StepBar/StepBar';
+import './CheckSteps.css';
+import { CheckConductCode } from './icons/CheckConductCode';
+import { CompleteProfile } from './icons/CompleteProfile';
+
+const CheckSteps = () => (
+	<div className='check-steps-wrapper'>
+		<StepBar step={1} />
+		<div className='check-steps'>
```
madcampos: **nitpick:** If this component is used only in one place (i.e. it is a screen), and will not be used in multiple places, please use ids instead of classes.‏

src/components/CheckSteps/CheckSteps.css
```diff
@@ -0,0 +1,48 @@
+@import 'open-props/open-props.min.css';
```
madcampos: **question:** Isn't open props already imported globally?

Or rather, I think we have a global css file with some default styles, shouldn't that be used here?

src/components/CheckSteps/CheckSteps.css
```diff
@@ -0,0 +1,48 @@
...
+
+.check-steps {
+	align-items: center;
+	display: flex;
+	flex-direction: column;
+	gap: var(--spacing-lg);
+	margin-inline: auto;
+	padding: var(--spacing-lg) var(--spacing-md);
+
+	@media screen and (min-width: 640px) { max-width: 400px; }
```
madcampos: **question:** Why not set a `max-width` of `100%` or a `vw` based size here?‏

The reasoning: with a screen size of 640 pixels, 400 will be most of it, at least part of the remaning space will be taken care by paddings, borders, and margins.

src/components/CheckSteps/icons/CheckConductCode.tsx
madcampos, line 1: **thought:** should we move all icons to a "root" `icons` folder and add the `icon` infix to the component names, like `SomethingIcon` or `IconSomething`?‏

cc @kbventures

src/reset.css
madcampos, line 1: **issue:** Let's not add multiple reset files, those rules should be taken care of by the global css files.‏

Also, please refer to this PR's comments, as some of them apply here as well: https://github.com/torontojs/communityhub-fe/pull/33

src/components/ReviewConductCode/icons/ArrowUp.tsx
madcampos, line 1: **suggestion:** Arrow icons are ‏usually just rotated versions of another icons, we can apply [CSS transforms](https://developer.mozilla.org/en-US/docs/Web/CSS/transform) to them to acheive the same outcome. With the advantage that transforms are animatable.

src/components/CheckSteps/CheckSteps.css
```diff
@@ -0,0 +1,48 @@
...
+	border-radius: var(--rounded-lg);
+	border-width: var(--border-sm);
+	display: flex;
+	flex-direction: column;
+	font-weight: var(--weight-bold);
+	gap: var(--spacing-xs);
+	justify-content: start;
+	padding: var(--spacing-sm);
+	text-align: center;
+}
```
madcampos: Same as the `button` CSS rule also applies here.

src/components/CheckSteps/CheckSteps.tsx
```diff
@@ -0,0 +1,37 @@
...
+				<li className='card'>
+					<CompleteProfile />
+					<p className='text-h6'>Complete your profile</p>
+				</li>
+			</ul>
+			<Button
+				isPrimary
+				isLarge
+				onClick={() => {
+					window.open('/pages/review-conduct-code/', '_self');
```
madcampos:
```suggestion
					window.location = '/pages/review-conduct-code/';
```

This also reloads the page.

src/components/CheckSteps/CheckSteps.tsx
```diff
@@ -0,0 +1,37 @@
...
+				<li className='card'>
+					<CheckConductCode />
+					<p className='text-h6'>Check the TorontoJS's conduct code</p>
+				</li>
+				<li className='card'>
+					<CompleteProfile />
+					<p className='text-h6'>Complete your profile</p>
+				</li>
+			</ul>
+			<Button
```
madcampos: **suggestion:** I know it is _super_ **super** counter intuitive to say that‏... But here maybe we would benefit from using a link tag instead of a button tag.

The reasoning for that is the logic of this page. If there is any required action by the user before navigating to the other page, like validation of some data, then yes, a button is the best option, but if it is just "navigating away to a new page", a link is the best option.

I don't like it _much_, but the most appropriate solution is to add a `ButtonLink` component, that does the same thing as a `Button` but uses the `<a>` tag instead and requires an `href`.

That will create al ink that _kinda looks like_ a button, but _behaves_ like a link. As long as we keep all of the _semantics_ of links and don't try to make it weird, we should be okay.

src/components/HeaderTitle/HeaderTitle.tsx
madcampos, line 1: **issue:** If the _only thing_ in a component‏ is one tag, then it should not be a component.

this makes the code hard to understand and more complex by adding one extra layer of abstraction/indirection where it is not needed. Both us when we have to read the code and react when it renders the page will have to follow one more thing that could very much be removed.

src/components/Header/Header.css
```diff
@@ -0,0 +1,6 @@
+header {
```
madcampos: **issue:** Same argument about namespacing here. This will apply to every `header` on the page. As it is the main element of the component, we should add an id or classname and reference that here instead.‏

src/components/Logo/Logo.tsx
```diff
@@ -1,7 +1,11 @@
 import './Logo.css';
 
-const Logo = () => (
-	<picture className='logo'>
+interface Props {
+	isLarge?: boolean;
+}
+
+const Logo = ({ isLarge = true }: Props) => (
+	<picture className={`logo ${isLarge ? 'large' : 'small'}`}>
```
madcampos, line 8: **question:** Why do we have two logo sizes? Are they from mobile and desktop versions of the page?

If it is mobile vs desktop, I think a media query would be better. 

If it is based on context, another trick would be to use [`data-*`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) attributes.‏

Yet another option is using [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries), if it makes sense.

If the problem is loading two diferent images, depending on the size, we should use [`srcset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset).

src/components/ReviewConductCode/ReviewConductCode.tsx
```diff
@@ -0,0 +1,111 @@
...
+								Lorem ipsum dolor sit amet consectetur adipisicing elit. Rem impedit consequatur, fugiat sit asperiores, sunt possimus delectus incidunt, iste atque
+								qui doloremque rerum. Ratione mollitia dicta corporis minus est neque.
+							</div>
+						</details>
+					</li>
+				</ul>
+
+				<label className='checkbox-label'>
+					<input
+						type='checkbox'
```
madcampos:
```suggestion
						type='checkbox'
						required
```
Mark the checkbox as required

src/components/ReviewConductCode/ReviewConductCode.tsx
```diff
@@ -0,0 +1,111 @@
...
+							setAgreementChecked(!agreementChecked);
+						}}
+					/>
+					I agree to TorontoJS’s conduct code and other forms
+				</label>
+
+				<Button
+					isPrimary
+					isLarge
+					disabled={!agreementChecked}
```
madcampos:
```suggestion
					type="submit"
```
We don't need extra state because the html form validation will already take care of that for us.

src/components/ReviewConductCode/ReviewConductCode.tsx
```diff
@@ -0,0 +1,111 @@
...
+					id='complete-profile-button'
+					onClick={() => {
+						window.open('/pages/complete-profile/', '_self');
+					}}
```
madcampos: Instead of adding an `onclick` event handler here, we can wrap the button and checkbox (and the rest of the markup if it makes sense) in a `form`, so clicking the button sends a `submit` event, and it can send us to the next page as well automatically.

src/components/ReviewConductCode/ReviewConductCode.css
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#arrow-down-icon { display: block; }
+
+#arrow-up-icon { display: none; }
+
+/* When details is open, swap visibility */
+details:open summary #arrow-down-icon { display: none; }
+
+details:open summary #arrow-up-icon { display: block; }
```
madcampos: **suggestion:** From the comment about arrows: here it makes more sense to use only one arrow and make it rotate instead of adding two.‏

src/components/ReviewConductCode/ReviewConductCode.css
```diff
@@ -0,0 +1,136 @@
...
+	display: flex;
+	font-size: var(--text-sm);
+	font-weight: var(--weight-semibold);
+	gap: var(--spacing-xs);
+}
+
+input[type='checkbox'] {
+	height: 16px;
+	width: 23px;
+}
```
madcampos: **nitpick:** Please us relative units for sizing things, like `rem`s‏

src/components/StepBar/StepBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+			</li>
+			<li className={`step-item ${step === 2 ? 'current active' : 'inactive'} ${step > 2 ? 'done' : ''}`}>
+				<span className='step-number'>{step > 2 ? <CheckIcon /> : '2'}</span>
+				{!isMobile || step === 2 ? <span className='step-text'>Check the conduct code</span> : ''}
+			</li>
+			<li className={`step-item ${step === 3 ? 'current active' : 'inactive'}`}>
+				<span className='step-number'>3</span>
+				{!isMobile || step === 3 ? <span className='step-text'>Complete your profile</span> : ''}
+			</li>
+		</ol>
```
madcampos: **suggestion:** Wrap this in a `nav` element with an `aria-label` of "Registration steps". That way it is marked as a navigation and also has a label telling screen reader users what the navigation is for.‏

src/components/StepBar/StepBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+	}, []);
+
+	return (
+		<ol className='step-bar'>
+			<li className={`step-item ${step === 1 ? 'current' : ''} inactive done`}>
+				<span className='step-number'>
+					<CheckIcon />
+				</span>
+				{!isMobile || step === 1 ? <span className='step-text'>Account confirmed</span> : ''}
+			</li>
```
madcampos: **question:** Why not use a loop here?‏

src/components/StepBar/StepBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+const StepBar = ({ step }: Props) => {
+	const [width, setWidth] = useState(window.innerWidth);
+	const isMobile = width <= 640;
+
+	useEffect(() => {
+		const handleResize = () => setWidth(window.innerWidth);
+		window.addEventListener('resize', handleResize);
+
+		return () => window.removeEventListener('resize', handleResize);
+	}, []);
```
madcampos: **issue:** This is provided by the browser and is never set by the component, don't use a state for that.‏

Also, this kind of check is much better handled by CSS. Hide or show parts of the markup based on media queries.

src/components/StepBar/StepBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+	return (
+		<ol className='step-bar'>
+			<li className={`step-item ${step === 1 ? 'current' : ''} inactive done`}>
+				<span className='step-number'>
+					<CheckIcon />
+				</span>
+				{!isMobile || step === 1 ? <span className='step-text'>Account confirmed</span> : ''}
```
madcampos: **issue:** It is an accessibility issue to not provide a proper label for each step. Instead of hiding the label completely, use the [screen reader only pattern](https://css-tricks.com/inclusively-hidden/).‏

src/components/StepBar/StepBar.css
```diff
@@ -0,0 +1,49 @@
...
+
+	&.current { width: 100%; }
+
+	&.active {
+		border-bottom: 4px solid var(--color-accent);
+
+		.step-number { color: var(--color-accent); }
+	}
+
+	&.inactive {
```
madcampos: **suggestion:** Use this as the default rules, and then apply overrides, based on the other states.‏

src/components/Button/Button.css
lylaminju: Good point! I created a new color in `index.css`.
https://github.com/torontojs/communityhub-fe/pull/37/commits/1798094155bf4792e5020cad87f361826b008c12

src/components/CheckSteps/CheckSteps.css
lylaminju: Oh, I think this is redundant. I removed it.

src/components/CheckSteps/CheckSteps.tsx
lylaminju: Makes sense! I created a new component `ButtonLink`

src/components/CheckSteps/CheckSteps.css
kbventures: @lylaminju why are we introducing new media queries to the design? For most other pages this has been sufficient.

src/components/CheckSteps/CheckSteps.css
lylaminju: @madcampos Do you mean not to use general class name like "card" because this css is applied to all components in the pages?

src/components/ReviewConductCode/ReviewConductCode.tsx
lylaminju: Yeah, I agree to use `type="submit"`, but is it possible to set the `disabled` true/false without the extra state?

madcampos: I think we don't need to disable the button, if any of the required fields is filled or checked, then it won't allow for submission

src/components/Logo/Logo.tsx
lylaminju: > Why do we have two logo sizes? Are they from mobile and desktop versions of the page?

For now, we should use a small logo only in this page, with both mobile and desktop versions

src/components/Button/Button.css
madcampos: We still need to check if the muted colour has enough contrast.

Here is a tool to check the contrast between two colours: https://coolors.co/contrast-checker/

lylaminju: It says the contrast is poor 😅 I don't know how to improve the contrast while keeping the figma design...
image

src/reset.css
lylaminju: Yeah I agree not to add multiple reset files. To avoid redundancy, I just renamed this file to `basic-css-reset.css` and we can resolve future conflicts if there are different parts from [Basic css reset PR](https://github.com/torontojs/communityhub-fe/pull/33)

src/components/Button/Button.css
madcampos: Hummm... So, what we can do is play around with the colors, the box besides the color hex code allows us to change it.

I would suggest using HSL/HSV or Lab to change the color. Basically they allow us to change the light/dark balance of the color without messing too much with that. RGB is not nice for those kinds of things.

But yeah, let's invite the designers to chime in...

cc @marianavrocha and @caducarvalho

src/reset.css
madcampos: Let's unblock that PR, so we can merge it back to this one.

src/components/CheckSteps/CheckSteps.css
lylaminju: @madcampos @kbventures Okay. I replaced media query with `max-width` of `100%`! 
But, I still have to use media query to modify the horizontal padding. I'm not sure if this is reasonable.

src/components/CheckSteps/icons/CheckConductCode.tsx
lylaminju: @madcampos Since the `ArrowDown.tsx` will be used in at least 2 pages (Review conduct code, complete profile form), I agree to move all icons to a "root" `icons` folder

madcampos (approved):

kbventures (approved):
Fantastic work @lylaminju 

Please merge into dev! :)

PR #38, opened 2025-04-08
Sign up from figma design

### LINKED ISSUE #28: Implement sign up from Figma design
We need to implement _02 - Sign up_, _03 - Sign up (filled)_, and _04 - Check your email_ from Figma design.

Image

Image

## Notes
- Start with the html structure, css comes second
- After the structure, it should connect to the backend for requests
- If you need any components, for now, stick to plain html and just wrap it to a dummy component

## Sub tasks

- [ ] Implement password strength check
- [ ] Add link back to sign in page
- [ ] Implement timer and interactions for re-sending email confirmation

kbventures:
DRAFT of sign-up design. 

image


src/components/Footer/Footer.tsx
```diff
@@ -45,7 +45,7 @@ const FooterNav = () => (
 							</defs>
 						</svg>
 
-						<span id='TorontoJS' className='sr-only'>TorontoJS</span>
+						{/* <span id='TorontoJS' className='sr-only'>TorontoJS</span> */}
```
madcampos: **nitpick:** ‏Please remove comments with unused code, let's keep things tidy. If we want the code back, we can always look at the git history 😉

src/components/Footer/Footer.tsx
```diff
@@ -69,7 +69,7 @@ const FooterNav = () => (
 							/>
 						</svg>
 
-						<span id='Youtube' className='sr-only'>YouTube</span>
+						{/* <span id='Youtube' className='sr-only'>YouTube</span> */}
```
madcampos: Same here

src/components/Footer/Footer.tsx
```diff
@@ -100,12 +100,14 @@ const FooterNav = () => (
...
-						<span
+						{
+							/* <span
 							id='Instagram'
 							className='sr-only'
 						>
 							Instagram
-						</span>
+						</span> */
+						}
```
madcampos: And here...

src/components/Footer/Footer.tsx
```diff
@@ -164,12 +168,14 @@ const FooterNav = () => (
...
-						<span
+						{
+							/* <span
 							id='LinkedIn'
 							className='sr-only'
 						>
 							LinkedIn
-						</span>
+						</span> */
+						}
```
madcampos: And here...

src/components/LoginForm/LoginForm.css
```diff
@@ -0,0 +1,72 @@
...
+
+.inputDim input {
+	background-color: #e5e5e5;
+	border: solid 1px #cccccc;
+	border-radius: var(--rounded-xs);
+	color: #999999;
+	height: 42px;
+	padding: var(--spacing-xs);
+	width: 100%;
+}
```
madcampos: **suggestion:** ‏CSS nesting has been supported natively for a while, how about we refactor the code to take advantage of that?

src/components/LoginForm/LoginForm.css
```diff
@@ -0,0 +1,72 @@
...
+#email-error {
+	display: none;
+	outline: solid red thin;
+}
+
+.login-form:has(#email-input:user-invalid) #email-error { display: block; }
+
+.red { color: var(--color-error); }
+
+.text-size { font-size: var(--text-xm); }
```
madcampos: **nitpick:** ‏Let's use a more descriptive class name.

src/components/LoginForm/LoginForm.tsx
madcampos, line 1: Again, in this file: please remove any unused and commented code.

src/components/LoginForm/LoginForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+					value={password}
+					onChange={handleChange}
+					placeholder='Your password'
+				/>
+			</div>
+			{password && (
+				<span className='passwordError'>
+					<div className=' text-size'>
+						<div>
+							<span className='red'>Password strength: {strengthLabels[strength || 0]}</span>
```
madcampos: **question:** ‏Why do we have password strength on the login form?

src/components/SignUp/SignUpForm.css
```diff
@@ -0,0 +1,112 @@
...
+.underline { text-decoration: underline; }
+
+.have-account { & p { text-align: center; } }
+
+.line {
+	background-color: #cccccc;
+	display: block;
+	height: 1px;
+	width: 100%;
+}
```
madcampos: **suggestion:** ‏If those utility classes are to be reused, declare them in a global file instead of duplicating them.

src/components/SignUp/SignUpForm.css
```diff
@@ -0,0 +1,112 @@
...
+
+.login-form:has(#email-input:user-invalid) #email-error { display: block; }
```
madcampos: **question:** ‏Why do we have login related classes in the signup CSS?

src/components/SignUp/SignUpForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+			<div className='input-block'>
+				<label className='block' htmlFor='password-input'>
+					Password:<span>REQUIRED</span>
+				</label>
+				<input
+					id='password-input'
+					type='password'
+					name='password'
+					value={password}
+					onChange={handleChange}
```
madcampos: **suggestion:** ‏This should be `onInput`, as for password strength, we want to give users instant feedback, instead of how it is done for other inputs where the error is shown after the user has moved away from the input field.

src/components/SignUp/SignUpForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+			<div className='input-block'>
+				<label className='block' htmlFor='password-input'>
+					Password:<span>REQUIRED</span>
+				</label>
+				<input
+					id='password-input'
+					type='password'
+					name='password'
+					value={password}
```
madcampos: **suggestion:** ‏As the event handler already gets the element as part of it's parameter, this state becomes redundant. We can simply use the value provided by the elements `value` property on the event handler to set the other states.

src/components/SignUp/SignUpForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+						</div>
+						<div id='password-input-suggestion' className='suggestion' data-password-strength={strengthLabels[strength || 0]}>
+							<span className='suggestion-icon' />
+							<p>Suggestions: {feedback}</p>
+						</div>
+					</div>
+				</div>
+			</div>
+
+			<Button type='submit' isLarge={true} style={{ color: 'white', background: '#ED343F' }} aria-label='Complete sign-up form button' disabled={isLoading}>
```
madcampos: **question:** ‏Why are we setting an inline style here?

madcampos:
```suggestion
			<Button type='submit' isLarge={true} style={{ color: 'white', background: '#ED343F' }} aria-disabled={isLoading}>
```

There is no need for using an `aria-label` here as it is redundant with the button text.

Also: setting the `disabled` attribute _hides_ the button from assistive technologies. Here the `aria-disabled` is better as it does not hide the button and also conveys to assistive technologies that it is disabled.

src/pages/home/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<title>Sign-In</title>
+	</head>
+	<body>
+		<div id="root"></div>
+		<script type="module" src="./home.tsx"></script>
```
madcampos: **question:** ‏Shouldn't the sign-in component be referenced here?

src/pages/home/home.tsx
madcampos, line 1: **nitpick:** Please remove the file if it is not being used.‏

src/pages/home/style.css
madcampos, line 1: **nitpick:** Please remove the file if it is not being used.‏

src/pages/sign-in/sign-in.tsx
```diff
@@ -0,0 +1,24 @@
...
+import '../../index.css';
+import './style.css';
+
+const root = document.getElementById('root') as HTMLDivElement;
+
+createRoot(root).render(
+	(
+		<StrictMode>
+			<div className='app'>
+				<div className='Sign-in-page'>
```
madcampos: **nitpick:** ‏Let's use either `snake-case` or `camelCase` for class names, not both. 🙃‏

madcampos: **question:** ‏Shouldn't those two divs be part of the login component?

src/pages/sign-in/style.css
```diff
@@ -0,0 +1,15 @@
+.app {
+	align-items: center;
+	display: flex;
+	flex-direction: column;
+	justify-items: center;
+}
```
madcampos: **question:** ‏Are we setting the same styles and using the same CSS for all pages? Why not make it global?

src/pages/sign-up/sign-up.tsx
```diff
@@ -0,0 +1,24 @@
...
+import '../../index.css';
+import './style.css';
+
+const root = document.getElementById('root') as HTMLDivElement;
+
+createRoot(root).render(
+	(
+		<StrictMode>
+			<div className='app'>
+				<div className='Sign-up-page'>
```
madcampos: Same here...

tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-	"include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.js", "src/**/*.jsx", "src/env.d.ts"],
+	"include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.js", "src/**/*.jsx", "src/env.d.ts", "src/vite-env.d.ts"],
```
madcampos: **question:** ‏This line already references a `src/env.d.ts` file, why are we adding another reference here?

vite-env.d.ts
madcampos, line 1: **question:** I think this file is in the wrong folder, shouldn't it be inside `src` and not on the root folder?‏

vite.config.ts
```diff
@@ -34,7 +34,8 @@ export default defineConfig(({ mode }) => {
 					'team': fileURLToPath(new URL('./src/pages/team/index.html', import.meta.url)),
 					'complete-profile': fileURLToPath(new URL('./src/pages/complete-profile/index.html', import.meta.url)),
 					'check-steps': fileURLToPath(new URL('./src/pages/check-steps/index.html', import.meta.url)),
-					'review-conduct-code': fileURLToPath(new URL('./src/pages/review-conduct-code/index.html', import.meta.url))
+					'review-conduct-code': fileURLToPath(new URL('./src/pages/review-conduct-code/index.html', import.meta.url)),
+					'sign-up': fileURLToPath(new URL('./src/pages/sign-in/index.html', import.meta.url))
```
madcampos: **thought:** ‏I think the other added pages like sign-in and home are missing here.

src/components/LoginForm/LoginForm.css
kbventures: I'm old enough to remember SASS.

vite-env.d.ts
kbventures: Config files usualy live in the root folder.  This isn't consiered a config file?

src/components/LoginForm/LoginForm.css
madcampos: Okay gramps LOL

I can't say anything though, I remember working with both less and sass... And layouts with tables and flash

vite-env.d.ts
madcampos: I'm not sure... It technically is, but it is also not as there are specifics related to the app itself...

I'd say it breaks the boundary of config vs code.

But the point is more that we do have two references on `tsconfig`:
- This file
- `src/env.d.ts`

I think we should pick one. I don't have any qualms to which and where it live, as long as we stick to one file in one place.

src/components/SignUp/SignUpForm.tsx
kbventures: I made the change but strangely the behavior of onChange in this instance has the "onInput" behavior.

PR #46, opened 2025-04-23
Home Page Initial Setup

### LINKED ISSUE #43: Create Home Page
This is the main page which you will be redirected to after sign-in

Ensure sign-in redirects to this page in development and production
Introduce /auth/heartbeat and /auth/sign-out 
Develop design as it is being rolled out

kbventures:
## Summary

Home Page(Post Sign-in)

Design in progress

#82 

- [x] Sign-in page succesfully fowards to home page
- [x] Sign-out butotn
- [x] Heartbeat check if authenticated and redirect if user is signed-out


kbventures:
This page was created to allow to to prepare for the Home page design implementation and enable e2e testing. 

This page also solves sub issue 70 

https://github.com/torontojs/communityhub-fe/issues/70

Of 

Make API calls from the frontend
[#67](https://github.com/torontojs/communityhub-fe/issues/67)

madcampos (changes requested):
All the reviews are here: https://github.com/torontojs/communityhub-fe/pull/79

manvinderjit (approved):

src/components/Home/Home.tsx
```diff
@@ -0,0 +1,72 @@
...
+			}
+		}
+	};
+
+	return (
+		<>
+			<nav>
+				<button type='button' onClick={handleLogOut}>Log Out button</button>
+			</nav>
+			<div>Is Authenticated: {authenticated.toString()}</div>
```
manvinderjit: I am assuming this is a placeholder for now.

src/utilities/getApiUrl.ts
```diff
@@ -1,3 +1,5 @@
 export function getApiUrl() {
-	return import.meta.env['APP_API_URL'] ?? 'https://community-hub.torontojs.com/';
+	const BE_URL = import.meta.env['APP_API_URL_DEV'] ?? 'https://community-hub.torontojs.com/';
```
manvinderjit: Will the other PR overwrite these changes because we don''t need backend and frontend URLs anymore?

madcampos: In this case, it will create a merge conflict as the other PR is based on this one.

But to the point of both URLs pointing to the same server there is this comment: https://github.com/torontojs/communityhub-fe/pull/79#discussion_r2108037892

To expand on that:

If we use only the paths for urls, like `/some/path` or `./some/path`, they will resolve to the current host for the page, that way we don't have to keep track of the host, it will for most of what we want just work.

src/components/Home/Home.tsx
kbventures: It is.  This is an initial step so we are able to test sign-in, heartbeat, log out and post sign-up flows.

src/utilities/getApiUrl.ts
kbventures: I modified back to having a single URL now that we are using a mono repo.

eslint.config.mjs
```diff
@@ -477,7 +477,7 @@ const languageOptions = {
 	}
 };
 
-const ignores = ['node_modules/**/*', 'dist/**/*', 'public/**/*', 'dev-dist/**/*', 'api/src/env.d.ts'];
+const ignores = ['node_modules/**/*', 'dist/**/*', 'public/**/*', 'dev-dist/**/*', 'src/content/**/*', 'src/env.d.ts'];
```
madcampos, line 480: **issue:** please revert this change as it will make eslint complain about the auto generated env file for the backend. Also we don't have a `content` folder, that is a leftover from copying the config from the blog repo.

src/components/Home/Home.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { handleLogOut } from '../../utilities/handleLogOut.ts';
+
+export const Home = () => {
+	useHeartBeat();
+	return (
+		<>
+			<nav>
+				<button type='button' onClick={handleLogOut}>Log Out button</button>
+			</nav>
+			<div>Home Page</div>
```
madcampos: **nitpick:** 
```suggestion
			<h1>Home Page</h1>
```

src/components/SignIn/SignInForm.tsx
```diff
@@ -109,7 +106,7 @@ const SignInForm = (): React.JSX.Element => {
 			</Button>
 
 			<div>
-				<a href={`${apiUrl}forgot-passsword`} className='do-not-remember-password'>I don't remember my password</a>
+				<a href='pages/forgot-passsword' className='do-not-remember-password'>I don't remember my password</a>
```
madcampos:
```suggestion
				<a href='/pages/forgot-passsword' className='do-not-remember-password'>I don't remember my password</a>
```

It is missing the leading slash, without it the url will be relative from the current url and not absolute.

PR #48, opened 2025-04-29
create [check your e-mail] page

### LINKED ISSUE #47: Create [Check your e-mail] Page
Implement a page of "Check your e-mail"

Image

lylaminju:
## Summary

- removed duplicated `basic-css-reset.css` file and unused `system-design.css` file
- created a page "Check your e-mail"

## Tests

http://localhost:3000/pages/check-your-email/

For a quick test of timer functionality, set `RESEND_MINUTES` to small number like 3, and the interval to `1 * 1000`

### Issue

- CSS [opacity 50%](https://github.com/torontojs/communityhub-fe/pull/48/files#diff-adb7179bce171b2307e7c9bcf2b6e8493c20cd541af22337b16de0be93016fcfR41) : This can reduce accessibility. We should discuss other color values


src/components/CheckYourEmail/CheckYourEmail.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+	return (
+		<div className='check-email-page'>
+			<Header>Check your e-mail</Header>
+			<main className='check-email-main'>
+				<p>
+					We&apos;ve sent a message to your e-mail <b>{email}</b> (Is it wrong?{' '}
+					<a href='#' className='change-email'>
+						Change it
+					</a>
```
madcampos: **issue:** This should be a button instead of a link. We can check with the design team to also update the design.‏

src/components/CheckYourEmail/CheckYourEmail.tsx
```diff
@@ -0,0 +1,69 @@
...
+		const timer = setInterval(() => {
+			setMinutes((prevMinutes) => {
+				if (prevMinutes <= 0) {
+					clearInterval(timer);
+					setHasResent(false);
+					return 0;
+				}
+				return prevMinutes - 1;
+			});
+		}, 60 * 1000);
```
madcampos: **suggestion:** Let's set this number to a variable so it is more readable what it means‏

src/components/CheckYourEmail/CheckYourEmail.tsx
```diff
@@ -0,0 +1,69 @@
...
+						<a
+							href='#'
+							className='resend-link'
+							onClick={() => {
+								setHasResent(true);
+								setMinutes(RESEND_MINUTES);
+							}}
+						>
+							Re-send activation link
+						</a>
```
madcampos: **suggestion:** ‏I think this should also be a button... More specifically, a submit button for a form.

The reasoning here is that (re)sending an activation triggers an action to the server, this should be done through a form.

lylaminju: Yeah, I agree that the button is more reasonable in this context. 
We should also discuss with the design team to update the design

src/css-layers/reset.css
```diff
@@ -11,7 +11,7 @@
 	}
 
 	/* --- Remove Default Styling for Lists --- */
-	nav :is(ul, ol) { list-style: none; }
+	ul, ol { list-style: none; }
```
madcampos: **suggestion:** Let's keep the style as it was before.‏

The reasoning here is we want to keep the default style for lists unchanged, so all lists by default have the list markers.

The exception is when we follow the pattern of creating a `nav` with a list inside, as this is a common pattern for creating navigation menus, and usually on those cases the list gets changed anyways.

Here is an example of a navigation menu:
```html
<nav>
  <ul>
    <li><a href="/">Home</a></li>
    <li><a href="/blog/">Blog</a></li>
    <li><a href="/contact/">Contact</a></li>
  </ul>
</nav>
```

src/pages/check-your-email/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<title>Check your e-mail</title>
+		<link rel="icon" type="image/png" href="/favicon.png" />
```
madcampos, line 7: **thought:** Just a small trivia/thing to think about: https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs‏

No need to change anything here!

src/components/CheckYourEmail/CheckYourEmail.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+		return () => clearInterval(timer);
+	}, [hasResent]);
+
+	return (
+		<div className='check-email-page'>
+			<Header>Check your e-mail</Header>
+			<main className='check-email-main'>
+				<p>
+					We&apos;ve sent a message to your e-mail <b>{email}</b> (Is it wrong?{' '}
```
madcampos: **suggestion:** I think we can skip most of the entity encodings...‏

Most of the entity encodings are a relic of XML, wich is like if HTML decided to behave like JS and if it have a missing semicolon the whole code breaks.

HTML 5 is very forgiving in how it parses content and tries it's best to make things work nicely.

In more modern HTML, we only strictly need to encode one thing: `<`. This is because the character opens and closes a tag and can mess up everything following it, specially when we are in fact trying to reference an html tag.

E.g.:

```markdown
This markdown will encode the character: `<div>`
```

The other "needed" encodings are situational:
- Quotes are needed when you want to add a quote character inside an HTML attribute[^1], but it is not needed in HTML text.
  ```html
  <abbr title="The quote character: &quot;">"</abbr>
  <abbr title="The apostrophe character: '">'</abbr>
  <abbr title='Another quote character without escaping: "'>Another "</abbr>
  ```
- The ampersand character (`&`) when it is [followed by anything that is not a space](https://html.spec.whatwg.org/multipage/syntax.html#syntax-ambiguous-ampersand).

[^1]: The special case here is that attributes in HTML [can be quoted either with single or double quotes](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2), so if you use one, you don't need to escape the other character inside the quotes. An extra bonus, HTML attributes [can be unquoted](https://html.spec.whatwg.org/multipage/syntax.html#unquoted), but that is a _very, very_ edge case.

madcampos (approved):

kbventures (approved):

PR #54, opened 2025-05-05
feat: handle complete profile form submission

### LINKED ISSUE #53: Complete profile form submission handling

sounmind:
Co-authored-by: Manvinderjit Rupal <manvinderjit@users.noreply.github.com>## Summary

Add the Complete Profile form submission function

## Tests

Tested manually. Pending unit, integration, e2e tests.

## Additional information

User must be logged in to access the page.


manvinderjit:
@madcampos @kbventures I have completed the form submission function. It does successfully send the data to the backend. 

Some refinement is required like improved error handling and showing error messages. It also needs to be integrated with the remaining completion flow.

It will not run online because it requires a cookie to be accessed. Must run locally.

madcampos:
@manvinderjit what is the thing with cookies? Could you expand on that?

src/utilities/getApiUrl.ts
```diff
@@ -1,3 +1,3 @@
 export function getApiUrl() {
-	return import.meta.env['APP_API_URL'] ?? 'https://community-hub.torontojs.com/';
+	return import.meta.env['APP_API_URL'] as string ?? 'https://community-hub.torontojs.com/';
```
madcampos: **question:** The null check (`??`) should already handle the case where the variable doesn't exist., so we need the extra type casting?

Env vars are always `string`s or `undefined`. The extra case we could handle is an empty string

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -16,30 +17,98 @@ interface SocialIcons {
...
+	skills: string[];
 }
 
+// Validation helpers
+const isValidBirthdayFormat = (birthday: string): boolean => {
+	const regex = /^(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])$/;
+	return regex.test(birthday);
+};
+
+const isRealDate = (birthday: string): boolean => {
```
madcampos, line 66: **nitpick:** ‏
```suggestion
/**
 * @deprecated We will remove update this when Temporal becomes stable
 */
const isRealDate = (birthday: string): boolean => {
```

I think this will change once we have better support for Temporal, so let's mark it as such.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -16,30 +17,98 @@ interface SocialIcons {
...
+	links: { platform: string, url: string }[] | [];
+	pronouns?: string;
+	socialAccounts?: string[];
+	birthday: string;
+	avatar?: string;
+	skills: string[];
 }
 
+// Validation helpers
+const isValidBirthdayFormat = (birthday: string): boolean => {
```
madcampos, line 49: **suggestion:** I think this function would benefit from some documentation...

Documentation for Typescript and Javascript follow the [JSDoc](https://jsdoc.app/) format.
In [TS](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) case it can be simplified as the types for parameters and function return are already provided by the language.

```suggestion
/**
 * <Insert docs here>
 */
const isValidBirthdayFormat = (birthday: string): boolean => {
```

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -16,30 +17,98 @@ interface SocialIcons {
...
+
+	const [monthStr, dayStr] = birthday.split('-');
+	const month = Number(monthStr);
+	const day = Number(dayStr);
+
+	const date = new Date(2000, month - 1, day);
+	return date.getMonth() + 1 === month && date.getDate() === day;
+};
+
+const updateProfile = async (args: UpdateProfileParams, profileId: string) => {
```
madcampos: **nitpick:** ‏
```suggestion
const updateProfile = async (data: UpdateProfileParams, profileId: string) => {
```

This is the data to be processed, so it is better to name it `data` or `body`, or other descriptive name.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -16,30 +17,98 @@ interface SocialIcons {
...
+	try {
+		const response = await fetch(`/api/profiles/${profileId}`, {
+			method: 'PATCH',
+			headers: {
+				'Content-Type': 'application/json'
+			},
+			body: JSON.stringify({ ...args }),
+			credentials: 'include'
+		});
+		if (!response.ok) {
```
madcampos: **issue:** Please confirm the status code for this request, if it is not a 200 - OK, this will fail.‏

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -16,30 +17,98 @@ interface SocialIcons {
...
+			const errorData = await response.json();
+			console.log('Response not ok: ', errorData);
```
madcampos: [Emo log](https://www.youtube.com/watch?v=dhZTNgAs4Fc)

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -16,30 +17,98 @@ interface SocialIcons {
...
-	const [skills, setSkills] = useState<Skill[]>([]);
 	const [socialIcons, setSocialIcons] = useState<SocialIcons[]>([
 		{ name: 'Instagram', element: <Instagram />, visible: true },
 		{ name: 'Facebook', element: <Facebook />, visible: true },
 		{ name: 'Threads', element: <Threads />, visible: true },
-		{ name: 'LinkedIn', element: <LinkedIn />, visible: true },
+		{ name: 'Mastodon', element: <Mastodon />, visible: true },
 		{ name: 'BlueSky', element: <BlueSky />, visible: true },
-		{ name: 'X', element: <XTwitter />, visible: true },
+		{ name: 'Twitter', element: <XTwitter />, visible: true },
```
madcampos:
```suggestion
		{ name: 'X/Twitter', element: <XTwitter />, visible: true },
```

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -57,11 +126,16 @@ const CompleteProfile = () => {
...
-			if (skillInputRef.current) { skillInputRef.current.value = ''; }
+		const index = profileData.skills.findIndex((skill) => skill.trim().toLocaleLowerCase() === skillName.trim().toLocaleLowerCase());
+		if (skillName.trim() !== '' && index == -1) {
+			setProfileData((prev) => {
+				return {
+					...prev,
+					skills: [...prev.skills, skillName]
+				};
+			});
 		}
```
madcampos: **suggestion:** ‏
```suggestion
		setProfileData((prev) => {
			return {
				...prev,
				skills: [...new Set([...prev.skills, skillName.toLowerCase()])]
			};
		});
```

A [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) guarantees each item is unique. A neat trick to filter an array of primitive values is to convert it to a set and then back to an array like so:

```typescript
[...new Set(originalArray)]
```

The Set constructor receives an iterable, like an array; Then we use the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) to iterate over all items of the set; and lastly wrap it in an array literal to convert back to an array.

madcampos: Side note: let's save all skills as lowercase strings to make it easier to compare

madcampos: Side note: only use locale aware conversions if you are in control of the locale being used. If you let the browser do it, unexpected results may occur as the case rules for different locales are different.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
+				updatedLinks = [...prev.links, { platform, url: trimmedValue }];
+			} else {
+				return prev;
+			}
+
+			return {
+				...prev,
+				links: updatedLinks
+			};
+		});
```
madcampos:
```suggestion
		if (trimmedValue) {
			setProfileData((prev) => {
				const existingLinkIndex = prev.links.findIndex((link) => link.platform === platform);
				const updatedLinks = prev.links;
				
				if (existingLinkIndex !== -1) {
					updatedLinks.splice(existingLinkIndex, 1);
				}

				return {
					...prev,
					links: [...updatedLinks, {
						platform,
						value: trimmedValue
					}]
				};
			});
		}
```

1. Only update the value if there is anything to update
2. Using [`Array.findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) and [`Array.splice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) will remove the item from the array while preserving the order of items if needed. (Another option is to just filter the item out).
3. The spread operator + the item is an easy shortcut to appending items to an array. Don't abuse it though as it will iterate over all items and then append the new one.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
+
+	const handleSliderToggle = (e: ChangeEvent<HTMLInputElement>) => {
+		const { name, checked } = e.target;
+		setProfileData((prev) => ({
+			...prev,
+			[name]: checked
+		}));
+	};
+
+	const handleSubmit: FormEventHandler<HTMLFormElement> = async (event) => {
```
madcampos, line 249: **nitpick:** If we are already handling the form's `submit` event, in theory we don't have to handle any other `input` or `change` events.‏

I figure we still have to do it though because react is a controling and gaslighting little bastard that freaks out if things change without it knowing.

So yeah, two options here:
- Be controlling and have a state variable for every single thing.
- Add `ref`s to every input so react doesn't destroy them and leave us alone.

madcampos: Curveball: when we wrap the inputs into a component, we should keep `ref`s to them. That way we have less work to do regarding react (i.e. we won't need to have state for every single piece of data) and form submitions will be a matter of creating a new `FormData` for the form element...

The only odd thing here will be "custom" inputs like the skills list or birthday. But that can be handled on a case by case basis.

manvinderjit:
> @manvinderjit what is the thing with cookies? Could you expand on that?

@madcampos You will have to  generate a cookie by signing in to be able to access the Complete Profile page.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
-														type='button'
-														aria-label='Remove Skill'
-														onClick={() => handleRemoveSkill(skill.id)}
-													>
-														x
-													</button>
-												</span>
-											))}
+		isLoading ?
+			<p>Loading...</p> :
```
madcampos: **nitpick:** Please prefer an early return...‏

```typescript
if (isLoading) {
    return <p>Loading...</p>;
}

return ...
```

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
+					...fetchedFields
+				}));
+
+				setProfileId(data.data.id);
+				setIsLoading(false);
+			})
+			.catch((_error) => {
+				// setErrorMessage(error.message);
+				setIsLoading(false);
+			});
```
madcampos: **suggestion:** Use an async [IIFE‏](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) instead of chaining `then`/`catch` calls.

```suggestion
		(async () => {
			setIsLoading(true);
			
			try {
				const response = await fetch(`/api/profiles/self`, {
					method: 'GET',
					headers: {
						'Content-Type': 'application/json'
					},
					credentials: 'include'
				});
				
				// If not authenticated, redirect to Sign In page
				if (response.status === 401) {
					window.location.href = `/pages/sign-in/`;
				} else if (!response.ok) {
					throw new Error('Network response was not ok');
				}

				const data = await response.json();
				const fetchedFields = {
					id: data.data?.id,
					name: data.data?.name,
					email: data.data?.email,
					isBasedOnGTA: data.data?.isBasedOnGTA,
					canJoinLocalEvents: data.data?.canJoinLocalEvents
				};
			
				setProfileData((prev) => ({
					...prev,
					...fetchedFields
				}));

				setProfileId(data.data.id);
			} catch (err) {
				// setErrorMessage(error.message);

			} finally {
				setIsLoading(false);
			}
		})();
```

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
+
+	useEffect(() => {
+		// Update profileData only if new date is valid
+		if (isRealDate(birthdayValue)) {
+			setProfileData((prev) => ({
+				...prev,
+				birthday: birthdayValue
+			}));
+		}
+	}, [birthdayValue]);
```
madcampos, line 326: **issue:** ‏I think this is unnecessary as we are already updating the data inside an event handler.‏

The `useEffect` hook in react is recomended for async work or things that affect react from the outside in. User input, if we have event handlers, covers this case.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
+				<>
+					<StepBar
+						currentStep={3}
+						steps={[
+							{ label: 'Account confirmed' },
+							{ label: 'Check the conduct code' },
+							{ label: 'Complete your profile' }
+						]}
+					/>
+					<form onSubmit={handleSubmit} encType='multipart/form-data' id='complete-profile-form'>
```
madcampos: **thought:** We should add support on the server for regular form data instead of only json‏

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
+												<label htmlFor='canJoinLocalEvents'>
+													<span>I can join TorontoJS&#8217; local events</span>
+												</label>
+											</div>
+										</div>
+									</div>
 								</div>
+							</details>
+
+							{/* TODO: Need to confirm adding green check icon on successful section completion on mobile view */}
```
madcampos: **question:** Is this confirmed yet?‏

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
-													aria-label={`Close ${socialInput} input`}
-													onClick={() => handleRemoveSocialAccount(socialInput)}
-												/>
-											</span>
+											<label htmlFor='linkedInProfile'>LinkedIn profile</label>
 											<input
-												id={`${socialInput}-input`}
-												name={socialInput}
+												id='linkedInProfile'
+												name='linkedInProfile'
```
madcampos: **nitpick:** ‏Please use `kebab-case` for html attribute values that are used cross languages (html, js, css)‏

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -92,327 +175,512 @@ const CompleteProfile = () => {
...
+		}
+	};
+
+	const linksByPlatform = useMemo(() => {
+		const map: Record<string, string> = {};
+		profileData.links.forEach(({ platform, url }) => {
+			map[platform] = url;
+		});
+		return map;
+	}, [profileData.links]);
```
madcampos: **suggestion:** this should be the way the data is represented here and only get transformed to an array when the data is submited to the server.‏

As a general tip, if the data don't quite fit the expectations of your "unit of code" (function, class, component, api endpoint, etc), transform it on input and then transform it back on the output. As long as the _contract_ between your unit of code and other units remains the same, you can do whatever with the data _inside_ your unit of code.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -16,30 +17,98 @@ interface SocialIcons {
...
 		{ name: 'Facebook', element: <Facebook />, visible: true },
 		{ name: 'Threads', element: <Threads />, visible: true },
-		{ name: 'LinkedIn', element: <LinkedIn />, visible: true },
+		{ name: 'Mastodon', element: <Mastodon />, visible: true },
 		{ name: 'BlueSky', element: <BlueSky />, visible: true },
-		{ name: 'X', element: <XTwitter />, visible: true },
+		{ name: 'Twitter', element: <XTwitter />, visible: true },
 		{ name: 'Dev.to', element: <DevTo />, visible: true }
 	]);
 	const [socialAccountInputs, setSocialAccountInputs] = useState<string[]>([]);
```
madcampos: See this example of how to use inputs with the new [invoker proposal](https://open-ui.org/components/invokers.explainer/): https://codepen.io/madcampos/pen/RNPGRoQ?editors=0011

It is working on chromium based browsers but also includes a basic pollyfill for other browsers.

Once support is available across the board, we can use it (mostly) without js.

src/utilities/getApiUrl.ts
manvinderjit: This is outdated, the file is now removed.

madcampos:
> > @manvinderjit what is the thing with cookies? Could you expand on that?
> 
> @madcampos You will have to generate a cookie by signing in to be able to access the Complete Profile page.

We had some brief conversations before about having the profile in different states/keeping flags for the steps of the profile... Maybe it is a good opportunity to revisit that.

manvinderjit:
> > > @manvinderjit what is the thing with cookies? Could you expand on that?
> > 
> > 
> > @madcampos You will have to generate a cookie by signing in to be able to access the Complete Profile page.
> 
> We had some brief conversations before about having the profile in different states/keeping flags for the steps of the profile... Maybe it is a good opportunity to revisit that.

Yeah, that is what is on the table. It is required for the integration that I am going to do next.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Every time I bring this up, the conversation gets sidetracked. :smile: 

https://torontojs.slack.com/archives/C081HT0SGMP/p1747333496124309

I am gonna skip it for now I think. :laughing:

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Console.log is present because we are waiting on design updates for error message component. Will update this once we have that available.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Good to know about the usage of set. Thanks for sharing.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @madcampos I am updating the value when its empty because I want to remove the value if a user empties out the input field after initially typing. Otherwise the input will look like { platform: "LinkedIn", url: "" }.
This removes the respective object from the links array altogether.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: The Complete Profile is a really complex page
1. We are fetching some data from the backend, like email, name.
2. We are adding location data with radio buttons, aka sliders.
3. We are adding skills data , which is an array of strings.
4. We are adding links, which are an object that must follow an enum key and a string value pattern.
5. The page will have two separate forms, one for updating profile, and another for updating image.

Therefore I focused on handling state segregation for clean code.

manvinderjit: A single update function would have become really big, making things complex and code unreadable.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Will look into this after we complete file upload.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: This is primarily to get the value of the `links` based on `platform` to show it in `HTML inputs`. It doesn't affect the data in the state or for the request.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: So I am using two states in the page
1.`birthdayValue` which stores the value of the birthday, it can have half values like `MM- `or `-DD` because it is waiting for the user to select both values. It is set using the handle function for either of the `selects` for MM or DD.
2. `profileData` is the main state that stores data fetched or sent to backend. It has a `birthday` value that stores the data to be sent to backend. It is updated using this useEffect if `birthdayValue` is valid. 

I can club both functions into the first function if you think that makes sense.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @madcampos  Do we implement this now or once it is supported across the board?

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: image

Was keeping this aligned with the backend enum as shown in the image.
1. Does zod enum allow `X/Twitter`? 
2. Do you recommend using lower case for such enums, like platform name?

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: Up to you 🤷‍♂️

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: Created an issue for that: https://github.com/torontojs/communityhub/issues/144

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: Was just a silly comment to link to an emo music video 🙊

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: Ah, on this case, yes. Let's keep two things:
- "id", lowercase, agnostic name
- "display name", can be whatever, have formatting, used for display

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: That is okay to have a sparse array of items. But then again: if we have a defined set of items, why not save them as an object with the format:
```typescript
Record<PlatformName = string, Url = string>
```

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: That is the thing, the form doesn't need to be big and complex.

For most of the data we only need to care about it when the form is submitted. It doesn't matter what the user types in until then. We can just let the browser handle all this state for us and only care about the more complex stuff.

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: We can definitely move the date selector to a component to make it easier to handle. Also, it can do all of the required magic for disabling dates based on the month, like February disabling days 30 and 31.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -1,10 +1,11 @@
-import { useRef, useState } from 'react';
+/* eslint-disable max-lines */
+import { type ChangeEvent, type FormEventHandler, useEffect, useMemo, useRef, useState } from 'react';
 import Button from '../Button/Button.tsx';
 import { BlueSky } from '../Icons/Social/BlueSky.tsx';
 import { DevTo } from '../Icons/Social/DevTo.tsx';
 import { Facebook } from '../Icons/Social/Facebook.tsx';
 import { Instagram } from '../Icons/Social/Instagram.tsx';
-import { LinkedIn } from '../Icons/Social/LinkedIn.tsx';
+import { Mastodon } from '../Icons/Social/Mastodon.tsx';
```
madcampos, line 8: **question:** did we remove linkedin and just added mastodon here?‏

manvinderjit: @madcampos Negative. We have both. For LinkedIn we have a full length input. It was also being repeated in the optional account icons (which must be clicked upon to open their corresponding input) which should have been Mastodon instead.

image

manvinderjit:
@madcampos I have for now removed most state updates that were not required. I am creating a separate issue for Input components because we have a figma design from the design team and it will be better to do it separately to not clutter this branch. I will do the same for the birthday component.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: I will probably leave it for later, once i refactor all the code and create smaller components.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Will move it to into a separate component once i am done with the rest of refactoring and created the input component.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -1,44 +1,154 @@
...
+	name?: string;
+	links: { platform: string, url: string }[] | [];
+	pronouns?: string;
+	socialAccounts?: string[];
+	birthday: string;
+	avatar?: string;
+	skills: string[];
 }
 
+interface UpdateProfileParams extends Omit<ProfileParams, 'email'> {}
```
madcampos: **nitpick**:
```suggestion
type UpdateProfileParams = Omit<ProfileParams, 'email'>;
```

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -244,17 +496,18 @@ const CompleteProfile = () => {
 											name='canJoinLocalEvents'
 											type='checkbox'
 											className='slider-checkbox'
+											checked={profileData.canJoinLocalEvents}
+											onChange={handleSliderToggle}
 										/>
 										<label htmlFor='canJoinLocalEvents'>
-											<span>I can join TorontoJS's local events</span>
+											<span>I can join TorontoJS&#8217; local events</span>
```
madcampos: **nitpick**:
```suggestion
											<span>I can join TorontoJS&apos; local events</span>
```

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: We are not manipulating this state anymore with onChange.

manvinderjit:
@madcampos I have applied the suggested changes. You will have to re-approve. Can you please do me a favor and run this locally with the new seeddata you have created and check if it works? I still use the old seed data.

api/src/routes/profile/validation.ts
```diff
@@ -1,7 +1,7 @@
 import { z } from 'zod';
 import { BaseDbEntitySchema, BaseDBFieldsToOmit } from '../../utils/db.ts';
 
-export const PlatformEnum = z.enum(['Slack', 'LinkedIn', 'GitHub', 'Portfolio', 'Instagram', 'Threads', 'Facebook', 'BlueSky', 'Mastadon', 'Twitter', 'Dev']);
+export const PlatformEnum = z.enum(['slack', 'linkedin', 'github', 'portfolio', 'instagram', 'threads', 'facebook', 'bluesky', 'mastodon', 'xtwitter', 'dev']);
```
kbventures, line 4: Question: Is this a Linter changing the casing, or is this on purpose?

src/components/CompleteProfile/CompleteProfile.css
```diff
@@ -258,7 +258,7 @@
 		grid-template-columns: repeat(3, 1fr);
 		justify-content: space-between;
 
-		label:is([for='linkedInProfile'], [for='githubProfile'], [for='portfolio']) {
+		label:is([for='linkedin'], [for='github'], [for='portfolio']) {
```
kbventures, line 261: Question: same as above

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -1,44 +1,154 @@
...
 }
 
-interface Skill {
-	id: string;
-	name: string;
+interface ProfileParams {
+	email: string;
+	isBasedOnGTA: boolean;
+	canJoinLocalEvents: boolean;
+	name?: string;
```
kbventures, line 25: Question: Why is this optional?

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -1,44 +1,154 @@
...
+
+	const [monthStr, dayStr] = birthday.split('-');
+	const month = Number(monthStr);
+	const day = Number(dayStr);
+
+	const date = new Date(2000, month - 1, day);
+	return date.getMonth() + 1 === month && date.getDate() === day;
+};
+
+const updateProfile = async (data: UpdateProfileParams, profileId: string) => {
```
kbventures, line 77: Question: I suspect updatingProfile will surely be re-used, so why should we not move this to the hook folder?

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -73,28 +185,150 @@ const CompleteProfile = () => {
...
-		setSocialIcons((prevIcons) => prevIcons.map((icon) => icon.name === iconName ? { ...icon, visible: !icon.visible } : icon));
+	const toggleSocialInputVisibility = (inputId: string) => {
+		setSocialIcons((prevIcons) => prevIcons.map((input) => input.id === inputId ? { ...input, inputVisible: !input.inputVisible } : input));
 	};
 
-	const handleAddSocialAccount = (socialAccountToAdd: string) => {
-		setSocialAccountInputs((previousAccounts) => {
-			if (!previousAccounts?.includes(socialAccountToAdd)) {
-				return [...previousAccounts, socialAccountToAdd];
+	const getProfileParams = (formData: FormData): UpdateProfileParams => {
```
kbventures, line 204: Question: I suspect getProfileParams will surely be re-used, so why should we not move this to the hook folder?

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -1,44 +1,154 @@
...
+		}
+	} catch (e) {
+		if (import.meta.env.MODE === 'development') {
+			if (e instanceof Error) {
+				console.error(e.name);
+				console.error(e.cause);
+				console.error(e.message);
+				console.error(e.stack);
+			} else {
+				throw new Error('Error Encountered!');
```
kbventures: This error message is very generic.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -73,28 +185,150 @@ const CompleteProfile = () => {
...
+				const response = await fetch(`/api/profiles/self`, {
+					method: 'GET',
+					headers: {
+						'Content-Type': 'application/json'
+					},
+					credentials: 'include'
+				});
+
+				// If not authenticated, redirect to Sign In page
+				if (response.status === 401) {
```
kbventures, line 289: Note: I know this is probably the way it is, because we're waiting for this issue to be handled.

https://github.com/torontojs/communityhub/issues/142

Once this PR is approved, please create an issue to utilize the heartbeat(it may need some redesign) hook to check if the person who is accessing the page has the correct access / state for this page.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -1,44 +1,154 @@
...
+ * Checks whether a birthday string is in the valid MM-DD format.
+ *
+ * A valid format must follow the pattern: two-digit month (01–12)
+ * followed by a dash and a two-digit day (01–31).
+ * This function does not validate actual calendar dates (e.g., "02-30" would be considered valid).
+ *
+ * @param birthday - The birthday string to validate, expected in MM-DD format.
+ * @returns `true` if the format is valid, otherwise `false`.
+ */
+const isValidBirthdayFormat = (birthday: string): boolean => {
```
kbventures, line 49: @madcampos @manvinderjit 

Discussion: Perhaps they don't fit the bill of utilities that belong in the utility directory, but this page is large.  Would it not make sense to move isValidBirthdayFormat, isRealDate, and updateProfile in another folder?

kbventures (approved):
@manvinderjit left some comments and discussion items, but overall, it looks great. 

Please create an issue to implement changes from https://github.com/torontojs/communityhub/issues/142 into the complete profile page once it's completed, and another to create API tests(hurl) .

api/src/routes/profile/validation.ts
manvinderjit: This is on purpose, initially i had used Camel case but Marco asked this to be converted to lowercase for uniformity.

src/components/CompleteProfile/CompleteProfile.css
manvinderjit: Same answer as above, for uniformity. We don't need to change fields when sending data from frontend to api.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @kbventures They will be moved along with the Date Component. I am working on it already.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: We need to authenticate the user and have created the `profiles/self` route for it. This will be in addition to the correct access/state for this page. We will need two check at least, one for authentication other for authorization.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: I was going by the Figma design. It lists name as optional. But it is a readonly field and this page won't update a user's name, only the following params
image

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: This is only for dev mode, and it is a fallback error. Ideally it won't trigger.
image

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Yeah, will definitely do once we have all the integration complete.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: For this we will be updating different profile params for different situations. E.g., For activation we are allowing updating of params shown in image below. It doesn't have  name, email, etc. For edit profile we will be including that. So it will be situational.
image

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit:
```suggestion
				console.log(e);
				throw new Error('Unknown Error Encountered!');
```
@kbventures  I can do the above for more clarity if required.

src/components/CompleteProfile/CompleteProfile.tsx
kbventures: Please create issue(s) so we can keep track of this.

kbventures (dismissed):

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @kbventures The `self` route for authentication is already implemented. And authorization will be based on the state that Lyla is working on right now.

src/components/CompleteProfile/CompleteProfile.tsx
kbventures: I mean "Error Encountered" could utilize words that help you figure out where the error is generated.

madcampos: +1 here. If the error is not an instance of error, throw that, like this:
```js
if (err instanceof Error) {
  //...
} else {
  throw new Error(err);
}
```

madcampos: Unrelated to that: how is the linting passing with a variable called `e`? We have a specific rule that forbids variable names lesser than 3 characters long.

manvinderjit: Need to evaluate this further. Other pages are also allowing shorter error variables.
image

manvinderjit: error message is now updated.

manvinderjit:
> @manvinderjit left some comments and discussion items, but overall, it looks great.
> 
> Please create an issue to implement changes from #142 into the complete profile page once it's completed, and another to create API tests(hurl) .

@kbventures Both issues are now created.

kbventures (approved):
Great work

PR #73, opened 2025-05-09
style: update Button component styles

### LINKED ISSUE #57: Add an optional ref prop to Button component
Add an optional `ref` prop that can be passed to the Button component.
Update the typescript types to include the ref prop.

manvinderjit:
## Summary

Update Button component styles for updated design changes. Add a ref prop to the Button component.
Create a separate file for showing sample buttons.

## Tests

No written tests. Visual verification only

## Additional information

May have to update other components that use the Button component for consistency with the Figma design.

image

### DIFF
--- a/src/components/Button/Button.css
+++ b/src/components/Button/Button.css
@@ -1,32 +1,35 @@
 .button {
-	background-color: var(--color-button);
+	background-color: var(--color-foreground);
 	border: 0;
 	border-radius: var(--rounded-md);
 	box-sizing: border-box;
-	color: var(var(--color-black));
+	color: var(--color-background);
 	cursor: pointer;
 	font-weight: var(--weight-bold);
-	height: 59px;
-	margin: 8px 0;
 	text-align: center;
 	width: 100%;
 
 	&.small {
 		font-size: var(--text-xs);
+		height: 2.625rem;
 		padding: var(--spacing-xs) var(--spacing-sm);
-		/* width: 19.375rem; */
 	}
 	&.large {
 		font-size: var(--text-lg);
 		height: 3.6875rem;
-		/* width: 23rem; */
-		padding: 12px 16px;
+		padding: 0.75rem 1rem;
 	}
 
 	&.outline {
 		background-color: transparent;
 		border: var(--border-base) solid var(--color-black);
 		color: var(--color-black);
+
+		&:disabled,
+		&[aria-disabled='true'] {
+			background-color: transparent;
+			border: var(--border-base) solid var(--color-grey);
+		}
 	}
 
 	&.primary {
@@ -36,6 +39,11 @@
 
 		&:disabled,
 		&[aria-disabled='true'] { color: var(--color-black); }
+
+		&.outline {
+			background-color: transparent;
+			color: var(--color-accent);
+		}
 	}
 
 	&:disabled,
--- a/src/components/Button/Button.css
+++ b/src/components/Button/Button.css
@@ -9,18 +9,19 @@
 	text-align: center;
 	width: 100%;
 
-	&.small {
+	&[data-size='small'] {
 		font-size: var(--text-xs);
 		height: 2.625rem;
 		padding: var(--spacing-xs) var(--spacing-sm);
 	}
-	&.large {
+
+	&[data-size='large'] {
 		font-size: var(--text-lg);
 		height: 3.6875rem;
 		padding: 0.75rem 1rem;
 	}
 
-	&.outline {
+	&[data-outline='true'] {
 		background-color: transparent;
 		border: var(--border-base) solid var(--color-black);
 		color: var(--color-black);
@@ -32,7 +33,7 @@
 		}
 	}
 
-	&.primary {
+	&[data-primary='true'] {
 		background-color: var(--color-accent);
 		border: var(--border-base) solid var(--color-accent);
 		color: var(--color-white);
@@ -40,7 +41,7 @@
 		&:disabled,
 		&[aria-disabled='true'] { color: var(--color-black); }
 
-		&.outline {
+		&[data-outline='true'] {
 			background-color: transparent;
 			color: var(--color-accent);
 		}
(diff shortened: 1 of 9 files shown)


manvinderjit:
@madcampos  The following are the primary files to be reviewed

1. src/components/Button/Button.css
2. src/components/Button/Button.tsx

Remaining files are mostly dprint formatted changes or files showing sample Button usage.

lylaminju:
In the current implementation, I think `forwardRef` is not strictly necessary since:

- The sample usage in `sample-buttons.tsx` doesn't show any instances where the `ref` is being passed to the Button component.
- None of the examples are using `ref`s to access the underlying button element.

However, using `forwardRef` can follow React's best practices for reusable components.
@manvinderjit Could you elaborate why you added `forwardRef()` to the Button component?

manvinderjit:
> In the current implementation, I think `forwardRef` is not strictly necessary since:
> 
>     * The sample usage in `sample-buttons.tsx` doesn't show any instances where the `ref` is being passed to the Button component.
> 
>     * None of the examples are using `ref`s to access the underlying button element.
> 
> 
> However, using `forwardRef` can follow React's best practices for reusable components. @manvinderjit Could you elaborate why you added `forwardRef()` to the Button component?

@lylaminju I need it in the Complete Profile page. When a user clicks on the Remove Photo button, I need to refocus on the Photo Upload button for accessibility reasons.

Demo Steps:
1. Go to the following link
https://26-profile-page-css.volunteer-ekr.pages.dev/pages/complete-profile/
2. Focus on the _Upload Photo_ button using a keyboard and upload the photo
3. Press tab to shift to the _Remove Photo_ button and press enter to remove it.
4. Notice that the focus is not on the _Upload Photo_ button. To focus on it, we will have to press _Shift + Tab_.

Using a Ref can make this focus to shift to the _Upload Photo_ button when a photo is removed.

src/components/Button/Button.tsx
```diff
@@ -7,19 +7,20 @@ interface Props extends ButtonHTMLAttributes<HTMLButtonElement> {
...
 	hasOutline = false,
 	isPrimary = false,
 	isLarge = false,
 	children,
 	...rest
-}: Props) => (
+}, ref) => (
 	<button
+		ref={ref}
 		className={`button ${isPrimary ? 'primary' : ''} ${hasOutline ? 'outline' : ''} ${isLarge ? 'large' : 'small'}`}
```
madcampos:
```suggestion
		className='button'
		data-size={isLarge ? 'large' : 'small'}
		data-outline={hasOutline}
		data-primary={isPrimary}
```

For those kinds of states, it is nicer to use [`data-*` attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes) instead of shoving everything into `class`es.

src/components/Button/Button.css
```diff
@@ -1,32 +1,35 @@
...
-	color: var(var(--color-black));
+	color: var(--color-background);
 	cursor: pointer;
 	font-weight: var(--weight-bold);
-	height: 59px;
-	margin: 8px 0;
 	text-align: center;
 	width: 100%;
 
 	&.small {
```
madcampos:
```suggestion
	&[data-size="small"] {
```

src/components/Button/Button.css
```diff
@@ -1,32 +1,35 @@
...
 		font-size: var(--text-xs);
+		height: 2.625rem;
 		padding: var(--spacing-xs) var(--spacing-sm);
-		/* width: 19.375rem; */
 	}
 	&.large {
```
madcampos:
```suggestion
	&[data-size="lagre"] {
```

src/components/Button/Button.css
```diff
@@ -1,32 +1,35 @@
...
 		font-size: var(--text-lg);
 		height: 3.6875rem;
-		/* width: 23rem; */
-		padding: 12px 16px;
+		padding: 0.75rem 1rem;
 	}
 
 	&.outline {
```
madcampos:
```suggestion
	&[data-outline="true"] {
```

src/components/Button/Button.css
```diff
@@ -1,32 +1,35 @@
...
 		color: var(--color-black);
+
+		&:disabled,
+		&[aria-disabled='true'] {
+			background-color: transparent;
+			border: var(--border-base) solid var(--color-grey);
+		}
 	}
 
 	&.primary {
```
madcampos:
```suggestion
	&[data-primary="true"] {
```

src/css-layers/utilities.css
madcampos, line 1: **question:** Are we using any of those utility classes?‏

src/pages/button-usage/sample-buttons.tsx
madcampos, line 1: **question:** How about we add storybook to the project?‏

src/components/Button/Button.tsx
manvinderjit: @madcampos Should we make these changes after we complete the frontend integration as we will have to update buttons across all pages to align with these changes.

src/css-layers/utilities.css
manvinderjit: @madcampos The file is used elsewhere. It is being highlighted here because my linter changed single quotes to double quotes.

src/pages/button-usage/sample-buttons.tsx
manvinderjit: Evan was the one who presented this idea, I am not sure on this. However, I can research into this after we complete frontend integration.

src/css-layers/utilities.css
madcampos: I don't know, we can do a global search for the class names, if they are not used, or if they are easy to replace and fold into other code, then let's remove it.

manvinderjit: Okay, I will look into the same.

manvinderjit: @madcampos Made the changes and rebased on dev. It was easy, didn't have to change any components at all. Please review.

src/components/Button/Button.tsx
madcampos: No as it will not affact consumers of the component, only it's internals

madcampos (approved):

lylaminju (approved):

PR #79, opened 2025-05-21
78 sign up connection

### LINKED ISSUE #69: Sign up
Make API calls from the frontend page to the backend routes.

### LINKED ISSUE #78: Sign-up Connection Sub Issue Of Make API calls from the frontend

kbventures:
## Summary

Connect sign-up page to back end in development or production and redirect to "check your email" page. 

## Tests

Manual testing with front and back end running on 3000 and 4242 respectively.

## Issue

Closes #69

### DIFF
--- a/src/components/CheckYourEmail/CheckYourEmail.tsx
+++ b/src/components/CheckYourEmail/CheckYourEmail.tsx
@@ -4,11 +4,17 @@ import Header from '../../components/Header/Header';
 import ClockIcon from '../../components/Icons/ClockIcon';
 import './CheckYourEmail.css';
 
-const CheckYourEmail = ({ email }: { email: string }) => {
+const CheckYourEmail = () => {
+	const [email, setEmail] = useState('');
 	const RESEND_MINUTES = 10;
 	const MILLISECONDS_PER_MINUTE = 60 * 1000;
 	const [minutes, setMinutes] = useState(RESEND_MINUTES);
 	const [hasResent, setHasResent] = useState(false);
+
+	useEffect(() => {
+		const params = new URLSearchParams(window.location.search);
+		setEmail(params.get('email') || '');
+	}, []);
 
 	useEffect(() => {
 		const timer = setInterval(() => {
--- a/src/components/SignUp/SignUpForm.tsx
+++ b/src/components/SignUp/SignUpForm.tsx
@@ -38,7 +38,7 @@ const SignUpForm = (): React.JSX.Element => {
 				const errorData = await response.json();
 				console.log('Response not ok: ', errorData);
 			} else {
-				window.location.href = `${FE_URL}check-your-email/`;
+				window.location.href = `${FE_URL}check-your-email/?email=${encodeURIComponent(email)}`;
 			}
 		} catch (e) {
 			if (import.meta.env.MODE === 'development') {
--- a/src/components/SignUp/SignUpForm.tsx
+++ b/src/components/SignUp/SignUpForm.tsx
@@ -38,7 +38,7 @@ const SignUpForm = (): React.JSX.Element => {
 				const errorData = await response.json();
 				console.log('Response not ok: ', errorData);
 			} else {
-				window.location.href = `${FE_URL}}profile-completion`;
+				window.location.href = `${FE_URL}check-your-email/`;
 			}
 		} catch (e) {
 			if (import.meta.env.MODE === 'development') {
--- a/src/pages/check-your-email/check-your-email.tsx
+++ b/src/pages/check-your-email/check-your-email.tsx
@@ -10,7 +10,7 @@ const root = document.getElementById('root') as HTMLDivElement;
 createRoot(root).render(
 	(
 		<StrictMode>
-			<CheckYourEmail email='hadi.kamal@gmail.com' />
+			<CheckYourEmail />
 		</StrictMode>
 	)
 );


package.json
```diff
@@ -13,7 +13,8 @@
 		"lint:js": "eslint --fix",
 		"lint": "npm run typecheck && npm run lint:js",
 		"prepreview": "npm run build",
-		"preview": "vite preview"
+		"preview": "vite preview",
+		"kill:3000": "lsof -ti:3000 | xargs kill -9"
```
madcampos: **nitpick:** No.

This is very linux specific and also a very specific command in general.‏

You can quit the vite server by using the commands it provides (press `h + enter`) or by pressing `Ctrl+C` to send a `SIGTERM` to the process[^1].

[^1]: You can also use `Ctrl+D` to send a `SIGKILL` to the process but it may kill even your shell.

src/components/CheckYourEmail/CheckYourEmail.tsx
```diff
@@ -4,12 +4,18 @@ import Header from '../../components/Header/Header';
...
-const CheckYourEmail = ({ email }: { email: string }) => {
+const CheckYourEmail = () => {
+	const [email, setEmail] = useState('');
 	const RESEND_MINUTES = 10;
 	const MILLISECONDS_PER_MINUTE = 60 * 1000;
 	const [minutes, setMinutes] = useState(RESEND_MINUTES);
 	const [hasResent, setHasResent] = useState(false);
 
+	useEffect(() => {
+		const params = new URLSearchParams(window.location.search);
```
madcampos, line 15: **question:** Why are we adding the email to the url?‏

src/components/Home/Home.tsx
```diff
@@ -0,0 +1,72 @@
...
+						console.error(err.name);
+						console.error(err.cause);
+						console.error(err.message);
+						console.error(err.stack);
+					} else {
+						throw new Error('Error sign-up');
+					}
+				}
+			}
+		};
```
madcampos: **suggestion:** Please extract this function to a util function as it will be used throughout the codebase.‏

src/components/Home/Home.tsx
```diff
@@ -0,0 +1,72 @@
...
+					console.error(err.name);
+					console.error(err.cause);
+					console.error(err.message);
+					console.error(err.stack);
+				} else {
+					throw new Error('Error login out');
+				}
+			}
+		}
+	};
```
madcampos: Same thing here. This logic to handle redirections on the client should be in a helper function

src/utilities/getApiUrl.ts
```diff
@@ -1,3 +1,5 @@
 export function getApiUrl() {
-	return import.meta.env['APP_API_URL'] ?? 'https://community-hub.torontojs.com/';
+	const BE_URL = import.meta.env['APP_API_URL_DEV'] ?? 'https://community-hub.torontojs.com/';
+	const FE_URL = import.meta.env['APP_URL_DEV'] ?? '';
```
madcampos: **issue:** Those will both point to the same host, so we don't need two separate urls for them.‏

Also, the url can be derived from [`window.location`](https://developer.mozilla.org/en-US/docs/Web/API/Window/location), [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta), and [`new URL('/')`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)

src/utilities/getApiUrl.ts
```diff
@@ -1,3 +1,5 @@
 export function getApiUrl() {
```
madcampos: **suggestion:** ‏This exports only two constants that don't change, instead of exporting a function, directly export the constants.

src/components/SignUp/SignUpForm.tsx
```diff
@@ -38,7 +38,7 @@ const SignUpForm = (): React.JSX.Element => {
 				const errorData = await response.json();
 				console.log('Response not ok: ', errorData);
 			} else {
-				window.location.href = `${apiUrl}profile-completion`;
+				window.location.href = `${FE_URL}check-your-email/?email=${encodeURIComponent(email)}`;
```
madcampos, line 41:
```suggestion
				window.location.href = `/check-your-email/?email=${encodeURIComponent(email)}`;
```

src/components/SignUp/SignUpForm.tsx
```diff
@@ -26,7 +26,7 @@ const SignUpForm = (): React.JSX.Element => {
 
 	const signup = async (name: string, email: string, password: string) => {
 		try {
-			const response = await fetch(`${apiUrl}sign-up`, {
+			const response = await fetch(`${BE_URL}auth/sign-up`, {
```
madcampos:
```suggestion
			const response = await fetch(`/api/auth/sign-up`, {
```
Fetch resolves to the current host, so it will Just Work™

src/components/SignUp/SignUpForm.tsx
```diff
@@ -157,7 +157,7 @@ const SignUpForm = (): React.JSX.Element => {
 
 			<div className='have-account'>
 				<p className='not-member'>
-					If you already have an account, <a href={`${apiUrl}sign-in`} className='underline'>Click here to log-in</a>
+					If you already have an account, <a href={`${FE_URL}sign-in`} className='underline'>Click here to log-in</a>
```
madcampos:
```suggestion
					If you already have an account, <a href="/sign-in" className='underline'>Click here to log-in</a>
```

src/components/SignIn/SignInForm.tsx
```diff
@@ -13,19 +13,20 @@ const SignInForm = (): React.JSX.Element => {
...
+			const response = await fetch(`${BE_URL}auth/sign-in`, {
 				method: 'POST',
+				credentials: 'include',
 				headers: {
 					'Content-Type': 'application/json'
 				},
 				body: JSON.stringify({ email, password })
 			});
-			if (!response.ok) {
+			if (response.status !== 201) {
```
madcampos: **question:** Isn't the server returning a 200 response?‏

src/components/SignIn/SignInForm.tsx
```diff
@@ -13,19 +13,20 @@ const SignInForm = (): React.JSX.Element => {
...
 				setIsLoading(false);
 				const errorData = await response.json();
 				console.log('Response not ok: ', errorData);
 			} else {
-				window.location.href = `${apiUrl}home`;
+				window.location.href = `${FE_URL}home/`;
```
madcampos:
```suggestion
				window.location.href = '/home/';
```

src/components/SignIn/SignInForm.tsx
```diff
@@ -109,7 +110,7 @@ const SignInForm = (): React.JSX.Element => {
 			</Button>
 
 			<div>
-				<a href={`${apiUrl}forgot-passsword`} className='do-not-remember-password'>I don't remember my password</a>
+				<a href={`${FE_URL}forgot-passsword`} className='do-not-remember-password'>I don't remember my password</a>
```
madcampos:
```suggestion
				<a href="/forgot-passsword" className='do-not-remember-password'>I don't remember my password</a>
```

src/components/SignIn/SignInForm.tsx
```diff
@@ -118,7 +119,7 @@ const SignInForm = (): React.JSX.Element => {
 
 			<div className='have-account'>
 				<p className='not-member'>
-					If you are a member of ToronoJS and don't have your account, <a href={`${apiUrl}sign-up`} className='underline'>click here to sign-up</a>
+					If you are a member of ToronoJS and don't have your account, <a href={`${FE_URL}sign-up`} className='underline'>click here to sign-up</a>
```
madcampos:
```suggestion
					If you are a member of ToronoJS and don't have your account, <a href="/sign-up" className='underline'>click here to sign-up</a>
```

src/components/SignIn/SignInForm.tsx
```diff
@@ -13,19 +13,20 @@ const SignInForm = (): React.JSX.Element => {
 
 	const signin = async (email: string, password: string) => {
 		try {
-			const response = await fetch(`${apiUrl}sign-in`, {
+			const response = await fetch(`${BE_URL}auth/sign-in`, {
```
madcampos:
```suggestion
			const response = await fetch('/api/auth/sign-in', {
```

madcampos:
I've changed the base branch for this PR, when creating chained branches make sure that your PRs are pointing to the right branch.

E.g.:
> You create this sequence: `dev` -> `branch-a` -> `branch-b`.
> `branch-a` is based on `dev`, and `branch-b` is based on `branch-a`.
> Your PRs should then be:
> 1. `dev`..`branch-a`
> 2. `branch-a`..`branch-b`

When creating a pull request github will show you the UI to select both branches:

Screenshot of github's UI showing the branch selection when comparing two branches.

src/components/CheckYourEmail/CheckYourEmail.tsx
manvinderjit: I think it is used to show the email in the activation message. 

image

@madcampos @kbventures Would it be better to use local storage instead of the url parameter to obscure the email, although we will need to remove it from storage. Or we could skip showing the email altogether.

madcampos: I'm thinking here how to solve that... The best option would be to embed the email on the page, that is close to an SSR approach and works extremely well.

Problem is: we are not doing any SSR here (yet?), so we need to convey this information somehow.

I think the two options we have are either how it is done with the email in the url or  with a request to the server to retrieve the email.

My concern with keeping it in the URL is that it may lead to abuse and create a vector for attacks. Imagine the scenario:

> A malicious actor gets my email address, which is not hard as it is available on my website, resume, slack, linkedin, etc.
>
> They then send a request to the confirm email page with `?email=me@madcampos.dev`.

How would the application work in this case? Would it allow the malicious actor to send requests using my email? Would it redirect to the signin page with an error, thus providing a way to probe if the email is registered? Would it error out with some other information, thus providing another way to probe for registered emails?

package.json
kbventures: CTRL-C doesn't always kill the port but I agree that linux specific commands don't belong here.

src/components/Home/Home.tsx
manvinderjit: Should we use react hooks?

src/components/CheckYourEmail/CheckYourEmail.tsx
manvinderjit: 1. How about we create an httpOnly cookie on `Sign Up` that has a `uuid`, could be account's id. 
2. We create a backend route that checks if there is an account awaiting activation based on the `uuid`.
3. On sign up, the user is directed to the `Check Your Email` page. This page then has a useEffect which hits the backend route to fetch the email.

On the downside, we may need a new route, possibly a new table for this.

src/components/Home/Home.tsx
kbventures: Moved both the heartbeat and handle log out to utility folder as individual reusable functions.

src/components/CheckYourEmail/CheckYourEmail.tsx
kbventures: @madcampos @manvinderjit should I create an issue for this and we can discuss it further but leave it as is for now?

src/components/SignIn/SignInForm.tsx
kbventures: That will not work.

src/components/SignIn/SignInForm.tsx
kbventures: I tried implenting this on sign-in and sign-up and it doesn't work.

src/components/CheckYourEmail/CheckYourEmail.tsx
madcampos: @kbventures yes please, create an issue and link to this discussion.

src/components/Home/Home.tsx
madcampos: @manvinderjit you mean like creating custom hooks? Yeah, that is a good idea.

src/components/SignIn/SignInForm.tsx
madcampos: @kbventures when you hover over the link on the browser, what url does it show?

It should work, the problem I think is the url we are linking to though

package.json
madcampos: You shouldn't need to kill the process holding the port that often... That seems like an edge case to me. 🤔

What is happening that made you add this command?

kbventures: @madcampos in linux often times when even after you CRTL + C the post remain s in use.  It happens a lot.

src/components/CheckYourEmail/CheckYourEmail.tsx
kbventures: https://github.com/torontojs/communityhub-fe/issues/88

src/components/SignIn/SignInForm.tsx
kbventures: Created is 201

image

Go ahead an change created 200 on the back end?

madcampos: No, keep it as 201, it makes more sense

package.json
madcampos: Oh I get it. Try quitting the server by pressing it's quit command: `q + Enter`.

Or else it may not receive as proper `SIGTERM` and node has the worst and most annoying handling of process termination in existence (it beats Go's date bs in how bad it is).

kbventures: Will give this a try.

src/components/SignUp/SignUpForm.tsx
```diff
@@ -38,7 +38,7 @@ const SignUpForm = (): React.JSX.Element => {
 				const errorData = await response.json();
 				console.log('Response not ok: ', errorData);
 			} else {
-				window.location.href = `${FE_URL}}profile-completion`;
+				window.location.href = `${FE_URL}check-your-email/?email=${encodeURIComponent(email)}`;
```
madcampos, line 41:
```suggestion
				window.location.href = `/check-your-email/?email=${encodeURIComponent(email)}`;
```

PR #117, opened 2025-06-04
89 intergration testing back end

### LINKED ISSUE #89: Intergration Testing Back End
- [x] Connect Resend to local development
- [x] Setup Intergration tests utilizing vitests and wrangler local development infra
- [ ] Intergration tests for authentication route
- [ ] Intergration tests for middleware
- [ ] Intergration tests for profile route
- [ ] Intergration tests for team route
- [ ] Intergration tests for role route
- [ ] Intergration tests for event route

kbventures:
## Summary

Inergration test for the auth route:

- Sign-up
- Sign-in
- Sign-out
- Heartbeat
- Profile
- Team
- Role


api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
+# To run tests:
+# hurl --variable host=http://localhost:3000 sign-up.hurl --verbose --test
```
madcampos: **suggestion:** Let's add a command to `package.json`‏'s `script` session called `test:api` or something and call if like:
```shell
npm run test:api
```

Also, I think those options would be helpful to use:
- [`--glob`](https://hurl.dev/docs/manual.html#glob) to make life easier.
- [`--continue-on-error`](https://hurl.dev/docs/manual.html#continue-on-error) so it runs the entire test suite.
- [`--test`](https://hurl.dev/docs/manual.html#test) to run like a test runner.
- [`--jobs`](https://hurl.dev/docs/manual.html#jobs) to run things in parallel and controll how many things are ran.
- [`--max-time`](https://hurl.dev/docs/manual.html#max-time) for limiting how long the test suite should run. It is good for CI.
- [`--no-color`](https://hurl.dev/docs/manual.html#no-color) for CI.
- [`--output-tap`](https://hurl.dev/docs/manual.html#report-tap) this is a semi standard format for test runners.
- [`--variable-file`](https://hurl.dev/docs/manual.html#variables-file) I think we can use the `.vars` file here or something... Or pass variables individually, but that can get tedious.

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+
+# Sign-up success OKAY
+
+POST {{host}}/api/auth/sign-up
+
+Content-Type: application/json
+
+{
+	"name":"Mr Robot",
+	"email":"sign-up@example.com",
```
madcampos: **suggestion:**‏
```suggestion
	"email":"rmalek@mr-robot.com",
```

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+# Sign-up weak password check
+
+# POST http://localhost:3000/api/auth/sign-up
+# https://github.com/torontojs/communityhub/issues/96
+#Content-Type: application/json
+
+#{
+#	"name":"Mr Weak Password",
+#	"email":"weak@weakpassword.com",
+#	"password":"quoison"
```
madcampos: **issue:** Please include the link to the meme as a comment 🙃‏

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+
+Content-Type: application/json
+
+{
+	"name":" Mr Edgar  Extra Spaces ",
+	"email":"name@extra-spaces.com",
+	"password":"ExtraSpaces911!@#$"
+}
+
+HTTP 200
```
madcampos: **thought:** ‏These tests should follow a subsequent request to check if the spaces are trimmed.‏

The problem is here the user id will be random so checking for the new user would require them to be activated, signed in, or something else... I think here sign in then heartbeat would be the easiest way to get this information.

Side note: we should have a similar test for editing a profile.

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+
+Content-Type: application/json
+
+{
+	"name":"Invalid Eamil Format",
+	"email":"invalid-format.com",
+	"password":"invalidEmailF911!@#$"
+}
+
+HTTP 422
```
madcampos: **praise:** nice test 10/10‏

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+
+POST {{host}}/api/auth/sign-up
+
+Content-Type: application/json
+
+{
+	"name":"x",
+	"email":"single-digit-name@example.com",
+	"password":"SingleDigitName84!#$"
+}
```
madcampos: **issue:** This should be allowed, no problem here.‏

The only problems with names are the field being empty, aside from that it is however the person would like to be called. 

If their name is "\u0000" ("null control code charcacter") that is fine. As long as we treat the strings properly [it _technically_ doesn't break the database](https://sqlite.org/nulinstr.html), and js can handle it just fine.

To make our lives easier we should filter and block [control characters](https://en.wikipedia.org/wiki/ASCII#Control_code_table) though, but everything else is game.

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+
+HTTP 200
+
+
+# Sign-up empty body
+
+POST {{host}}/api/auth/sign-up
+
+HTTP 422
```
madcampos: **suggestion:** Add a couple of tests here:
- Add the content type, but malformed/invalid json (`invalid json`)
- Add content type but empty string as body (`""`)
- Add content type but empty json (`{}`)

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+
+Content-Type: application/image
+
+{
+	"name":"Missinger Header",
+	"email":"missing@header.com",
+	"password":"ThisMissingHeader911!@#$"
+}
+
+HTTP 422
```
madcampos: **issue:** This is the _wrong_ content type, the previous one have a _missing_ content type.‏

api/src/routes/auth/sign-up.hurl
```diff
@@ -0,0 +1,220 @@
...
+
+# Content-Type: application/json
+
+# {
+#	"name":"BortMaximiliano-Alexanderthegreat VonSchwarzensteinberg-MacDonaldson O'Flannagan-Smith-Johansson-Ricci",
+#	"email":"longname@header.com",
+#	"password":"ShouldNotBeLongerThan100124!@#$"
+#}
+
+#HTTP 413
```
madcampos: **suggestion:** We can put a base64 encoded image as the content, that will easily blow up the size limits 🙃‏

api/src/routes/auth/signin-signout-heartbeat.hurl
```diff
@@ -0,0 +1,142 @@
...
+# Heartbeat invalid token.
+
+GET {{host}}/api/auth/heartbeat
+
+Cookie: auth_token=00000000-0000-0000-0000-000000000000
+
+HTTP 401
+
+
+# TODO: Heartbeat session expired(needs seed data)
```
madcampos: **issue:** This cannot be easily seeded, as sessions are kept in KV and not D3 🥲‏

api/src/routes/auth/signin-signout-heartbeat.hurl
```diff
@@ -0,0 +1,142 @@
...
+# https://github.com/torontojs/communityhub/issues/96
+
+POST {{host}}/api/auth/sign-in
+
+Content-Type: application/json
+
+{
+  "email": "profile1@example.com",
+  "password": "abcd"
+}
```
madcampos: **issue:** Password strength is irrelevant for sign-in,‏ it should be checked when creating the password. During sign-in we only validate if the string matches the hashed password, the strength needs to be checked when we create the hash.

api/src/routes/auth/signin-signout-heartbeat.hurl
```diff
@@ -0,0 +1,142 @@
...
+
+HTTP 400
```
madcampos: **question:** I think here we just revalidate the existing token, no? There was a discussion about that but I don't remember what we decided.‏

api/src/routes/auth/signin-signout-heartbeat.hurl
```diff
@@ -0,0 +1,142 @@
...
+Cookie: auth_token=00000000-0000-0000-0000-000000000000
+
+HTTP 404
+
+
+# Sign out missing token BAD_REQUEST
+
+POST {{host}}/api/auth/sign-out
+
+HTTP 401
```
madcampos: **suggestion:** ‏we should add a test for signing out again with the same token.

api/src/routes/auth/signin-signout-heartbeat.hurl
```diff
@@ -0,0 +1,142 @@
...
+
+
+# TODO: Heartbeat accessing page requiring organizer or admin access FORBIDDEN
```
madcampos: **issue:** ‏heartbeat is not tied to any other page/api endpoint. This test does not make sense.

We do however need to test trying to do things that are only allowed to admins with other accounts.

api/src/routes/profile/index.ts
```diff
@@ -1,4 +1,4 @@
-// FIXME: readd checks for existing ids, based on the doesTeamExist function
+// FIXME: read checks for existing ids, based on the doesTeamExist function
```
madcampos:
```suggestion
// FIXME: re-add checks for existing ids, based on the doesTeamExist function
```

api/src/routes/profile/index.ts
```diff
@@ -88,7 +88,7 @@ profileRoutes.openapi(
 	async (context) => {
 		const sessionData = getSession(context);
 		const profile = await getProfileById(context.env.Database, sessionData.id);
-
+		console.log('profile', profile);
```
madcampos: **nitpick:** ‏please remove console statements

api/src/routes/profile/profile.hurl
```diff
@@ -0,0 +1,156 @@
...
+jsonpath "$.data[1].email" == "profile2@example.com"
+jsonpath "$.data[1].name" == "Jane Smith"
+jsonpath "$.data[1].description" == "Another sample profile description"
+jsonpath "$.data[1].isBasedOnGTA" == true
+jsonpath "$.data[1].canJoinLocalEvents" == true
+jsonpath "$.data[1].happenedAt" == "2025-01-20T11:00:00Z"
+jsonpath "$.data[1].insertedAt" == "2025-01-20T11:00:00Z"
+jsonpath "$.data[1].activatedAt" == "2025-01-20T11:10:00Z"
+jsonpath "$.data[1].links" count == 0
+jsonpath "$.data[1].skills" count == 0
```
madcampos: **issue:** ‏This test is brittle and beyond the point. We should not be testing if every field exists as it makes the test tightly coupled to the code that generates/formats the output. Here we should check for basic things like the overall shape of the API, like if pagination is correct, or if one of the profile ids exist.

A test for all fields is better done as a unit test, as it is already tightly coupled with the code and tests if the code outputs the expected format.

api/src/routes/profile/profile.hurl
```diff
@@ -0,0 +1,156 @@
...
+
+# TODO test pagination when feature is constructed: public profiles GET
+
+# Fix me(should be 401) Get self profile UNAUTHORIZED
+GET {{host}}/api/auth/profiles/self
+
+HTTP 404
```
madcampos: **thought:** I think this endpoint is not implemented yet...‏

api/src/routes/profile/profile.hurl
```diff
@@ -0,0 +1,156 @@
...
+jsonpath "$.data.email" == "profile1@example.com"
+jsonpath "$.data.name" == "John Doe"
+jsonpath "$.data.description" == "A sample profile description"
+jsonpath "$.data.isBasedOnGTA" == true
+jsonpath "$.data.canJoinLocalEvents" == true
+jsonpath "$.data.happenedAt" == "2025-01-20T10:00:00Z"
+jsonpath "$.data.insertedAt" == "2025-01-20T10:00:00Z"
+jsonpath "$.data.activatedAt" == "2025-01-20T10:05:00Z"
+jsonpath "$.data.links" count == 0
+jsonpath "$.data.skills" count == 0
```
madcampos: Same thing here: the test should verify the id match the known id for this profile you are testing, that is sufficient here.

api/src/routes/profile/profile.hurl
```diff
@@ -0,0 +1,156 @@
...
```
madcampos: Same thing here: just checking ids is sufficient.

api/src/routes/profile/profile.hurl
```diff
@@ -0,0 +1,156 @@
...
+HTTP 404
+
+
+# Fix me(Should be 404): Patch profile by id not authorized UNAUTHORIZED
+PATCH {{host}}/api/profiles/3227114d-43c4-42ed-8aea-f3860fe42222
+
+HTTP 422
+
+
+# Patch profile by id (volunteer patch self) OKEY
```
madcampos:
```suggestion
# Patch profile by id (volunteer patch self) OKAY
```

kbventures:
@madcampos this issue is still a draft pr.  I'l apply what I can but It is not ready for review.

api/src/routes/auth/sign-up.hurl
kbventures: @lylaminju

api/src/routes/profile/profile.hurl
kbventures: Thanks this will save  me a LOT of time. :)

api/src/routes/auth/01_sign-up.hurl
```diff
@@ -32,7 +27,8 @@ HTTP 200
 
 
 # Sign-up duplicate email with different casing OKAY
-
+# FIX ME:
+# https://github.com/torontojs/communityhub/issues/97
```
madcampos:
```suggestion
# FIXME: https://github.com/torontojs/communityhub/issues/97
```

api/src/routes/auth/01_sign-up.hurl
```diff
@@ -46,6 +42,42 @@ Content-Type: application/json
...
+HTTP 200
+
+
+POST http://localhost:3000/api/auth/sign-up
+Content-Type: application/json
+
+{
+  "name": "Mr 😎🔥",
+  "email": "emoji@unicode.com",
+  "password": "Emoji911!@#"
```
madcampos:
```suggestion
  "password": "Emoji911🙃♻️!@#"
```

api/src/routes/auth/01_sign-up.hurl
```diff
@@ -46,6 +42,42 @@ Content-Type: application/json
...
+
+HTTP 200
+
+
+POST http://localhost:3000/api/auth/sign-up
+Content-Type: application/json
+
+{
+  "name": "Mr 😎🔥",
+  "email": "emoji@unicode.com",
```
madcampos: **thought**: Emails [also support Unicode](https://en.wikipedia.org/wiki/Email_address#Internationalization), but it is less common.

api/src/routes/auth/02_signin-signout-heartbeat.hurl
```diff
@@ -58,85 +54,137 @@ HTTP 422
...
-
+# Sign in intergration test CREATED
 POST {{host}}/api/auth/sign-in
 
 Content-Type: application/json
 
 {
-  "email": "profile1@example.com",
-  "password": "abcd"
+  "email": "sir.robin@cowardly.co",
```
madcampos: https://www.youtube.com/watch?v=BZwuTo7zKM8

madcampos:
#protip: when merging from a branch that has at some point been merged into the main branch. Merge it back into yours to avoid re-review of code that is already reviewed and merged.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
+# Public profiles GET (Testing the first tree basic data points) OKAY
+GET {{host}}/api/profiles
+
+HTTP 200
+[Asserts]
+jsonpath "$.data[1].id" == "3c5123c0-8548-4a02-a83c-32e9ce67eae8"
+jsonpath "$.data[2].id" == "bf950534-e8f9-4f5a-a246-610aa63faf21"
+jsonpath "$.data[3].id" == "37a4882c-3b7d-46c0-b019-c37e2445dae8"
+jsonpath "$.data[4].id" == "890cae10-7bbd-4f0c-961a-df495ec7527b"
```
madcampos: **suggestion:** Can we test if those ids exist but not their order? I think order can be a little bit brittle as we can't guarantee the seed file order will ever be kept the same.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+
+
+# TODO test pagination when feature is constructed
+
+
+# Fix me: (should be 401) Get self profile UNAUTHORIZED
```
madcampos:
```suggestion
# FIXME: (should be 401) Get self profile UNAUTHORIZED
```
A `FIXME` comment starts with the word `FIXME`, all caps, followed by a `:`.

api/src/routes/auth/02_signin-signout-heartbeat.hurl
madcampos, line 1: Can we reword the comments, the "integration" part in most of them makes the comments more confusing, are they integrating with what exactly?

Please be descriptive on what is being tested so someone can jump to an arbitrary test and can still understand it by the comment/context.

api/src/routes/auth/02_signin-signout-heartbeat.hurl
```diff
@@ -58,85 +54,137 @@ HTTP 422
...
+GET {{host}}/api/auth/heartbeat
+Cookie: auth_token={{ auth-token-cookie-var }}
+
+HTTP 200
+
+
+POST {{host}}/api/auth/sign-out
+Cookie: auth_token={{ auth-token-cookie-var }}
+
+HTTP 204
```
madcampos: Those tests, the way they are right now are redundant as any user that logs in will have the heartbeat return something to them 🤷‍♂️.

It would be better here to also check for the access role for that user to confirm they are indeed the role you are testing for.

api/src/routes/auth/01_sign-up.hurl
madcampos, line 1: **nitpick**: please remove the number prefixes from the file names.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+jsonpath "$.data.id" == "37a4882c-3b7d-46c0-b019-c37e2445dae8"
+
+
+POST {{host}}/api/auth/sign-out
+Cookie: auth_token={{ auth-token-cookie-var }}
+
+HTTP 204
+
+
+# GET profile using id OKEY
```
madcampos:
```suggestion
# GET profile using id OKAY
```

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+GET {{host}}/api/profiles/self
+Cookie: auth_auth={{ auth-token-cookie-var }}
+
+HTTP 200
+[Asserts]
+jsonpath "$.data.id" == "37a4882c-3b7d-46c0-b019-c37e2445dae8"
```
madcampos: **suggestion**: This ID should be saved in a variable. I think hurl has a way to define variables in the file itself, so we don't bloat a `.env` file with random things.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+# GET profile using id OKEY
+GET {{host}}/api/profiles/37a4882c-3b7d-46c0-b019-c37e2445dae8
+HTTP 200
+
+[Asserts]
+jsonpath "$.data.id" == "37a4882c-3b7d-46c0-b019-c37e2445dae8"
+
+
+# GET profile using non existant id NOT_FOUND
+GET {{host}}/api/profiles/3227114d-43c4-42ed-8aea-f3860fe42122
```
madcampos:
```suggestion
GET {{host}}/api/profiles/00000000-0000-0000-0000-000000000000
```

This is a special UUID that is never used but is still valid.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+
+HTTP 404
+
+
+# Get profile using malformed uuid
+GET {{host}}/api/profiles/00000000-0000-0000-0000-000000000000
```
madcampos:
```suggestion
GET {{host}}/api/profiles/00000000-ZZZZ-0000-0000-000000000000
```

This is a malformed UUID

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+
+# Fix me(Should be 404): Patch profile by id not authorized UNAUTHORIZED
```
madcampos:
```suggestion
# FIXME: (Should be 404) - Patch profile by id not authorized UNAUTHORIZED
```

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+PATCH {{host}}/api/profiles/3227114d-43c4-42ed-8aea-f3860fe42222
+
+HTTP 401
```
madcampos: Please add more context to the comment on why this should be unauthorized.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+# Patch profile by id UNAUTHORIZED
+PATCH {{host}}/api/profiles/37a4882c-3b7d-46c0-b019-c37e2445dae8
+
+Content-Type: application/json
+
+{
+  "name":"new name"
+}
+
+HTTP 401
```
madcampos: Sending a request to any profile without being authenticated first will be unauthorized. Please add more context on what this test is testing.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+  "password": "RunAway!1234"
+}
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+PATCH {{host}}/api/profiles/37a4882c-3b7d-46c0-b019-c37e2445dae8
+Cookie: auth_auth={{ auth-token-cookie-var }}
```
madcampos: **suggestion**: If you are using the same user for most tests, do the signing and save the auth token in the top of the file and then reuse the same auth token.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+  "email": "sir.robin@cowardly.co",
+  "password": "RunAway!1234"
+}
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+PATCH {{host}}/api/profiles/37a4882c-3b7d-46c0-b019-c37e2445dae8
```
madcampos: **thought:** Should we add a special `/self` path for editing your own profile?

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+Cookie: auth_auth={{ auth-token-cookie-var }}
+
+Content-Type: application/json
+
+{
+  "name":"Sir Robin"
+}
+
+HTTP 200
```
madcampos: No need to "undo" the changes to an edit. The idea of a test is that it should be reproductible by throwing away everything and then restarting from an empty state and applying the same series of steps so it generates the same output.

api/src/routes/profile/03_profile.hurl
```diff
@@ -0,0 +1,284 @@
...
+  "email": "root@torontojs.com",
+  "password": "correct horse battery staple"
+}
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+DELETE {{host}}/api/profiles/7a4882c-3b7d-46c0-b019-c37e2445dae
```
madcampos: **issue:** What is wrong with this ID? Is it just non existant or is it malformed or something else?

package.json
```diff
@@ -27,8 +27,16 @@
 		"lint:js": "eslint",
 		"lint": "npm run typecheck:api && npm run typecheck:client && npm run lint:js",
 
-		"test": "vitest",
-		"coverage": "vitest run --coverage"
+		"test:unit": "vitest",
+		"coverage": "vitest run --coverage",
+
+		"test:api:locally": "hurl --variables-file .dev.vars api/src/routes/**/*.hurl --verbose --test",
+		"test:api:ci": "hurl --glob 'tests/**/*.hurl' --continue-on-error --test --jobs 4 --max-time 300 --no-color --output-tap results.tap --variable-file .env.vars",
```
madcampos: **question**: Why do we have two tests, one for local and other for CI? Can we always run the same?

package.json
```diff
@@ -27,8 +27,16 @@
...
+		"test:api:auth1": "hurl --variables-file .dev.vars api/src/routes/auth/01_sign-up.hurl --test --verbose",
+		"test:api:auth2": "hurl --variables-file .dev.vars api/src/routes/auth/02_signin-signout-heartbeat.hurl --test --verbose",
+		"test:api:profiles": "hurl --variables-file .dev.vars api/src/routes/profile/03_profile.hurl --test --verbose",
+		"test:api:teams": "hurl --variables-file .dev.vars api/src/routes/team/04_teams.hurl --test --verbose",
+		"test:api:team-members": "hurl --variables-file .dev.vars api/src/routes/team-members/05_team-members.hurl --test --verbose"
```
madcampos: Try this:
```suggestion
		"test:api:file": "hurl --variables-file .dev.vars --test --verbose"
```

You can pass the other argument to npm like common unix tools:
```sh
npm run test:api:file -- path/to/file
```

api/src/routes/team/data.ts
```diff
@@ -42,11 +42,13 @@ export async function insertTeam(database: D1Database, profileId: string, { name
 }
 
 export async function updateTeamById(database: D1Database, id: string, data: UpdateTeamData) {
+	const keys = Object.keys(data);
+	const fields = keys.map((keyId) => `${keyId} = ?`).join(', ');
```
madcampos: Back to my previous comment: please make those fields explicit

api/src/routes/team/04_teams.hurl
```diff
@@ -0,0 +1,341 @@
+# Get all teams OKAY
+GET {{host}}/api/teams
+
+HTTP 200
+[Asserts]
+jsonpath "$.data[0].id" == "b3410598-ecbc-41be-9f68-925da74bc613"
+jsonpath "$.data[1].id" == "72bdd945-1479-4f87-8ce7-ef0a6c5101c8"
+jsonpath "$.data[2].id" == "70ca1ff2-6ab6-4533-ae79-203f5b08742c"
+jsonpath "$.data[3].id" == "82df0f4a-2a32-4c14-86d9-4b17fc7566e4"
```
madcampos: Same as with profiles: can we check if the list includes those IDs?

api/src/routes/team/04_teams.hurl
madcampos, line 1: **suggestion**: How about we split the test files into smaller chunks and add them to a folder?

E.g.:
- `tests/create-teams.hurl`
- `tests/update-teams.hurl`
- `tests/delete-teams.hurl`

Or even:

- `test/public-teams.hurl`
- `test/volunteer-teams.hurl`
- `test/organizer-teams.hurl`
- `test/admin-teams.hurl`

api/src/routes/team-members/05_team-members.hurl
madcampos, line 1: Same thing about the comments here: let's expand them a little bit to add more context on what the tests do.

api/src/routes/team-members/05_team-members.hurl
```diff
@@ -0,0 +1,587 @@
...
+
+[
+	{
+		"name":"My Role is Quoison",
+		"profileId":"bf950534-e8f9-4f5a-a246-610aa63faf21",
+		"description":"Marco's assistant"
+	},
+	{
+		"name":"Unauthorized quoison role",
+		"profileId":"22fa5d4f-1e53-42f3-a00d-7010e7973b12"
```
madcampos: **question**: Why would a user be unauthorized here?

madcampos (commented):
Looks great, my main comments are around organization and ergonomics of the tests.

api/src/routes/team/04_teams.hurl
kbventures: Should the test folders be found within their respective routes or a main tests folder found at the root of the project, or /api/tests

kbventures: I centralized our api test into a single test directory in lexicographic ordering.

api/src/routes/auth/sign-up.hurl
kbventures: Can't be actioned until issue linke attached is actioned.

.husky/pre-commit
madcampos, line 1: **issue:** Keep the space or else it may not be interpreted correctly by the shell

.husky/post-checkout
madcampos, line 1: **issue:** Keep the space or else it may not be interpreted correctly by the shell

.husky/pre-push
```diff
@@ -1,3 +1,6 @@
-#! /bin/bash
+#!/bin/bash
```
madcampos, line 1: **issue:** Keep the space or else it may not be interpreted correctly by the shell

package.json
```diff
@@ -8,28 +8,26 @@
...
 		"lint:js": "eslint",
 		"lint": "npm run typecheck:api && npm run typecheck:client && npm run lint:js",
-
-		"test": "vitest",
-		"coverage": "vitest run --coverage"
+		"test:unit": "vitest",
+		"coverage": "vitest run --coverage",
+		"test:api": "hurl --variables-file .dev.vars api/api_tests/*.hurl --verbose --test",
+		"test:api:ci": "hurl --glob 'tests/**/*.hurl' --continue-on-error --test --jobs 4 --max-time 300 --no-color --output-tap results.tap --variable-file .env.vars",
+		"reset:db": "rm -rf .wrangler  &&  yes | npm run db:setup && npm run db:seed"
```
madcampos: **issue**: this won't work on Windows and will break the pre-push hook

api/src/routes/team/data.ts
```diff
@@ -11,20 +11,19 @@ export async function doesTeamExist(database: D1Database, id: string) {
...
 	const { id, schemaVersion, happenedAt, insertedAt } = generateBaseDBfields();
 
 	const results = await database.batch([
 		database.prepare(`
 			INSERT INTO ${DBTables.TEAM} (
 				id, schemaVersion, happenedAt, insertedAt,
-				name
-				${description ? ', description' : ''}
+				name,
+				description
```
madcampos: **nitpick**:
```suggestion
				name, description
```

.github/ci.yml
madcampos, line 1: We need to update the rest command here to actually run the tests.

A nitpick: I think we should rename the files to something like `ci-tests` to be more explict.

package.json
```diff
@@ -8,28 +8,26 @@
...
 		"typecheck:client": "tsc -p ./src --noEmit",
 		"lint:js": "eslint",
 		"lint": "npm run typecheck:api && npm run typecheck:client && npm run lint:js",
-
-		"test": "vitest",
-		"coverage": "vitest run --coverage"
+		"test:unit": "vitest",
+		"coverage": "vitest run --coverage",
+		"test:api": "hurl --variables-file .dev.vars api/api_tests/*.hurl --verbose --test",
+		"test:api:ci": "hurl --glob 'tests/**/*.hurl' --continue-on-error --test --jobs 4 --max-time 300 --no-color --output-tap results.tap --variable-file .env.vars",
```
madcampos:
```suggestion
		"test:api": "hurl --glob \"./api/**/*.hurl\" --test --continue-on-error --jobs 4 --max-time 300 --output-tap \"./api-test-results.tap\"  --variables-file \"./.dev.vars\" --verbose",
		"test:api:ci": "hurl --glob \"./api/**/*.hurl\" --test --continue-on-error --jobs 4 --max-time 300 --output-tap \"./api-test-results.tap\" --no-color",
```

Let's keep the commands as similar as possible. Also, on CI, there is no `.dev.vars` file, it has to be provided from somewhere else.

madcampos (changes requested):

package.json
kbventures: Question: Is there a convention why it would be named result.tap?

api/src/routes/profile/03_profile.hurl
madcampos: @kbventures I'm being very picky here because that is the most common pattern, it is "`FIXME:`", one word, all upper case, followed by a colon. This is the same as "`TODO:`".

madcampos (dismissed):

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+
+Content-Type: application/json
+
+{
+	"name":"Mr Duplicate Email",
+	"email":"sign-up@example.com",
+	"password":"FruityThisIs911!@#"
+}
+
+HTTP 200
```
manvinderjit, line 26: question: Do we want to return status code as 200 for duplicate email instead of 4xx?

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+
+Content-Type: application/json
+
+{
+	"name":"Mr Duplicate Email",
+	"email":"SIGN-up@example.com",
+	"password":"FruityThisIs911!@#"
+}
+
+HTTP 200
```
manvinderjit, line 41: question: Same here. Do we want to return status code as 200 for duplicate email instead of 4xx?

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+	"name":"Missing Passowrd",
+	"email":"email@missingpassword.com",
+	"password":""
+}
+
+HTTP 422
+
+
+# Sign-up name extra spaces
+# https://github.com/torontojs/communityhub/issues/118
```
manvinderjit: nitpick: This is issue is marked as complete. Do we want to remove it from here?

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+	"name":" Mr Edgar  Extra Spaces ",
+	"email":"name@extra-spaces.com",
+	"password":"ExtraSpaces911!@#$"
+}
+
+HTTP 200
+
+
+# Sign-up email extra spaces
+# https://github.com/torontojs/communityhub/issues/118
```
manvinderjit: nitpick: Same here.

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+	"name":"Mr Edgar Extra Spaces",
+	"email":" email@extra-spaces.com ",
+	"password":"ExtraSpaces911!@#$"
+}
+
+HTTP 422
+
+
+# Sign-up password extra spaces
+# https://github.com/torontojs/communityhub/issues/118
```
manvinderjit: nitpick: Same here.

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+# Missing content-type
+POST {{host}}/api/auth/sign-up
+
+{
+  "name": "No Header",
+  "email": "noheader@example.com",
+  "password": "NoHeader911!@#"
+}
+
+HTTP 200
```
manvinderjit, line 68: Are we returning 200 and not 4xx for missing content type?

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+
+Content-Type: application/json
+
+{
+	"name":"Mr Edgar Extra Spaces",
+	"email":"email@extra-spaces.com",
+	"password":" ExtraS paces911!@#$ "
+}
+
+HTTP 200
```
manvinderjit: nitpick: I am assuming we are allowing spaces in password? Recommend adding comments for clarity.

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+
+Content-Type: application/json
+
+{
+	"name":"x",
+	"email":"single-digit-name@example.com",
+	"password":"SingleDigitName84!#$"
+}
+
+HTTP 200
```
manvinderjit, line 209: Are we allowing single digit names?

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -0,0 +1,246 @@
...
+	"email":"missing@header.com",
+	"password":"ThisMissingHeader911!@#$"
+}
+
+HTTP 422
+
+# Maximum name email password 512 characters
+# For create profile field 10240(10kb) characters(text area, profile description, longer text areas and )
+# Sign-up name should not be longer than 100 characters
+# https://github.com/torontojs/communityhub/issues/122
```
manvinderjit, line 233: praise: love the comments here for clarity. we should make these a standard for future test cases for clairty

api/api_tests/03_auth_sign-out.hurl
```diff
@@ -0,0 +1,46 @@
...
+  "email": "sir.robin@cowardly.co",
+  "password": "RunAway!1234"
+}
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+# Sign-in already signed in without token BAD REQUEST
```
manvinderjit, line 16: query: here, are we trying to sign in again without a token? Is this separate from signing in when we are already signing in?

api/api_tests/01_auth_sign-up.hurl
madcampos: @manvinderjit yes, it was discussed before. This is for security reasons so an attacker cannot guess the emails in the platform.

api/api_tests/04_auth_heartbeat.hurl
```diff
@@ -0,0 +1,93 @@
...
+
+HTTP 401
+
+
+# Heartbeat check with no token UNAUTHORIZED
+GET {{host}}/api/auth/heartbeat
+
+HTTP 401
+
+
```
manvinderjit, line 14: nitpick: Would recommend we add before /after or setup / teardown comments for clarity.

manvinderjit: nitpick: Would recommend we add before /after or setup / teardown comments for clarity.

```suggestion
# Before heartbeat check
```

api/api_tests/05_profile_get.hurl
```diff
@@ -0,0 +1,68 @@
...
+
+# GET profile using non existant id NOT_FOUND
+GET {{host}}/api/profiles/3227114d-43c4-42ed-8aea-f3860fe42122
+
+HTTP 404
+
+
+# Get profile using malformed uuid
+GET {{host}}/api/profiles/00000000-0000-0000-0000-000000000000
+
```
manvinderjit, line 63: nitpick: for clarity, we should specify the status code here as well.

api/api_tests/07_profile_patch.hurl
```diff
@@ -0,0 +1,186 @@
...
+
+{}
+
+HTTP 422
+
+
+POST {{host}}/api/auth/sign-out
+Cookie: auth_auth={{ auth-token-cookie-var }}
+
+HTTP 204
```
manvinderjit: nitpick: please add comments like `before` / `after` for clarity. Example structure:

Before / Setup
...

Tests
...

Expect / Assert
...

After / Teardown
...

api/api_tests/07_profile_patch.hurl
```diff
@@ -0,0 +1,186 @@
...
+  "password": "RunAway!1234"
+}
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+PATCH {{host}}/api/profiles/37a4882c-3b7d-46c0-b019-c37e2445dae8
+Cookie: auth_auth={{ auth-token-cookie-var }}
```
manvinderjit, line 69: query: is this a typo or deliberate? someplaces we are using `Cookie: auth_auth` and other places as `Cookie: auth_token`

api/api_tests/07_profile_patch.hurl
```diff
@@ -0,0 +1,186 @@
...
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+PATCH {{host}}/api/profiles/890cae10-7bbd-4f0c-961a-df495ec7527b
+Cookie: auth_token={{ auth-token-cookie-var }}
+
+Content-Type: applicaion/json
```
manvinderjit: typo here:
```suggestion
Content-Type: application/json
```

api/api_tests/08_profile_delete.hurl
```diff
@@ -0,0 +1,86 @@
...
+  "email": "knight.ni@forestsayni.com",
+  "password": "NiNiNiNi42?!"
+}
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+DELETE {{host}}/api/profiles/37a4882c-3b7d-46c0-b019-c37e2445dae8
+Cookie: auth_auth={{ auth-token-cookie-var }}
```
manvinderjit: query: again, is this `auth_auth` a typo or deliberate?

api/api_tests/11_team_create.hurl
```diff
@@ -0,0 +1,127 @@
...
+{
+	"name":"This time Quoison will rise!"
+}
+
+HTTP 401
+
+# Create team with volunteer account FORBIDEN
+POST {{host}}/api/auth/sign-in
+
+Content-type: application/json
```
manvinderjit, line 13: typo? Not sure on if this impacts tests, but here `t` of `Content-type` is lower case.
```suggestion
Content-Type: application/json
```
If this is an issue, then it occurs multiple times across this file.

api/api_tests/12_patch_patch.hurl
```diff
@@ -0,0 +1,190 @@
+# Patch team information with no session token
```
manvinderjit: nitpick: I am assuming, the intention was to name the file as `team_patch ` and not `patch_patch`.

api/api_tests/13_team_delete.hurl
```diff
@@ -0,0 +1,97 @@
+# Delete team with no session token UNAUTHORIZED
+DELETE {{host}}/api/teams/82df0f4a-2a32-4c14-86d9-4b17fc7566e4
+
+HTTP 401
+
+
+# Delele team with volunteer account FORBIDEN
+POST {{host}}/api/auth/sign-in
+
+Content-type: application/json
```
manvinderjit, line 10: typo?: not sure on the functional impact of casing of `Content-type`

api/api_tests/14_team-members_get.hurl
```diff
@@ -0,0 +1,15 @@
+# Get all team members from a team id
+
+GET {{host}}/api/teams/b3410598-ecbc-41be-9f68-925da74bc61/members
+
+HTTP 422
```
manvinderjit, line 6: suggestion: add test case description for clarity. Does it checks that the team doesn't exists or it has no members?

If it is the former, might want to create another test case for a team that exists but has no members. 

If it is the latter, shouldn't the status code be 2xx?

api/api_tests/16_team-members_create.hurl
```diff
@@ -0,0 +1,236 @@
...
+	"email":"lancelot@heroics.inc",
+	"password":"LeapotFaith!"
+}
+
+HTTP 201
+[Captures]
+auth-token-cookie-var: cookie "auth_token"
+
+
+POST {{host}}/api/teams/b3410598-ecbc-41be-9f68-925da74bc613//members
```
manvinderjit: typo: possibly incorrect url. the 404 may be triggered by the api route not being found instead of the request failing
```suggestion
POST {{host}}/api/teams/b3410598-ecbc-41be-9f68-925da74bc613/members
```

api/api_tests/18_team-members_delete.hurl
```diff
@@ -0,0 +1,134 @@
...
+	"841f251a-6197-4282-a527-6db44c6d2215"
+]
+
+HTTP 200
+
+
+POST {{host}}/api/auth/sign-out
+Cookie: auth_token={{ auth-token-cookie-var} }
+
+HTTP 204
```
manvinderjit, line 134: suggest: add a test case to check if a team member can be deleted if there is no auth token present
```suggestion
HTTP 204

# Unauthenticated (No token) deletion attempt
DELETE {{host}}/api/teams/{teamId}/members

Content-Type: application/json
[
    "6f73e09d-8481-4a05-b4a3-524cd86080cd"
]

HTTP 401
```
(Double check on the status code here.)

manvinderjit (commented):
@kbventures Great work. These are quite a lot of test cases, a lot of hours must have gone into them. Just left a few comments related to typos, clarifications, and suggestions.

api/api_tests/01_auth_sign-up.hurl
kbventures: Created issue

https://github.com/torontojs/communityhub/issues/171

api/api_tests/01_auth_sign-up.hurl
kbventures: Added notes.

api/api_tests/03_auth_sign-out.hurl
kbventures: Yes. Hurl requires us to manually include tokens.

api/api_tests/07_profile_patch.hurl
kbventures: Not deliberate! This is a mistake.

api/api_tests/13_team_delete.hurl
kbventures: @manvinderjit 

Generally speaking, they are supported:
https://stackoverflow.com/questions/7718476/are-http-headers-content-type-c-case-sensitive

And so does hurl:
https://github.com/Orange-OpenSource/hurl/issues/639
https://github.com/Orange-OpenSource/hurl/pull/658

api/api_tests/14_team-members_get.hurl
kbventures: This is for an incorrect UUID. I added a note.

manvinderjit (approved):
Excellent work Ken.

manvinderjit:
For this, you can use either Before or Setup. No need of both. Apologies, I should have expressed it with clarity.
image

PR #121, opened 2025-06-05
120 continuous integration

### LINKED ISSUE #120: Continuous Integration
Create CI (Continuous Integration) workflow with Github Actions

## References
- [About continuous integration with GitHub Actions](https://docs.github.com/en/actions/about-github-actions/about-continuous-integration-with-github-actions)
  - https://github.com/actions/starter-workflows/blob/main/ci/node.js.yml

lylaminju:
## Summary

- Added a CI file that installs dependencies and run tests
- The CI is triggered when code is pushed to the `main` or `dev` branches, or when a pull request is opened targeting those branches.

## Additional information

- Although `Husky` handles linting locally, I included a lint check in the CI pipeline to provide a more reliable double-check.

## References
https://github.com/Orange-OpenSource/hurl?tab=readme-ov-file#debian--ubuntu

### DIFF
--- a/.github/ci.yml
+++ b/.github/ci.yml
@@ -24,6 +24,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Run lint
+        run: npm run lint
+
       # Optional: Only if needed for testing later
       # - name: Build project
       #   run: npm run build
--- a/.github/ci.yml
+++ b/.github/ci.yml
@@ -10,40 +10,22 @@ jobs:
   lint:
     name: 🔎 Lint
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 22
-          cache: "npm"
-      - run: npm ci
+      - uses: ./.github/actions/setup-node-deps
       - run: npm run lint
 
   unit-tests:
     name: 🧪 Unit Tests
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 22
-          cache: "npm"
-      - run: npm ci
+      - uses: ./.github/actions/setup-node-deps
       - run: npm test
 
   integration-tests:
     name: 🧪 Integration Tests
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 22
-          cache: "npm"
-      - run: npm ci
+      - uses: ./.github/actions/setup-node-deps
 
       # Optional: Only if needed for testing later
       # - name: Build project
--- a/.github/ci.yml
+++ b/.github/ci.yml
@@ -48,11 +48,7 @@ jobs:
       - name: Run integration tests
         run: |
           set -e
-          find . -name "*.hurl" | while read file;
-          do
-            echo "Running: $file";
-            hurl "$file";
-          done
+          hurl --test --variable host=http://localhost:3000 '**/*.hurl'
 
       - name: Kill preview server
         if: always()
--- a/.github/ci.yml
+++ b/.github/ci.yml
@@ -7,8 +7,8 @@ on:
     branches: [main, dev]
 
 jobs:
-  test:
-    name: 🔎 Test
+  lint:
+    name: 🔎 Lint
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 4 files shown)


manvinderjit:
> Just my suggestions, @madcampos and @kbventures might have more insights on this.

nitpick: The repo already has some bash script files in the `.husky` folder. We can leverage them in the yaml file instead of writing commands. 
We can also create bash files to segregate the commands for different purposes, like put all the integration testing commands in a single integration-tests.sh bash file.

.github/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+            echo "Running: $file";
+            hurl "$file";
+          done
+
+      - name: Kill preview server
+        if: always()
+        run: |
+          if [ -f preview.pid ]; then
+            kill $(cat preview.pid)
+          fi
```
madcampos: Let's put this on a different job, so each type of testing runs on it's own and if things fail we can re-run only one thing like unit tests instead of running everything.

Another question: can we cache/reuse steps? Like, the step for installing node and installing dependencies is repeated multiple times, could we cache that?

wrangler.toml
```diff
@@ -19,6 +19,7 @@ enabled = true
 head_sampling_rate = 1
 
 [assets]
+directory = "public"
```
madcampos: "Oops..." I was trigger happy and remove this line, sorry.

.github/ci.yml
lylaminju: > Another question: can we cache/reuse steps? Like, the step for installing node and installing dependencies is repeated multiple times, could we cache that?

Yes, great observation! I'm already using `cache: "npm"` in the setup-node action, which caches npm dependencies between runs. 

To reduce the duplicate setup code across jobs, I'll create a [Composite Action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action?platform=mac) that bundles the checkout, node setup, and npm install steps. This will make the workflow much cleaner and more maintainable.

madcampos:
@manvinderjit 
> > Just my suggestions, @madcampos and @kbventures might have more insights on this.
> 
> nitpick: The repo already has some bash script files in the `.husky` folder. We can leverage them in the yaml file instead of writing commands. We can also create bash files to segregate the commands for different purposes, like put all the integration testing commands in a single integration-tests.sh bash file.

I'm going to be very annoying here...

The bash scripts on the `.husky` folder are for git hooks, even if there is some duplication, we should not mess with them. They basically just call something when we do something on git.

What we can do however is update the `scripts` section on the `package.json` file to include whatever script we would like, that is better and it also allows anyone to run the scripts independently.

On adding shell scripts I will say the same thing I said to Ken before: no. 🙃

I would like to avoid shell scripts in particular because they are not portable and don't work well across OSes.

I think for most direct things we can use `npm run ...` and set the scripts we want. Note that every script can also have two other attached scripts to it: `pre<name>` and `post<name>`, and they can run multiple commands using `&&` (this works cross OS).

If we need more complex setups, that may be the case for writing build scripts, on that case, we should write them in js so they can run with node and we don't introduce a new dependency, and we can add them to the `build` folder.

kbventures (changes requested):
@lylaminju 

The scripts will need include to include  {{host}} variable for the hurl scripts to work. 

Currently we are passing the host value using a terminal command paramater "--variable host=localhost:3000"

```
set -e
find . -name "*.hurl" | while read file;
do
  echo "Running: $file";
  hurl --variable host=http://localhost:3000 "$file";
done
```

madcampos:
@kbventures does hurl accept glob parameters for files, or a folder?

That would be nice if we could do it 🥲

kbventures:
@lylaminju @madcampos 

```
Variables can be injected in a Hurl file:

by using [--variable option](https://hurl.dev/docs/manual.html#variable)
by using [--variables-file option](https://hurl.dev/docs/manual.html#variables-file)
by defining environment variables, for instance HURL_foo=bar
by defining variables in an [[Options] section](https://hurl.dev/docs/request.html#options)
https://hurl.dev/docs/templates.html
```

I'm guessing we're going then environment variable route.

madcampos:
@kbventures Sorry for the confusion, but my previous commentary was actually about [file globbing](https://en.wikipedia.org/wiki/Glob_(programming)), like, how to pass a directory or a list of files like `src/**/*.hurl`, instead of _individual_ files.

Did some search and found it on the hurl docs: https://hurl.dev/docs/manual.html#glob

About variables, I think it will be less messy to load a `.env` file: https://hurl.dev/docs/manual.html#variables-file

lylaminju:
@kbventures @madcampos Thanks for the suggestions!  
Since we only have a single `host` variable for now, I opted not to create a `.env` file.
If we add more variables in the future, setting up an environment file would definitely make things cleaner.

madcampos (dismissed):

madcampos (approved):

kbventures (approved):

PR #126, opened 2025-06-07
Implement Forgot Password Functionality

### LINKED ISSUE #119: Forgot Password Page
1) Go to http://localhost:3000/pages/sign-in/
2) Select "I don't remember my password link"

**result:** browser goes to http://localhost:3000/pages/forgot-passsword and it is a 404 page.

ALSO there seems to be a  typo. Look closely. It is "/forgot-passsword" when it the correct spelling is "forgot-password".  

"http://localhost:3000/pages/forgot-password" was tried and it also goes to a 404 page

Ekep-Obasi:
Closes: #119 

## Summary

This PR implements Forgot Password Functionality

## Tests

- [x] Integration tests
- [ ] Unit Tests

## Additional information

This PR introduces a `useAsync` hook to make handling async operations more consistent across our components.

### **How `useAsync` works in `ForgotPasswordForm`**
I used the hook to manage password recovery requests. It handles the loading states, errors, and success callbacks automatically, so the component can focus on what it needs to do when things succeed or fail (like starting countdown timers or showing error messages). There's also a handy `reset` function to clear everything when needed.

This approach keeps the async stuff separate from the UI logic, which makes the code cleaner and easier to follow (my suggestion ?)

### **What we were doing before**
Previously, components like `SignInForm` were managing all their async operations directly inside the component. This meant we were writing the same loading state logic, error handling, and API calls over and over again in different places.


src/components/Header/Header.tsx
```diff
@@ -1,17 +1,18 @@
...
 	isLogoLarge?: boolean;
+	textProps?: ComponentProps<'h1'>;
 }
 
-const Header = ({ children, isLogoLarge = true }: Props) => (
+const Header = ({ children, isLogoLarge = true, textProps }: Props) => (
 	<header id='header-with-logo'>
 		<Logo isLarge={isLogoLarge} />
-		<HeaderTitle>{children}</HeaderTitle>
+		<HeaderTitle {...textProps}>{children}</HeaderTitle>
```
manvinderjit, line 15: Out of curiosity: Is there an advantage or case where passing css styles as props is preferred instead of specifying the styles directly in `HeaderTitle.css`

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
+.password-recovery-form {
+	border: 1px solid #cccccc;
```
manvinderjit: suggestion:  There is a variable available in `src/csslayers/design-system.css` layer for border styles. Recommend using this for easy adjustment in future.
`73 --border: var(--border-base) solid var(--color-border);`

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.input-block label {
+	color: #333333;
+	font-weight: var(--weight-semibold);
+}
+
+.input-block input {
+	background-color: var(--color-button);
+	border: solid 1px #cccccc;
```
manvinderjit: Please use border variable here as well.

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.input-block {
+	display: flex;
+	flex-direction: column;
+	gap: var(--spacing-xs);
+}
+
+.input-block label {
+	color: #333333;
```
manvinderjit: suggestion: For colors, there are variables available in the `design-system.css`. Please refer to these. 

image

manvinderjit: If a color is unavailable, please add it here or ask the design team for suggestions.

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.success-message {
+	border-radius: var(--rounded-xs);
+	text-align: center;
+}
+
+hr {
+	border: none;
+	border-top: 1px solid #cccccc;
```
manvinderjit: It would be better to use individual variables here. Please do the same. 
`var(--border-base) solid var(--color-border);`

src/css-layers/pages.css
```diff
@@ -1,14 +1,18 @@
...
 }
 
 .page {
 	align-items: center;
 	display: flex;
 	flex-direction: column;
+	height: inherit;
 	justify-content: center;
-	max-width: 400px;
+	width: clamp(20rem, 90vw, 25rem);
```
manvinderjit: suggestion: `16.875rem` is the minimum I would recommend so that it covers devices with width as low as 320px, like older Androids and iPhones. Not sure on the max limit though as it varies by page.

manvinderjit:
Praise: This is excellent work @Ekep-Obasi . Really well written and documented code.

Ekep-Obasi:
@manvinderjit, thanks for the reviews. I'll address them asap.

src/components/Header/Header.tsx
Ekep-Obasi: Since the component is used across most auth pages, I needed to make it more flexible for my use-case and couldn't think of a way to manage that just with css

src/css-layers/pages.css
Ekep-Obasi: Agreed, thanks for the suggestion!

src/components/Header/Header.tsx
kbventures: @madcampos

src/css-layers/pages.css
```diff
@@ -1,14 +1,18 @@
+#root { height: inherit; }
```
kbventures: Question: can you explain rational behind this change that will affect all pages?

api/src/routes/auth/integration-test.hurl
```diff
@@ -1,5 +1,5 @@
 # Sign in intergration test
```
kbventures: There have been major changes made to how we do intergration testings.

https://github.com/torontojs/communityhub/pull/117/files

Please refer to this file and update your tests accordingly. 

There's no need to assert code in eveyr case you can simply write HTTP 200 in a lot of cases.

src/hooks/useAsync.ts
```diff
@@ -0,0 +1,59 @@
+import { useCallback, useEffect, useState } from 'react';
```
kbventures: I'm in favor of creating a hook, especially if it can be reused across pages. That said, we should also consider maintainability, loose coupling, ease of use, and testability.

Can we achieve creating a reusable adaptable hook that is easy to read, maintain and tests and can we dig deeper since this hook as why we NEED it considering it contains more code  than many pages in the front end so it probably deserves more explanation/rational  to support why it should exists.

src/components/Header/Header.tsx
madcampos: @manvinderjit that depends. For super generic components like `Card` or `ColumnLayout`, it is a valid way to customize the component. If you are using css-in-js, it also makes sense as this is how the thing to do. But css-in-js is a terrible idea.

On this case as the `HeaderTitle` is very consistent and only have small changes. It doesn't make much sense to change it.

@Ekep-Obasi maybe it makes more sense to use another component here or confirming with the design team if it is okay to use the same component here?

api/src/index.ts
```diff
@@ -67,7 +67,7 @@ apiRoutes.doc('/open-api.json', {
 	}
 });
 
-apiRoutes.get('/docs', swaggerUI({ url: '/open-api.json' }));
+apiRoutes.get('/docs', swaggerUI({ url: '/api/open-api.json' }));
```
madcampos, line 65: **praise:** Nice catch‏

api/src/routes/auth/index.ts
```diff
@@ -221,6 +221,64 @@ authRoutes.openapi(
...
+			[StatusCodes.INTERNAL_SERVER_ERROR]: {
+				description: 'Internal server error occurred while processing the password recovery request.',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
+		}
+	}),
+	async (context) => {
+		const { email } = context.req.valid('json');
+
+		const response = { message: 'If an account with that e-mail exists, a password reset link has been sent.' };
```
madcampos, line 266: **suggestion:** Let's add a comment here about not exposing existing emails and security. Like‏ the one we have on other auth endpoints.

api/src/routes/auth/index.ts
```diff
@@ -221,6 +221,64 @@ authRoutes.openapi(
...
+
+		const resetToken = crypto.randomUUID();
+		// eslint-disable-next-line @typescript-eslint/no-magic-numbers
+		const TEN_MINUTES_IN_SECONDS = 60 * 10;
+
+		await context.env.PasswordResetTokens.put(
+			resetToken,
+			email,
+			{ expirationTtl: TEN_MINUTES_IN_SECONDS }
+		);
```
madcampos: **issue:** Please move this code to `api/utils/‏auth.ts`, that is where the code that handles authentication lives. I think we even have some constants for timeout there.

api/src/email/index.ts
```diff
@@ -57,3 +57,32 @@ export async function sendAccountConfirmationEmail(context: Context, {
...
+		to: email,
+		subject: '[TorontoJS] Reset your password',
+		text: `You requested a password reset.\n\nPlease reset your password by visiting: ${resetUrl}`,
+		html: `
+			<h2>You requested a password reset</h2>
+			<p>Please reset your password by visiting the link below</p>
+			<p><a href="${resetUrl}">${resetUrl}</a></p>
+		`
+	});
+}
```
madcampos, line 88: **issue:** ‏You already have `context` as one of the parameters, the environment variables should be inferred from it here and not passed as parameters.

This provides better encapsulation for the consumer and don't expose implementation details. If we change from Resend to another email provider, we would only have to change this function implementation and not every call to it.

```suggestion
export async function sendPasswordResetEmail(context: Context, token: string, email: string) {
	const resetUrl = new URL(`/pages/reset-password?token=${token}`, context.req.url).toString();

	return sendEmail(context, {
		apiKey: context.env.RESEND_API_KEY,
		from: context.env.SENDER_EMAIL,
		to: email,
		subject: '[TorontoJS] Reset your password',
		text: `You requested a password reset.\n\nPlease reset your password by visiting: ${resetUrl}`,
		html: `
			<h2>You requested a password reset</h2>
			<p>Please reset your password by visiting the link below</p>
			<p><a href="${resetUrl}">${resetUrl}</a></p>
		`
	});
}
```

api/src/routes/auth/integration-test.hurl
```diff
@@ -80,3 +80,63 @@ status == 404
...
+Content-Type: application/json
+
+{
+  "email": "profile1@example.com"
+}
+
+HTTP 200
+[Asserts]
+status == 200
+jsonpath "$.message" == "If an account with that e-mail exists, a password reset link has been sent."
```
madcampos: **question:** I think hurl already does the assertion of the http status code from the `HTTP ...` line. Do we really need these?‏

Also, here just testing for the response is enough. Asserting on the message is brittle as if we change it in the future we would also have to fix the test. The message itself is not relevant, as it is a hint for consumers of the API, what really says if things or not in this case is the status code.

api/src/routes/auth/integration-test.hurl
```diff
@@ -80,3 +80,63 @@ status == 404
...
+Content-Type: application/json
+
+{
+  "email": "nonexistent@example.com"
+}
+
+HTTP 200
+[Asserts]
+status == 200
+jsonpath "$.message" == "If an account with that e-mail exists, a password reset link has been sent."
```
madcampos: Same thing here.

api/src/routes/auth/integration-test.hurl
```diff
@@ -80,3 +80,63 @@ status == 404
...
+POST http://localhost:3000/api/auth/forgot-password
+Content-Type: application/json
+
+{
+  "email": "invalid-email"
+}
+
+HTTP 422
+[Asserts]
+status == 422
```
madcampos: Same here

api/src/routes/auth/integration-test.hurl
```diff
@@ -80,3 +80,63 @@ status == 404
...
+
+# Forgot password with missing email field
+POST http://localhost:3000/api/auth/forgot-password
+Content-Type: application/json
+
+{}
+
+HTTP 422
+[Asserts]
+status == 422
```
madcampos: Same here

api/src/routes/auth/integration-test.hurl
```diff
@@ -80,3 +80,63 @@ status == 404
...
+
+# Forgot password with missing email field
+POST http://localhost:3000/api/auth/forgot-password
+Content-Type: application/json
+
+{}
```
madcampos: **praise:** Nice test‏

api/src/routes/auth/integration-test.hurl
```diff
@@ -80,3 +80,63 @@ status == 404
...
+POST http://localhost:3000/api/auth/forgot-password
+Content-Type: application/json
+
+{
+  "email": ""
+}
+
+HTTP 422
+[Asserts]
+status == 422
```
madcampos: Same here

api/src/routes/auth/validation.ts
```diff
@@ -30,3 +30,11 @@ export type SignUpData = z.infer<typeof SignUpSchema>;
 export const ActivateSchema = z.object({ token: z.string().uuid('Invalid ID format') });
 
 export type ActivateData = z.infer<typeof ActivateSchema>;
+
+export const ForgotPasswordSchema = z.object({
+	email: z.string({ required_error: 'Email is required' })
+		.min(1, 'Email must be at least one character long')
+		.email('This is not a valid e-mail address')
```
madcampos: **nitpick:** [`email`](https://zod.dev/api?id=emails) already tests for a non empty string‏
```suggestion
	email: z.string({ required_error: 'Email is required' })
		.email('This is not a valid e-mail address')
```

wrangler.toml
```diff
@@ -68,6 +73,11 @@ binding = "ActivationTokens"
 id = "4fc17eeab6964947ba5ee0d50aeec1de"
 preview_id = "vms-dev-activation-tokens"
 
+[[env.dev.kv_namespaces]]
+binding = "PasswordResetTokens"
+id = "682d719e25304b57a5c95db6815da556"
```
madcampos: **question:** Where does this id come from?‏

wrangler.toml
```diff
@@ -89,3 +99,8 @@ preview_id = "vms-production-session-tokens"
 binding = "ActivationTokens"
 id = "987a04058cc34aad84b9f54c658f6a01"
 preview_id = "vms-production-activation-tokens"
+
+[[env.production.kv_namespaces]]
+binding = "PasswordResetTokens"
+id = "f917eef00a5f4146a2d8d8fc2d274fba"
```
madcampos: **question:** Where does this id come from?‏

src/hooks/useCountDown.ts
```diff
@@ -0,0 +1,31 @@
+import { useCallback, useEffect, useState } from 'react';
+
+const ONE_MINUTE_IN_MS = 60000;
```
madcampos: **nitpick:** ‏
```suggestion
const ONE_MINUTE_IN_MS = 60 * 1000;
```

src/hooks/useCountDown.ts
```diff
@@ -0,0 +1,31 @@
...
+		setMinutes(mins);
+		setIsActive(true);
+	}, [initialMinutes]);
+
+	return {
+		minutes,
+		isFinished: minutes === 0,
+		start
+	};
+};
```
madcampos: **issue:** Use [`setInterval`](https://developer.mozilla.org/en-US/docs/Web/API/Window/setInterval) instead of `setTimeout`, it does wat you want[^1] with less logic, i.e. not needing to keep the `isActive` state.

Here is an example:
```typescript
export function useCountdown(initialTime) {
  const [timeRemaining, setTimeRemaining] = useState(initialTime);

  /**
   * Reset the timer and start a new countdown
   */
  const startCountdown = () => {
    setTimeRemaining(initialTime);

    if (timeRemaining > 0) {
      const interval = setInterval(() => {
        if (timeRemaining === 0) {
          clearInterval(interval);
          return;
        }
  
        setTimeRemaining(timeRemaining -1);
      }, ONE_MINUTE_IN_MS);
    }
  }

  return {
    timeRemaining,
    startCountdown
  };
}
```

Also a nitpick: prefer function declaration over arro functions whenever possible.

[^1]: One caveat to `setInterval`, and `setTimeout` for that matter, is they are not exactly precise, so there is drift in the time for both of them. But, for our use case they are good enough.

src/css-layers/pages.css
madcampos: **question:** ‏Why all the `inherit`ed `height`s?

A flex layout should already take all of the space available compared to flow layout. If you really really want to make the page take all the available space use this instead:

```css
html, body { height: 100%; }
body { margin: 0; }
```

All the flex containers should take all the available vertical space after that.

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+
+.recovery-content {
+	display: flex;
+	flex-direction: column;
+	gap: var(--spacing-md);
+	padding: var(--spacing-sm);
+}
+
+.recovery-content p {
+	font-size: 1.125rem;
```
madcampos: **question:** Don't we already have a variable for that?‏

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
+.password-recovery-form {
```
madcampos: **issue:** As a general rule of thumb: if the element will be repeated multiple times, use a class, if it is going to be only one of it, use an id.‏

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+	margin: 1px 0;
+}
+
+.form-footer {
+	display: flex;
+	flex-direction: column;
+	gap: var(--spacing-md);
+}
+
+.success-message p {
```
madcampos, line 85: **issue:** This rule is a du[licate of another one, if you want to change the `<p>` elements in an entire block, target them in like this:‏
```
<grandparent element> p { ... }
```

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+	align-items: center;
+	color: var(--color-error);
+	display: flex;
+	font-size: 0.875rem;
+	gap: var(--spacing-xs);
+}
+
+.error-message::before {
+	color: var(--color-red);
+	content: '❌';
```
madcampos: **issue:** This content is not accessible and hard to style. Please prefer an inline svg instead.‏

Here is a place to find svg icons: https://icon-sets.iconify.design/

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+	color: var(--color-foreground);
+	font-weight: var(--weight-semibold);
+}
+
+.input-block input {
+	background-color: var(--color-button);
+	border: var(--border);
+	border-radius: var(--rounded-xs);
+	box-sizing: border-box;
+	color: #333333;
```
madcampos: **issue:** Please use variables for colors.‏

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.input-block input.error {
+	background-color: #fff5f5;
+	border-color: var(--color-error);
+}
+
+.input-block input:disabled {
+	background-color: #e5e5e5;
+	color: #999999;
```
madcampos: **issue:** Please use variables for colors.‏

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,110 @@
...
+	width: 100%;
+}
+
+.input-block input:focus {
+	border-color: var(--color-error);
+	outline: none;
+}
+
+.input-block input.error {
+	background-color: #fff5f5;
```
madcampos: **issue:** Please use variables for colors.‏

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,139 @@
...
+						name='email'
+						value={email}
+						autoComplete='email'
+						placeholder='Your account e-mail'
+						required
+						ref={emailInputRef}
+						onChange={handleEmailChange}
+						className={validationError ? 'error' : ''}
+						disabled={isSubmitted}
+						aria-invalid={!!validationError}
```
madcampos: **nitpick:** Do not use double negation for casting a value to boolean, use proper checks.‏

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,139 @@
...
+						)}
+				</div>
+
+				<div className='form-footer'>
+					<Button
+						type='submit'
+						isLarge
+						isPrimary
+						disabled={isLoading || isSubmitted || isEmailEmpty}
+						aria-describedby={isSubmitted ? 'countdown-status' : undefined}
```
madcampos: **thought:** ‏This may be confusing, no need to add it here as the timer is below on the page anyways.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,139 @@
...
+					>
+						{isLoading ? 'Sending...' : 'Send recovery link'}
+					</Button>
+
+					{isSubmitted && (
+						<>
+							<div className='success-message' role='status' aria-live='polite'>
```
madcampos: **issue:** Please remove the `role` and `aria-live` attributes they are excessive.‏

Adding it will make screen readers read it, but not necessarily in a sensible order, and because react destroys and recrates the dom, it will happen every re-render.

Just remove it as a user can read the page again. This is a simple enough page that skimming through the content of the page again should be fine.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,139 @@
...
+								<p>Check your e-mail. We've sent you a recovery link. If you didn't get the link, check your spam folder or try again in 10 minutes.</p>
+							</div>
+							<hr />
+							<div className='resend-timer' id='countdown-status' aria-live='polite'>
+								<ClockIcon aria-hidden='true' />
+								<p>Wait {minutes} {minutes === 1 ? 'minute' : 'minutes'} for a new activation link</p>
```
madcampos: **issue:** Do not do internationalization by hand, use [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules)‏ instead.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,139 @@
...
+						{isLoading ? 'Sending...' : 'Send recovery link'}
+					</Button>
+
+					{isSubmitted && (
+						<>
+							<div className='success-message' role='status' aria-live='polite'>
+								<p>Check your e-mail. We've sent you a recovery link. If you didn't get the link, check your spam folder or try again in 10 minutes.</p>
+							</div>
+							<hr />
+							<div className='resend-timer' id='countdown-status' aria-live='polite'>
```
madcampos: **issue:** Accessible timers are _very tricky_ to get right. I suggest for now to remove the `aria-live` attribute here and research more on how to do this properly.‏

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,139 @@
...
+	});
+
+	const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault();
+		setValidationError('');
+
+		if (!emailInputRef.current?.checkValidity()) {
+			setValidationError('This is not a valid email address');
+			emailInputRef.current?.focus();
+			return;
```
madcampos: **issue:** The validation message here is redundant as the constraint validation api for forms already adds a message and displays it to the user.

Also, moving the focus unprompted can be a failure on [WCAG SC 3.2.1](https://www.w3.org/WAI/WCAG21/Understanding/on-focus.html).

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,139 @@
...
+					<input
+						id='email-input'
+						type='email'
+						name='email'
+						value={email}
+						autoComplete='email'
+						placeholder='Your account e-mail'
+						required
+						ref={emailInputRef}
+						onChange={handleEmailChange}
```
madcampos: **thought:** If we are not doing any validation with the email, there is no need to handle a change event, one less state for email. Just let html control the field normally.‏

src/hooks/useAsync.ts
```diff
@@ -0,0 +1,59 @@
...
+
+interface UseAsyncState<T, Args extends readonly unknown[]> {
+	data: T | null;
+	error: Error | null;
+	isLoading: boolean;
+	execute(...args: Args): Promise<void>;
+	reset(): void;
+}
+
+export function useAsync<T, Args extends readonly unknown[] = []>(
```
madcampos: **nitpick:** `T` is not a descriptive name, please pick a name that better describes what this generic is.‏

src/hooks/useAsync.ts
madcampos, line 1: I haven't finished reviewing yet, but this code is very complex and convoluted. It seems over engineered to me.

Yes, it is interesting to have a helper for async code, but at the same time, we don't need that many bells and whistles. It needs to:
- Call a function
- Return the data from that function
- Return a potential error
- Update a loading state

In this code it seems like `enabled` and `isLoading` are doing the same thing. Also, we don't need calls to different situations and data types from the async operation, the consumer of this hook should handle and care about the data or error that resulted from the async call, not the hook.

api/src/routes/auth/integration-test.hurl
kbventures: @Ekep-Obasi more changes are coming to intergration testing very soon.  We'll be introducing env variables instead of passing an options argument with terminal commands.  It might make sesnse to do do your intergration/unit testing in a a seperate pull request issue(only this time as we are still setting up Ci and unit/intergreation tests setups.  Once this ocify we can implement test for respective features we in the same PR).

api/src/email/index.ts
Ekep-Obasi: @madcampos, this makes! I just followed the same structure of the `sendAccountConfirmationEmail` here https://github.com/torontojs/communityhub/blob/280ad415e6412abccd9a9541169ab65874cc84a0/api/src/email/index.ts#L39

I'll just modify both here.

src/hooks/useCountDown.ts
Ekep-Obasi: @madcampos, I noticed the setInterval approach might have a cleanup issue. if startCountdown gets called multiple times or the component unmounts, we could end up with multiple running intervals. Would using useEffect with proper cleanup be safer here?

api/src/routes/auth/validation.ts
Ekep-Obasi: I followed this from other instances in the same file. I'll just correct them as well

src/components/ForgotPassword/ForgotPassword.tsx
Ekep-Obasi: @madcampos, this is so cool! thanks 🙌

api/src/email/index.ts
madcampos: Okay. This is probably an oversight there as well. Changing both would be great

src/hooks/useCountDown.ts
madcampos: The `if` check on the `startCountdown` function will verify if the timer is already running. If you call this function multiple times and the timer is already running it will simply continue to run and won't start a new interval.

Looking from another side, the page is not intended to be long lived (i.e. running for hours or days on an open tab) so long tasks and memory leaks are not big issues here.

What we can do to improve the code is move the `startCountdown` outside of the hook so the function and all it's internal variable references will still be there in the closure.

And we could also move the variable binding for the interval outside of the hook, so the reference is always kept regardless. A module keeps the variable bindings so importing the same module again will not introduce new variables and instead will reference the same existing ones.

src/hooks/useAsync.ts
```diff
@@ -0,0 +1,59 @@
...
+	}, []);
+
+	useEffect(() => {
+		if (enabled && asyncFn) {
+			void (execute as () => Promise<void>)();
+		}
+	}, [enabled, execute]);
+
+	return { data, error, isLoading, execute, reset };
+}
```
madcampos:
```suggestion
export function useAsync<Data>(asyncFunction: Function) {
	const [data, setData] = useState<Data | undefined>(undefined);
	const [error, setError] = useState<Error | undefined>(undefined);
	const [isLoading, setIsLoading] = useState(false);

	const execute = async (...args: unknown[]) => {
		try {
			setIsLoading(true);
			setError(undefined);

			const result = await asyncFunction(...args);
			setData(result);
		} catch (err) {
			setError(err);
		} finally {
			setIsLoading(false);
		}
	};

	return { data, error, isLoading, execute };
}
```

Some notes here:
- React hooks will memoize the results, track dependencies and all that. We don't care about it here, the burden of executing a call again should be on the component, so a simple async function will do. The component consuming the code needs to check if they don't have data, nor error and then execute the function.
- Please use `undefined` over `null`, they have subtle differences. The main one being `null instanceof Object === true`.
- Do not overtype things, the compiler is your friend, let it do it's job and infer most of the types for you. Only add types if the compiler cannot figure that out otherwise or you are absolutely certain the thing is what it is expected to be.
- `unknown` can fit any type and is better than `any` (in fact eslint complains about using `any`), it can accept anything and is useful when we don't care about the type of things.
- `await`ing on a regular function will just return the result of that function, it won't cause a problem. Also, deeply nested promises get unwrapped when they are `await`ed.

src/css-layers/pages.css
kbventures: It results in this problem. 

image

The TorontoJS banner is cut off.

api/src/routes/auth/index.ts
```diff
@@ -222,6 +222,64 @@ authRoutes.openapi(
...
+		}
+	}),
+	async (context) => {
+		const { email } = context.req.valid('json');
+
+		const response = { message: 'If an account with that e-mail exists, a password reset link has been sent.' };
+
+		const emailExists = await checkExistingEmail(context.env.Database, email);
+		if (!emailExists) {
+			return context.json(response satisfies StatusResponse, StatusCodes.OKAY);
```
bcha92, line 272: Should this `StatusCodes.OKAY` response be changed to `StatusCodes.BAD_REQUEST` instead since this conditional looks for a falsy `emailExists` value?

Also do we have a default catcher for `StatusCodes.INTERNAL_SERVER_ERROR` or is this initiated by default?

bcha92 (commented):
Other than the one comment about StatusCodes, I think it looks okay.

api/src/routes/auth/index.ts
kbventures: @bcha92 This is because we're trying to obfuscate the existence of emails to stop someone from phishing for information. 

We do not, to the best of my knowledge, have a default catcher for 500 errors.  The only 500 status code caught in our API Tests is from an explicit try-catch instance.

bcha92 (dismissed):

src/hooks/useCountDown.ts
```diff
@@ -0,0 +1,18 @@
...
+	useEffect(() => {
+		if (seconds <= 0) { return; }
+		const timer = setTimeout(() => {
+			setSeconds(seconds - 1);
+		}, MS);
+		return () => clearTimeout(timer);
+	}, [seconds, setSeconds]);
+
+	return { isFinished: seconds <= 0 };
+}
```
madcampos:
```suggestion
import { useEffect, useState } from 'react';

const ONE_SECOND_IN_MS = 1000;

export function useCountdown(seconds: number) {
	const [remaningTime, setRemainingTime] = useState(seconds);

	useEffect(() => {
		const interval = setInterval(() => {
			setRemainingTime(remainingTime - 1);

			if (remainingTime <= 0) {
				clearInterval(interval);
			}
		}, seconds * ONE_SECOND_IN_MS);
	}, [seconds]);

	return { isFinished: remainingTime <= 0, remainingTime };
}
```

- Use `setInterval` instead of `setTimeout`, this is what the function is for.
- Keep the state _inside_ the hook.
- Return the remaining time, it is data from the hook and not external to it.

wrangler.toml
```diff
@@ -51,6 +51,11 @@ binding = "ActivationTokens"
 id = "vms-local-activation-token"
 preview_id = "vms-local-activation-token"
 
+[[kv_namespaces]]
+binding = "PasswordResetToken"
+id = "vms-local-password-reset-token"
+preview_id = "vms-local-password-reset-token"
+
```
madcampos, line 58: **note**: this needs to be repeated for the other envs.

Use this:
```
[[env.dev.kv_namespaces]]
binding = "PasswordResetToken"
id = "5392322d41c44203bce89425fcc12f12"
preview_id = "vms-dev-password-reset-token"

[[env.prod.kv_namespaces]]
binding = "PasswordResetToken"
id = "08ad1ddf3a134007bbed3b9266cd6d1f"
preview_id = "vms-production-password-reset-token"
```

api/src/routes/auth/index.ts
```diff
@@ -225,6 +225,72 @@ authRoutes.openapi(
...
+		const response = { message: 'If an account with that e-mail exists, a password reset link has been sent.' };
+
+		const emailExists = await checkExistingEmail(context.env.Database, email);
+
+		// We are deliberately sending OK statuscodes to disable an attackers ability to fish for information on exisiting emails
+		if (!emailExists) {
+			return context.json(response satisfies StatusResponse, StatusCodes.OKAY);
+		}
+
+		// TODO: Discussion around wether we should create a access entry to deal with this state
```
madcampos: **question**: Haven't we already discussed that? Is it okay to remove this comment?

api/src/routes/auth/index.ts
```diff
@@ -225,6 +225,72 @@ authRoutes.openapi(
...
+		const ForgotPasswordList = await context.env.PasswordResetToken.list();
+
+		for (const key of ForgotPasswordList.keys) {
+			const keyName = await context.env.PasswordResetToken.get(key.name);
+			if (keyName === email) {
+				return context.json(response satisfies StatusResponse, StatusCodes.OKAY);
+			}
+		}
```
madcampos: **issue**: do a get here and check for a null value instead

api/src/routes/auth/index.ts
madcampos, line 1: **question**: When a user resets their password, to which endpoint they send the new password?

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useRef, useState } from 'react';
+
+import './ForgotPassword.css';
+import { useCountdown } from '../../hooks/useCountDown.ts';
+import Button from '../Button/Button.tsx';
+import ClockIcon from '../Icons/ClockIcon.tsx';
+
+// TODO: Investigate if we're able to use heartbeat to check if forgot password request already in progress and how the page react if there is one
```
madcampos, line 6: To answer that: yes, we can add a check to the profile status for the password reset token, if it exists, we return that the user is in the "reset password" state.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,115 @@
...
+					</Button>
+					{isSubmitted && (
+						<>
+							<div className='success-message'>
+								<p>Check your e-mail. We've sent you a recovery link. If you didn't get the link, check your spam folder or try again in 10 minutes.</p>
+							</div>
+							<hr />
+							<div className='resend-timer' id='countdown-status'>
+								<ClockIcon aria-hidden='true' />
+								<p>Wait {Math.ceil(seconds / 60)} minutes for a new activation link</p>
```
madcampos:
```suggestion
								<p>Wait {Math.ceil(seconds / 60)} minutes to send a new activation link</p>
```

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,115 @@
...
+				if (error instanceof Error) {
+					console.error(error.name);
+					console.error(error.cause);
+					console.error(error.message);
+					console.error(error.stack);
+				} else {
+					throw new Error('Invalid email error');
+				}
+			}
+		}
```
madcampos: **nitpick**: please cleanup this error log.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,115 @@
...
+	const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault();
+
+		const formData = new FormData(event.currentTarget);
+		const emailValue = (formData.get('email') as string).trim().toLowerCase();
+
+		setButtonDisabled(true);
+		setEmailDisabled(true);
+
+		requestPasswordRecovery(emailValue);
```
madcampos: **nitpick**: this function is only called here, please move the code here and remove the extra function

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,115 @@
...
+					throw new Error('Invalid email error');
+				}
+			}
+		}
+	};
+
+	const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault();
+
+		const formData = new FormData(event.currentTarget);
```
madcampos:
```suggestion
		const formData = new FormData(event.target as HTMLFormElement);
```

Submit events don't bubble, this should work just fine.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,115 @@
...
+		const emailValue = (formData.get('email') as string).trim().toLowerCase();
```
madcampos, line 55: I think this check should be moved to the backend, everywhere we have emails, we should normalize it to lowercase.

src/components/ForgotPassword/ForgotPassword.css
madcampos: @kbventures please change this to a color variable.

src/components/ForgotPassword/ForgotPassword.css
madcampos: @kbventures here as well

src/components/ForgotPassword/ForgotPassword.css
madcampos: @kbventures here as well

api/api_tests/19_auth_forgot_password.hurl
madcampos, line 1: The file name will conflict with another existing one. Why don't we move the tests to different folders, like this:
- routes
  - route a
    - api-tests
    - ...
  - route b
    - api-test
    - ...

madcampos (changes requested):

src/components/ForgotPassword/ForgotPassword.tsx
kbventures: https://github.com/torontojs/communityhub/pull/199

src/components/ForgotPassword/ForgotPassword.tsx
kbventures: @madcampos This pattern is utilized throughout the project.  Please be more specific.

madcampos: The `console` API allows objects, so the first part should only be:
```ts
console.log(error);
```

Also, re-throwing an error in a `catch` block seems like a code smell. Why are we `catch`ing the error if we are throwing it again? A `catch` block should _handle_ the error there

src/components/ForgotPassword/ForgotPassword.tsx
kbventures: Utilizing .target instead of .currentTarget triggers a TypeScript check error. 

```
Argument of type 'EventTarget' is not assignable to parameter of type 'HTMLFormElement'.
  Type 'EventTarget' is missing the following properties from type 'HTMLFormElement': acceptCharset, action, autocomplete, elements, and 319 more.ts(2345)
```

madcampos: This is a shortcoming of typescript and event handlers. It is very annoying. You should cast to the element in this case.

I've updated the previous comment with the casting.

src/components/ForgotPassword/ForgotPassword.tsx
kbventures: The purpose is to catch errors that aren’t instances of the Error class.

That said, in retrospect, this is overkill for something intended to simplify development and is disabled in production.

Changed all of them.

api/src/routes/auth/data.ts
```diff
@@ -162,3 +162,11 @@ export async function activateProfile(database: D1Database, email: string) {
 
 	return success;
 }
+
+export async function resetPassword(database: D1Database, password: string, email: string) {
+	const { success } = await database
+		.prepare(`UPDATE ${DBTables.PROFILE} SET password = ? WHERE email ?`)
```
madcampos: **question:** Should we check as well if the profile is activated and not deleted?

api/src/routes/auth/validation.ts
```diff
@@ -78,3 +78,20 @@ export type SignUpData = z.infer<typeof SignUpSchema>;
...
+		.trim()
+		.toLowerCase()
+		.min(1)
+});
+
+export type ForgotPasswordData = z.infer<typeof ForgotPasswordSchema>;
+
+export const ResetPasswordSchema = z.object({
+	password: z.string({ error: 'Password is required' })
+		.min(15),
```
madcampos: **issue:** Do not use a magic number. Without context, a plain number doesn't have much meaning.

Once the text fixes PR is merged, you can export this constant and reference it: https://github.com/torontojs/communityhub/blob/ed929ca6edef011324f9de9410db9c74876afc0a/api/src/utils/passwordStrengthCheck.ts#L4

src/components/ResetPassword/ResetPassword.css
```diff
@@ -0,0 +1,127 @@
...
+		margin-top: 0.5rem;
+
+		&[data-password-fail=''] { display: none; }
+
+		p { color: var(--color-error); }
+	}
+}
+
+.suggestion-icon {
+	background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='none'%3E%3Cpath d='M19 3H5C3.89543 3 3 3.89543 3 5V19C3 20.1046 3.89543 21 5 21H19C20.1046 21 21 20.1046 21 19V5C21 3.89543 20.1046 3 19 3Z' stroke='black' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'/%3E%3Cpath d='M12 8V16' stroke='black' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'/%3E%3Cpath d='M8.5 14L15.5 10' stroke='black' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'/%3E%3Cpath d='M8.5 10L15.5 14' stroke='black' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'/%3E%3C/svg%3E");
```
madcampos, line 107: **issue:** Same as this comment https://github.com/torontojs/communityhub/pull/173#discussion_r2217986531

Also this comment on how to solve implement it: https://github.com/torontojs/communityhub/pull/173#discussion_r2223609295

src/components/ResetPassword/ResetPassword.css
```diff
@@ -0,0 +1,127 @@
...
...:%23010101;%7D%3C/style%3E%3C/defs%3E%3Ctitle%3Ewarning%3C/title%3E%3Cpath class='cls-1' d='M2.35,84.42,45.28,10.2l.17-.27h0A23,23,0,0,1,52.5,2.69,17,17,0,0,1,61.57,0a16.7,16.7,0,0,1,9.11,2.69,22.79,22.79,0,0,1,7,7.26q.19.32.36.63l42.23,73.34.24.44h0a22.48,22.48,0,0,1,2.37,10.19,17.63,17.63,0,0,1-2.17,8.35,15.94,15.94,0,0,1-6.93,6.6c-.19.1-.39.18-.58.26a21.19,21.19,0,0,1-9.11,1.75v0H17.61c-.22,0-.44,0-.65,0a18.07,18.07,0,0,1-6.2-1.15A16.42,16.42,0,0,1,3,104.24a17.53,17.53,0,0,1-3-9.57,23,23,0,0,1,1.57-8.74,7.66,7.66,0,0,1,.77-1.51Z'/%3E%3Cpath class='cls-2' d='M9,88.75,52.12,14.16c5.24-8.25,13.54-8.46,18.87,0l42.43,73.69c3.39,6.81,1.71,16-9.33,15.77H17.61C10.35,103.8,5.67,97.43,9,88.75Z'/%3E%3Cpath class='cls-3' d='M57.57,83.78A5.53,5.53,0,0,1,61,82.2a5.6,5.6,0,0,1,2.4.36,5.7,5.7,0,0,1,2,1.3,5.56,5.56,0,0,1,1.54,5,6.23,6.23,0,0,1-.42,1.35,5.57,5.57,0,0,1-5.22,3.26,5.72,5.72,0,0,1-2.27-.53A5.51,5.51,0,0,1,56.28,90a5.18,5.18,0,0,1-.36-1.27,5.83,5.83,0,0,1-.06-1.31h0a6.53,6.53,0,0,1,.57-2,4.7,4.7,0,0,1,1.14-1.56Zm8.15-10.24c-.19,4.79-8.31,4.8-8.49,0-.82-8.21-2.92-29.34-2.86-37.05.07-2.38,2-3.79,4.56-4.33a12.83,12.83,0,0,1,5,0c2.61.56,4.65,2,4.65,4.44v.24L65.72,73.54Z'/%3E%3C/svg%3E");
```
madcampos, line 119: Same here as the previous comment

src/components/TextInputComponent/TextInputComponent.css
```diff
@@ -13,18 +13,19 @@
...
 
 		& > input {
-			background-color: var(--color-card);
+			font-size: var(--text-xs);
+			background-color: var(--color-button);
 			border: var(--border);
 			border-radius: var(--rounded-xs);
 			gap: var(--spacing-xs);
-			height: auto;
+			height: 42px;
```
madcampos: **issue:** Do not use pixel values. Furthermore, it is an input, which by default has a `display` of `inline`. Inline elements don't compute `height` and `width` like `block` elements. So this won't make much of a difference and may actually be prejudicial.

src/components/TextInputComponent/TextInputComponent.css
```diff
@@ -49,3 +50,30 @@
...
+.input-block input.error {
+	background-color: var(--color-gray-50);
+	border-color: var(--color-error);
+}
+
+.input-block input:disabled {
+	background-color: #e5e5e5;
+	color: var(--color-gray-500);
+	cursor: not-allowed;
+} */
```
madcampos: **nitpick**: Please clean up commented code, we want to keep the repo tidy. Also, there seems to be a typo in the first two lines.

src/index.html
```diff
@@ -10,6 +10,8 @@
 		<h1>VMS pages</h1>
 		<ul>
 			<li><a href="/pages/sign-in/">Sign-In Page</a></li>
+			<li><a href="/pages/forgot-password/">Forgot Password Page</a></li>
+			<li><a href="/pages/reset-password/?token=cb81a6dc-7328-4e7a-b0a1-14e83ccb3468">Reset Password Page</a></li>
```
madcampos: **question**: how will this token work? Doesn't the page check if the token is valid when loading the page?

src/pages/reset-password/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<title>Sign-In</title>
```
madcampos: **issue**: the title here is wrong

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,141 @@
...
+	const [feedback, setFeedback] = useState<string>('');
+	// const [invalidPasswordMessage, setInvalidPasswordMessage] = useState<string>('');
+
+	const token = new URLSearchParams(window.location.search).get('token');
+
+	if (token == null) {
+		window.location.href = '/pages/sign-in';
+	}
+
+	// TODO how to handle errors from server
```
madcampos: **question**: what kinds of errors?

We can simply show a message to users when there is an error.

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,141 @@
...
+		const passwordValue = formData.get('password') as string;
+
+		setButtonDisabled(true);
+		setPasswordDisabled(true);
+
+		await resetPassword(token, passwordValue);
+	};
+
+	const resetPassword = async (token: string | null, password: string) => {
+		// setInvalidPasswordMessage('');
```
madcampos: **nitpick**: please remove unused code to keep the repo tidy.

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,141 @@
...
+			});
+			if (!response.ok) {
+				window.location.href = '/pages/sign-in';
+			} else {
+				window.location.href = '/pages/sign-in';
+			}
+		} catch (error) {
+			if (import.meta.env.MODE === 'development') {
+				console.log(error);
+			}
```
madcampos: **thought**: I really don't like this pattern of error reporting.
We should just use `console.error` (not `log`) to log the messages to the console anyways, regardless of environment...

This helps with debugging for tech savy users.

For non tech users we should provide a modal dialog or something with an error message, a reference code, and some UI that the users can use to help report the error.

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,141 @@
...
+						required
+						aria-describedby='password-input-strength password-input-suggestion password-input-helper-text'
+						ref={passwordInputRef}
+					/>
+					<div className='passwordError' hidden={strength === null}>
+						<div className='text-size'>
+							<div id='password-input-strength' aria-live='polite'>
+								<span>Password strength: {strengthLabels[strength || 0]}</span>
+							</div>
+							<div className='password-meter' data-password-strength={strengthLabels[strength || 0]} aria-hidden='true'>
```
madcampos, line 139: **question**: why are we hiding the password meter?

madcampos: The best markup for a password meter would be:
```jsx
<label for="password-meter">Password trength</label>
<div id="password-meter-wrapper" data-password-strength={strengthLabels[strength || 0]}>
  <meter
    className="sr-only"
    id="password-meter"
    min="0"
    {/* This is whatever the max score for the password is. */}
    max="4"
    value={passwordStength}
    {/* This is the minimum score for a bad password, in this case 0. */}
    low="0"
    {/* This is the minimum score for a "decent" password. I think for us it is 3. */}
    high="3"
    {/* This is the best value for a password. In our case this is the same as the decent one. */}
    optimum="3"
  />
  <div className='password-meter-level'></div>
  <div className='password-meter-level'></div>
  <div className='password-meter-level'></div>
</div>
```

This exposes the password meter as an actual meter, associates a label to it so it will actually be read as something that makes sense.

The 3 lines to show the password strength are shown just as squares without any extra semantics. We have to have them as styling the `<meter>` element is a nightmare.

We apply the `sr-only` class to hide the meter element so it doesn't show for all users. The divs will display as a visual equivalent.

api/src/routes/auth/index.ts
madcampos: @kbventures remove this comment please

api/src/routes/auth/index.ts
```diff
@@ -234,6 +234,129 @@ authRoutes.openapi(
...
+	}),
+	async (context) => {
+		const { email } = context.req.valid('json');
+
+		const response = { message: 'If an account with that e-mail exists, a password reset link has been sent.' };
+
+		const emailExists = await checkExistingEmail(context.env.Database, email);
+
+		// We are deliberately sending OK statuscodes to disable an attackers ability to fish for information on exisiting emails
```
madcampos:
```suggestion
		// INFO: Hide non existing emails to reduce attack surface from guessing registered emails.
```

api/src/routes/auth/index.ts
```diff
@@ -234,6 +234,129 @@ authRoutes.openapi(
...
+		if (!emailExists) {
+			// INFO: Hide non existing emails to reduce attack surface from guessing registered emails.
+			return context.json({ message: 'Internal error reseting password' } satisfies StatusResponse, StatusCodes.OKAY);
+		}
+
+		const hashedPasswordWithSalt = await hashPassword(password);
+
+		const success = await resetPassword(context.env.Database, email, hashedPasswordWithSalt);
+
+		// TODO: Send email to user stating their password has been succesfully changed(Bonus: remedy if they didn't make the change)
```
madcampos: Please resolve this TODO

madcampos (changes requested):

src/components/ResetPassword/ResetPassword.tsx
kbventures: @madcampos 

https://github.com/torontojs/communityhub/issues/210

src/index.html
kbventures: @madcampos Marco, this branch is still marked "draft" :).  I needed a dummy UUID  to test the page while building it.

src/components/ResetPassword/ResetPassword.tsx
kbventures: > **question**: why are we hiding the password meter?

That's the paradigm utilized previously in other pages and still being requested through the current design(Figma).

kbventures: https://github.com/torontojs/communityhub/issues/212

kbventures: https://github.com/torontojs/communityhub/issues/212

src/components/ResetPassword/ResetPassword.css
kbventures: https://github.com/torontojs/communityhub/issues/213

bcha92 (dismissed):
Looks good so far~

bcha92 (dismissed):
I can't think of anything else to critique aside creating unit tests for the utility function `sendPasswordResetEmail`, but I think this is something we can put in the burner for now.

src/index.html
madcampos: When testing "dummy UUIDs" please use one that is all the same character, so all 0s, all 1s, or something like that.

api/src/routes/auth/data.ts
```diff
@@ -162,3 +162,11 @@ export async function activateProfile(database: D1Database, email: string) {
 
 	return success;
 }
+
+export async function resetPassword(database: D1Database, password: string, email: string) {
+	const { success } = await database
+		.prepare(`UPDATE ${DBTables.ACCESS} SET password = ? WHERE email ? AND activatedAt IS NOT NULL AND deletedA IS NULL LIMIT 1`)
```
manvinderjit: typo:
```suggestion
		.prepare(`UPDATE ${DBTables.ACCESS} SET password = ? WHERE email = ? AND activatedAt IS NOT NULL AND deletedAt IS NULL LIMIT 1`)
```

api/src/routes/auth/index.ts
```diff
@@ -234,6 +234,165 @@ authRoutes.openapi(
...
+		const emailExists = await checkExistingEmail(context.env.Database, email);
+
+		if (!emailExists) {
+			// INFO: Hide non existing emails to reduce attack surface from guessing registered emails.
+			return context.json({ message: 'Internal error reseting password' } satisfies StatusResponse, StatusCodes.OKAY);
+		}
+
+		const hashedPasswordWithSalt = await hashPassword(password);
+
+		const success = await resetPassword(context.env.Database, email, hashedPasswordWithSalt);
```
manvinderjit: typo: the order of arguments is incorrect

image

api/src/routes/auth/index.ts
```diff
@@ -234,6 +234,165 @@ authRoutes.openapi(
...
+		await createPasswordReset({ context, email, resetToken });
+
+		await sendPasswordResetEmail(context, {
+			apiKey: context.env.RESEND_API_KEY,
+			senderEmail: context.env.SENDER_EMAIL,
+			token: resetToken,
+			email
+		});
+
+		await context.env.PasswordResetToken.put(resetToken, email);
```
manvinderjit: query: do we have a TTL for the token?

api/src/routes/documents/api-tests/get-documents/19_auth_forgot_password.hurl
```diff
@@ -0,0 +1,58 @@
+# Forgot password with valid email
+POST http://localhost:3000/api/auth/forgot-password
+Content-Type: application/json
+
+{
+  "email": "king.arthur@camelot.uk",
+  "password":"H0lyGr@il42!"
```
manvinderjit: query: why is a user sending a password when they have forgotten their password?

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,96 @@
+#password-recovery-form {
+	border: var(--border);
+	border-radius: var(--rounded-md);
+	display: flex;
+	flex-direction: column;
+	max-width: 400px;
```
manvinderjit, line 6: suggest: let us use rem for consistency
```suggestion
	max-width: 25rem;
```

src/components/ForgotPassword/ForgotPassword.css
```diff
@@ -0,0 +1,96 @@
...
+
+.input-block input {
+	background-color: var(--color-button);
+	border: var(--border);
+	border-radius: var(--rounded-xs);
+	box-sizing: border-box;
+	color: var(--color-gray-900);
+	font-size: 1rem;
+	font-weight: var(--weight-regular);
+	height: 42px;
```
manvinderjit: suggest:
```suggestion
	height: 2.625rem;
```

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,116 @@
...
+	useEffect(() => {
+		if (!isSubmitted) { return; }
+
+		if (remainingTime === 0) { return; }
+
+		const timer = setInterval(() => {
+			setRemainingTime((prev) => prev - 1);
+		}, 1000);
+
+		return () => clearTimeout(timer);
```
manvinderjit: query: why are we using `setInterval` and `clearTimeout` ? Shouldn't it be either setTimeout + clearTimeout or setInterval + clearInterval ?

src/components/ResetPassword/ResetPassword.css
```diff
@@ -0,0 +1,127 @@
...
+	width: 100%;
+	display: flex;
+	flex-direction: column;
+	gap: var(--spacing-xs);
+}
+
+.input-block label {
+	display: flex;
+	font-weight: var(--weight-semibold);
+	height: 29px;
```
manvinderjit: suggest: rem > px

src/components/TextInputComponent/TextInputComponent.css
```diff
@@ -13,18 +13,18 @@
...
+				font-size: var(--text-md);
 				font-weight: var(--weight-semibold);
 				justify-content: space-between;
 			}
 		}
 
 		& > input {
-			background-color: var(--color-card);
+			font-size: var(--text-xs);
+			background-color: var(--color-button);
```
manvinderjit: query: why are we using button color for input background?

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+	const [strength, setStrength] = useState<number | null>(null);
+	const [feedback, setFeedback] = useState<string>('');
+	const [isTokenValid, setIsValidToken] = useState<boolean | null>(null);
+
+	const token = new URLSearchParams(window.location.search).get('token');
+
+	useEffect(() => {
+		try {
+			const result = fetch('/api/auth/valid-reset-pw-token', {
```
manvinderjit: typo: missing await

manvinderjit (changes requested):

src/components/ForgotPassword/ForgotPassword.tsx
madcampos: I've commented that before... [TWO](https://github.com/torontojs/communityhub/pull/126#discussion_r2217075335) [TIMES](https://github.com/torontojs/communityhub/pull/126#discussion_r2144191119)!

madcampos: It should be set/clear interval for both.

src/components/ResetPassword/ResetPassword.tsx
madcampos: This should not be passing linting...

api/src/routes/documents/api-tests/get-documents/19_auth_forgot_password.hurl
madcampos, line 1: **issue**: this file is in the wrong folder.

It should be under:
```
api/src/auth/api-tests/forgot-password/<num>-<test-name>.hurl
```

api/src/routes/documents/api-tests/get-documents/19_auth_forgot_password.hurl
```diff
@@ -0,0 +1,58 @@
...
+}
+
+HTTP 200
+[Asserts]
+status == 200
```
madcampos:
```suggestion
```
The previous line already asserts the status code.

api/src/routes/documents/api-tests/get-documents/19_auth_forgot_password.hurl
madcampos, line 1: **issue**: also please break down different scenarios in different files. See other test files for reference.

api/src/routes/documents/api-tests/get-documents/19_auth_forgot_password.hurl
```diff
@@ -0,0 +1,58 @@
+# Forgot password with valid email
+POST http://localhost:3000/api/auth/forgot-password
+Content-Type: application/json
```
madcampos:
```suggestion
```

The content type is not required, from hurl docs: https://hurl.dev/docs/request.html#json-body

api/src/routes/documents/api-tests/get-documents/19_auth_forgot_password.hurl
madcampos: **issue**: Please use variables for this. Preferably, get a new user that is not referenced in any of the previous tests just for this case to avoid conflicts.

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault();
+
+		const formData = new FormData(event.target as HTMLFormElement);
+		const passwordValue = formData.get('password') as string;
+
+		setButtonDisabled(true);
+		setPasswordDisabled(true);
```
madcampos: **question**: why are there two state variables to disable the form> This should only be one state variable.

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+		await resetPassword(token, passwordValue);
```
madcampos, line 78: **issue**: unless there is a redirect, reenable the form after the request is sent.

src/components/ResetPassword/ResetPassword.tsx
```diff
@@ -0,0 +1,161 @@
...
+				window.location.href = '/pages/sign-in';
+			} else {
+				window.location.href = '/pages/sign-in';
+			}
+		} catch (error) {
+			if (import.meta.env.MODE === 'development') {
+				console.error(error);
+			}
+		}
+	};
```
madcampos: **issue**: If you are not using any state, move the function outside react's rendering function.

Big functions take longer to parse, execute, and process. Depending on the function size, it will bail out of compilation entirely and be interpreted making the code even slower.

On top of that, react is already slow, adding too many things inside it's rendering loop will make it even slower.

Also, use function declarations instead of function expressions whenever possible. It is in our coding style guidelines.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,116 @@
...
+			if (response.ok) {
+				setIsSubmitted(true);
+				setRemainingTime(600);
+			}
+		} catch (error) {
+			if (import.meta.env.MODE === 'development') {
+				console.error(error);
+			}
+		}
+	};
```
madcampos: **issue**: move the logic outside of the main rendering function.
This can be refactored to simplify the need of so many states.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+// TODO: Investigate if we're able to use heartbeat to check if forgot password request already in progress and how the page react if there is one
+
+const ForgotPasswordForm = (): React.JSX.Element => {
+	const emailInputRef = useRef<HTMLInputElement>(null);
+	const [isSubmitted, setIsSubmitted] = useState(false);
+	const [buttonDisabled, setButtonDisabled] = useState(true);
+	const [validEmail, setIsValid] = useState(false);
+	const [emailDisabled, setEmailDisabled] = useState(false);
+	const [remainingTime, setRemainingTime] = useState(0);
```
madcampos: **issue**: There are too many overlapping states. Please think which ones of those can be combined together or derived from other variables?

package.json
```diff
@@ -37,7 +37,7 @@
 
 		"test:coverage": "vitest run --coverage",
 		"test:unit": "vitest run",
-		"test:api-single": "hurl --continue-on-error --jobs 4 --max-time 300 --test --variables-file .dev.vars --variables-file .api-test.vars",
+		"test:api-file": "hurl --continue-on-error --jobs 4 --max-time 300 --test --variables-file .dev.vars --variables-file .api-test.vars",
 		"test:api": "npm run test:api-single -- --glob \"**/*.hurl\"",
```
madcampos:
```suggestion
		"test:api": "npm run test:api-file -- --glob \"**/*.hurl\"",
```

ProTip: when renaming things, use search and replace as much as possible, do global searches and use regex searches.

api/src/routes/auth/validation.ts
```diff
@@ -78,3 +79,26 @@ export type SignUpData = z.infer<typeof SignUpSchema>;
 export const ActivateSchema = z.object({ token: z.uuid('Invalid ID format') });
 
 export type ActivateData = z.infer<typeof ActivateSchema>;
+
+export const ForgotPasswordSchema = z.object({
+	email: z.string({ error: 'Email is required' })
+		.trim()
+		.toLowerCase()
+		.min(1)
```
madcampos: **issue**: instead of creating new schemas reference existing ones.

```suggestion
	email: ProfileSchema.shape.email
```

api/src/routes/auth/validation.ts
```diff
@@ -78,3 +79,26 @@ export type SignUpData = z.infer<typeof SignUpSchema>;
...
+	token: z.uuid()
+});
+
+export type ResetPasswordData = z.infer<typeof ResetPasswordSchema>;
+
+export const ResetPasswordValidTokenSchema = z.object({
+	token: z.uuid()
+});
+
+export type ResetPasswordValidTokenData = z.infer<typeof ResetPasswordValidTokenSchema>;
```
madcampos: **question**: Why to we have this extra schema that is similar to the previous one? Can't we reuse parts of that one?

Can't we invert the logic so there is one schema with the token and then another one that extends the one with the token to also add a password?

api/src/routes/auth/index.ts
```diff
@@ -234,6 +234,165 @@ authRoutes.openapi(
...
+			return context.json({ message: 'Failed user password change ' } satisfies StatusResponse, StatusCodes.INTERNAL_SERVER_ERROR);
+		}
+		return context.json({ message: 'Succesfully changed password' } satisfies StatusResponse, StatusCodes.OKAY);
+	}
+);
+
+authRoutes.openapi(
+	createRoute({
+		method: 'post',
+		path: '/valid-reset-pw-token',
```
madcampos:
```suggestion
		path: '/check-password-reset-token',
```

madcampos: **question**: Why do we have this endpoint?

madcampos (changes requested):
Please note:
- Some of the comments apply to global patterns, so they are about one place, and should be thought about how to apply them on the rest of the changes.
- I've avoided commenting multiple times the same thing to reduce the spam.
- There are still unresolved comments from previous reviews. Please resolve them.

src/components/ResetPassword/ResetPassword.tsx
kbventures: @madcampos 

https://github.com/torontojs/communityhub/issues/218

api/src/routes/auth/index.ts
kbventures: In the event a user tries to utilize the reset-password page with an invalid token, they are redirected.

api/src/routes/documents/api-tests/get-documents/19_auth_forgot_password.hurl
kbventures: Great question. This is @madcampos code.  :)

api/src/routes/auth/index.ts
kbventures: ????

src/components/ForgotPassword/ForgotPassword.tsx
kbventures: I went with a derived state for simplicity.

madcampos (dismissed):

madcampos (dismissed):

manvinderjit:
Why are local tests passing/skipped but remote ones are failing.

image

wrangler.toml
```diff
@@ -103,6 +113,11 @@ binding = "ActivationTokens"
 id = "987a04058cc34aad84b9f54c658f6a01"
 preview_id = "vms-production-activation-tokens"
 
+[[env.dev.kv_namespaces]]
```
manvinderjit: shouldn't this be
```suggestion
[[env.prod.kv_namespaces]]
```

src/pages/reset-password/reset-password.tsx
```diff
@@ -0,0 +1,22 @@
+import { StrictMode } from 'react';
+import { createRoot } from 'react-dom/client';
+import FooterNav from '../../components/Footer/Footer.tsx';
+import Header from '../../components/Header/Header.tsx';
+import { ResetPassword } from '../../components/ResetPassword/ResetPassword.tsx';
+import '../../index.css';
+
+const root = document.getElementById('root') as HTMLAnchorElement;
```
manvinderjit: query: Why is the `HTMLAnchorElement` and not `HTMLElement`
```
const root = document.getElementById('root') as HTMLElement;
```

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,118 @@
...
+		setIsValid(emailInputRef.current?.checkValidity() ?? false);
+	};
+
+	const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault();
+
+		const formData = new FormData(event.target as HTMLFormElement);
+		const emailValue = (formData.get('email') as string).trim().toLowerCase();
+
+		setIsSubmitted(true);
```
manvinderjit, line 57: Why is this twice. Here and again on line 64.
I think this one should be removed.

src/components/ForgotPassword/ForgotPassword.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+		const success = await requestPasswordRecovery(emailValue);
+
+		if (success) {
+			setIsSubmitted(true);
```
manvinderjit: See above comment.

manvinderjit (dismissed):
Most of this looks good. Just a couple of questions and improvements.

kbventures:
> Why are local tests passing/skipped but remote ones are failing.
> 
> image

@manvinderjit, the unit tests are not included in the CICD pipeline at the moment. When I check for failing tests, I only see a single api test failing, and it's also failing locally.

PR #145, opened 2025-06-10
Feature: export data CRON job

### LINKED ISSUE #9: Have an API/feed for querying public data
E.g.: "Marco has volunteered for event X", include a handle/email for Marco and metadata or a link for the event

**Stretch goal**: expose that as an RSS feed

madcampos:
## Summary

Adds a CRON job to run every day a data export on the application. This will upload json copy of the files to R2.

## Tests

Manually tested and data is exported correctly.

## Additional information

Resolves: https://github.com/torontojs/communityhub-be/issues/9


docs/Contribution Guide/cron-testing.md
```diff
@@ -0,0 +1,22 @@
+# Testing Scheduled CRON jobs
+
+**NOTE**: There is [no support yet](https://github.com/cloudflare/workers-sdk/issues/8466) to test this trhough vite.
```
kbventures:
```suggestion
**NOTE**: There is [no support yet](https://github.com/cloudflare/workers-sdk/issues/8466) to test this through vite.
```

eslint.config.mjs
```diff
@@ -54,7 +54,7 @@ const rules = {
 	'no-alert': 'error',
 	'no-array-constructor': 'off',
 	'no-async-promise-executor': 'error',
-	'no-await-in-loop': 'error',
+	'no-await-in-loop': 'off',
```
kbventures, line 57: Question: no objection but what is the rational here?

wrangler.toml
```diff
@@ -46,6 +51,10 @@ binding = "ActivationTokens"
 id = "vms-local-activation-token"
 preview_id = "vms-local-activation-token"
 
+[[r2_buckets]]
+binding = "ExportedFiles"
```
kbventures, line 55: No objection but I'm wondering if might be more descriptive here as ExportedFiles is very generic.

api/src/scheduler/data-export.ts
```diff
@@ -0,0 +1,111 @@
...
+			teamId: team.id,
+			members: teamMembers
+		});
+	}
+
+	return results;
+}
+
+async function getEventLog(_app: OpenAPIHono<EnvironmentBindings>, _env: Env, _context: ExecutionContext) {
+	// TODO: implement when we have endpoints for event logs
```
kbventures, line 61: Nit pick: Should we get in the habit of creating issues whenever we utilize TODO comments and add the linke below them?

eslint.config.mjs
madcampos: This is an annoying eslint rule that blocks this pattern:

```typescript
for (const item of list) {
  await somethingAsync(item);
}
```

It forces us to instead do this:

```typescript
for await (const item of list) {
  await somethingAsync(item);
}
```

But this is wrong as a `for await..of` is for async generators and not async code blocks on the loop.

The rule won't cause any issues for us.

wrangler.toml
madcampos: `ExportedPublicData` maybe?

api/src/scheduler/data-export.ts
madcampos: Yes, that is a good practice

madcampos: https://github.com/torontojs/communityhub/issues/152

kbventures (approved):

PR #146, opened 2025-06-11
Confirm account page

### LINKED ISSUE #61: "Confirm Account" page

will-frankland:
## Summary

- Created the confirm your account page.
- Modified the button to have medium size font for use
- Added code to backend to redirect user to this page when receiving confirm account email
- Added logic to extract token from URL and pass to backend, then redirect user to home

## Tests

No testing done yet. Could pair code with someone to do this, I havent written tests in a while.

## Additional information

Feedback welcome!


package.json
```diff
@@ -60,7 +55,7 @@
 		"husky": "9.1.7",
 		"typescript": "5.8.3",
 		"vite": "^6.2.2",
-		"wrangler": "^4.16.1",
+		"wrangler": "^4.19.1",
```
kbventures: Why are introducing a new version of wrangler to the project? Generally it's a project wide decision.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
...
+							Confirm my e-mail
+						</Button>
+					</div>
+					<p className='confirm-account-page-text-subtext'>
+						If you didn't request this change, please ignore this email.
+					</p>
+				</div>
+			</div>
+			<footer>
+				<p className='text-helper'>TorontoJS - 2025</p>
```
kbventures, line 78: Nit pick: I'm guess we need some font-weight increase here to m atch the Figma design.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
+import { useEffect } from 'react';
+import Button from '../Button/Button.tsx';
+import Header from '../Header/Header.tsx';
+import './ConfirmAccount.css';
+import '/index.css';
```
kbventures, line 5: This file isn't doing anything.  Please remove it.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const ConfirmAccount = () => {
+	useEffect(() => {
+		const params = new URLSearchParams(window.location.search);
+		const tokenFromUrl = params.get('token');
+		if (tokenFromUrl) {
+			console.log('Token from URL:', tokenFromUrl);
```
kbventures: We don't want to introduce console.log in our production code. 

Here's an exmample of how to do it safely if we want to maintain it in development: 

```
if (import.meta.env.MODE === 'development') {
				if (event instanceof Error) {
					console.error(event.name);
					console.error(event.cause);
					console.error(event.message);
					console.error(event.stack);
				} else {
					throw new Error('Error sign-up');
				}
```

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
...
+	}, []);
+
+	const authenticateAccount = async (token: string) => {
+		try {
+			const response = await fetch(`/api/auth/activate?token=${token}`, {
+				method: 'GET'
+			});
+			if (!response.ok) {
+				const errorData = await response.json();
+				console.error('Response not ok:', errorData);
```
kbventures: Same as previous comment.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
...
+				return;
+			}
+			window.location.href = '/pages/homepage';
+		} catch (error) {
```
kbventures, line 45: A good use of the pattern I mentioned earlier here.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
...
+			console.error('Error during authentication:', error);
+		}
+	};
+
+	// take the UUID from the Url which will be my page
```
kbventures: The code is self-explanatory here so we probably should remove these notes.

src/css-layers/design-system.css
```diff
@@ -1,4 +1,11 @@
 /* Design system */
+@font-face {
+	font-display: swap;
+	font-family: Uai;
+	font-style: normal;
+	font-weight: 100 900;
```
kbventures, line 6: Question: Are the font weights availble only 100 and 900?

src/components/ConfirmAccount/ConfirmAccount.css
```diff
@@ -0,0 +1,57 @@
+.confirm-account-page-container {
```
kbventures, line 1: @madcampos should all PX values here be converted to Rem? I'm guessing yes.

kbventures (changes requested):

vite.config.ts
```diff
@@ -44,6 +44,7 @@ export default defineConfig(({ mode }) => {
 					'review-conduct-code': fileURLToPath(new URL('./src/pages/review-conduct-code/index.html', import.meta.url)),
 					'sign-up': fileURLToPath(new URL('./src/pages/sign-up/index.html', import.meta.url)),
 					'sign-in': fileURLToPath(new URL('./src/pages/sign-in/index.html', import.meta.url)),
+					'confirm-account': fileURLToPath(new URL('./src/pages/confirm-account.html', import.meta.url)),
```
kbventures, line 47: Your path should be:
```
'confirm-account': fileURLToPath(new URL('./src/pages/confirm-account/index.html', import.meta.url)),
```

package.json
madcampos: +1 let's do package updates separately. Like, when we update packages we do this in one single PR updating everything and fixing the issues and changes the updates may introduce

madcampos: Semi related comment: once a month dependabot will open a PR with dependency updates for us: https://github.com/torontojs/communityhub/pull/87

will-frankland: Oh apologies, I think I had to do a wrangler update when i had an issue. Will sort this out!

vite.config.ts
will-frankland: Updated!

src/components/ConfirmAccount/ConfirmAccount.tsx
will-frankland: Completed!

src/components/ConfirmAccount/ConfirmAccount.tsx
will-frankland: Removed console.log

src/components/ConfirmAccount/ConfirmAccount.tsx
will-frankland: Would we want to remove the console.error altogether here?

or should I add the instance as mentioned?

src/components/ConfirmAccount/ConfirmAccount.tsx
will-frankland: Removed index.css import

src/css-layers/design-system.css
madcampos: On an `@font-face` declaration, [the two values are a range](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face#:~:text=font%2Dweight%20value.-,Accepts%20two%20values%20to%20specify%20a%20range%20that%20is%20supported%20by%20a%20font%20face,-%2C%20for%20example%20font). This is used to hint the browser about the supported range for variable fonts.

src/components/ConfirmAccount/ConfirmAccount.css
madcampos: Yes, but that is a larger refactoring task as pixel values are **_EVERYWHERE_**

package.json
```diff
@@ -7,26 +7,21 @@
...
 		"deploy:prod": "wrangler deploy --minify --env production",
-
 		"format": "dprint fmt --staged",
 		"typecheck:api": "tsc -p ./api --noEmit",
 		"typecheck:client": "tsc -p ./src --noEmit",
 		"lint:js": "eslint",
 		"lint": "npm run typecheck:api && npm run typecheck:client && npm run lint:js",
-
 		"test": "vitest",
 		"coverage": "vitest run --coverage"
```
madcampos: **nitpick:** ‏If possible, please keep the spaces between the commands. It helps with readability.

src/components/Button/Button.tsx
```diff
@@ -5,19 +5,21 @@ interface Props extends ButtonHTMLAttributes<HTMLButtonElement> {
...
 	isLarge = false,
 	children,
+	size,
 	...rest
 }, ref) => (
 	<button
 		ref={ref}
 		className='button'
-		data-size={isLarge ? 'large' : 'small'}
+		data-size={size || 'small'}
```
madcampos: **issue:** ‏The default here should be "medium" instead of "small".‏

The reasoning is: when we add one of the modifiers ("small", or "large") they are _modifiers_ so the base size should be in between them

src/components/ConfirmAccount/ConfirmAccount.css
```diff
@@ -0,0 +1,57 @@
...
+	align-items: center;
+	background-color: #e5e5e5;
+	display: flex;
+	flex-direction: column;
+	gap: var(--spacing-lg);
+	justify-content: center;
+	margin-inline: auto;
+	max-width: min(100%);
+	min-height: 100vh;
```
madcampos: **nitpick:** Better to use [`dvh`](https://developer.mozilla.org/en-US/docs/Web/CSS/length#dynamic_viewport_units)‏ here.

```suggestion
	min-height: 100dvh;
```

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
...
+		const tokenFromUrl = params.get('token');
+		if (tokenFromUrl) {
+			console.log('Token from URL:', tokenFromUrl);
+			authenticateAccount(tokenFromUrl);
+		}
+	}, []);
+
+	const authenticateAccount = async (token: string) => {
+		try {
+			const response = await fetch(`/api/auth/activate?token=${token}`, {
```
madcampos, line 37: **question:** have you tested this response when the token is malformed, invalid, or a random string?‏

It would be good to test it using different inputs to ensure the errors get caught.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+	// Change the src > email > index.ts to add /api on the URL call
+
+	// Change the email to redirect to /page/ConfirmAccount
+
+	// Go read the auth routes on the backend
+	// read dev.vars inside backend
+	// UUID Will live inside the session (superfast cache for activation tokens)
+	// Add heartbeat check, if they are alrady signed in then send them to the homepage - later
+	// Fixing btn
```
madcampos: **suggestion:** when documenting js, use [JSDoc](https://jsdoc.app/).

```suggestion
	/**
	 * Take the UUID from the Url which will be my page.
	 * The link in the email will be this page: /api/auth/activate
	 *
	 * Change the src > email > index.ts to add /api on the URL call
	 *
	 * Change the email to redirect to /page/ConfirmAccount
	 *
	 * Go read the auth routes on the backend read dev.vars inside backend UUID Will live inside the session (superfast cache for activation tokens).
	 * Add heartbeat check, if they are alrady signed in then send them to the homepage - later
	 * Fixing btn
	 */
```

Also, feel free to write longer lines, we are not in the 70s anymore where terminal monitors only support up to 80 columns of text.

worker-configuration.d.ts
madcampos, line 1: **issue:** this is a generated file, please remove it from the repository and add to `.gitignore`.

Also: this file for worker types should already be generated and ignore with another name, try running: `npm run generate:types`.

src/components/ConfirmAccount/ConfirmAccount.tsx
kbventures: @will-frankland, as a general rule, if it's likely to be useful when testing in development, then go that route.  If it won't likely be, then just remove.

src/components/ConfirmAccount/ConfirmAccount.tsx
madcampos: Adding more context: the token here can be manipulated by the use, for example: the user copies the url from the email and pastes it in the browser. But doing so, they copy only part of the token. It is malformed and we can validate it before hitting the server.

kbventures (dismissed):
Updated code for @will-frankland during his absence.

madcampos:
@kbventures please merge dev into this branch, there are a couple of conflicts

kbventures (approved):

src/components/ConfirmAccount/ConfirmAccount.css
madcampos, line 1: **issue**: please change the values from pixels to variables

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -0,0 +1,84 @@
...
+		} catch (error) {
+			if (import.meta.env.MODE === 'development') {
+				if (error instanceof Error) {
+					console.error(error.name);
+					console.error(error.cause);
+					console.error(error.message);
+					console.error(error.stack);
+				} else {
+					throw new Error('Authenticate Account Error');
+				}
```
madcampos, line 54: **issue**: please simplify the error logging

madcampos (approved):

PR #147, opened 2025-06-11
Feature: update project configuration

madcampos:
## Summary

- Update project docs (mostly move them around)
- Remove annoying `post-checkout` git hook
- Add `CODEOWNERS` file. It is used to auto-assign people as reviewers for all pull requests.
- Set CodeQL analysis config.
- Update general files
   - Security reports
   - Funding to point to Open Collective
   - Code of Conduct
   - Support file
- Add project license
- Add github actions
   - Auto assign all issues to the project
   - Apply labels to PR based on what they change
   - Close stale issues and PRs
   - Review dependencies for vulnerabilities
   - Generate a list of contributors
   - Auto assign author to PR
   - Lint action
- Fix small linting error

## Tests

No

### Project documentation

Split the contributing file into smaller sections. All of them are now markdown files.

Also added sections about testing based on recent conversation.

### Annoying git hook

The post checkout git hook was configured to install all packages and regenerate types.

This is slow and makes vscode hang. Also on Windows it may fail because some files are in use while installing new packages.

We can definitely review it to make it run faster, or keep it as well.

### Codeowners

This file tells github to auto-assign some people to be reviewers for incoming pull requests. For now the added people are: @kbventures and @madcampos. We can review the list to add more people, but be aware this will happen for _**every**_ pull request.

### CodeQL

One of the tools github provides for code analysis is CodeQL, it runs some basic checks for general code quality or weird things and for exposed secrets.

### License

Added a license file and put the code under MIT license. Please feel free to object or suggest a better license for the project. This is open for discussion and should be a consensus.

#### Contributors list

It is cool to add a list with the people who contributed to the project. This action does that. 

#### Project assignment and labels

Just some housekeeping to add non intrusive automation to the project

#### Vulnerability review

Let's test this out and see how it goes...

#### Close stale issues and PRs

Just added this to help with the automation part, the values are pretty arbitrary, so feel free to propose changes to them.


.github/.workflows/monthly-contributors.yml
```diff
@@ -0,0 +1,45 @@
...
+          echo "START_DATE=$start_date" >> "$GITHUB_ENV"
+          echo "END_DATE=$end_date" >> "$GITHUB_ENV"
+
+      - name: Run contributor action
+        uses: github/contributors@v1
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          START_DATE: ${{ env.START_DATE }}
+          END_DATE: ${{ env.END_DATE }}
+          REPOSITORY: torontojs/communityhub
```
lylaminju: What do you think about using `REPOSITORY: ${{ github.repository }}`? 
This would make it more flexible in case the repository name changes (although hopefully that won't happen).

ref) [GitHub Actions Documentation - Accessing Contextual Information](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs?utm_source=chatgpt.com#github-context)

lylaminju (commented):
It was great to learn that we can automate many tasks with github workflow 🤓

.github/.workflows/close-stale-issues.yml
kbventures, line 1: I think moving old issues  to stale is a great idea for filtering but deleting them 30 days later could lead to many great ideas we didnt have the resource or time to address to disapear needlessly.  I'm leaning towards against that specific step.  Can you explain your rational behind this specific step?

.github/.workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,24 @@
+name: "Close stale issues and PRs"
```
kbventures, line 1: I think closing issues should be something the maintainers do manually.   For example git issues might not be actioned for years due to lack of resources not because we don't want to address them.  In terms of pull request I'm leaning towards the same thought process with less intensity.  Can you share your rational for this git action?

docs/contributing-guide/testing/03-integration-testing.md
```diff
@@ -0,0 +1,26 @@
...
+> 3. Submit the information
+> 4. Check result
+>    1. Wrong information was provided, an error message is shown
+>    2. Right informatio was provided, redirect to home page
+
+## Tools
+
+For integration testing we use [`Playwright`](https://playwright.dev/).
+
+<!-- TODO -->
```
kbventures: This probably needs to  be deleted?

docs/contributing-guide/tools.md
```diff
@@ -0,0 +1,43 @@
...
+- _Recommended:_ [VS Code](https://code.visualstudio.com/Download)
+- _Optional:_ [Resend account](https://resend.com) (Used for sending emails)
+
+## `node.js` and `volta`
+
+Most of the tools used are based on `node.js`, so you should have it installed on your machine. It may be installed from an installer, or a version manager for node.
+
+Our recommendation is using `volta` for managing the node versions, it is already configured for the project, so you only need to install `volta` on your machine and it will take care of downloading the correct version of node when you first try installing the dependencies.
+
+If you already have `nvm` installed we recommend you uninstall since it may conflict with `volta`.
```
kbventures: Question.. the project now has a .nvmrc so is this still relevant?

src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -42,7 +42,7 @@ const ProfileCard = () => {
 				if (!responseProfile.ok) {
 					setIsErrorProfile(true);
 				}
-				const jsonDataProfile = await responseProfile.json();
+				const jsonDataProfile = await responseProfile.json<MemberProfile[]>();
```
kbventures: Question, did we not have problems with this some time ago? (could be wrong)

kbventures (changes requested):
I left a few minor actionable comments and questions, but otherwise almost ready to be merged.

.github/.workflows/close-stale-issues.yml
madcampos: It is more as a reminder than to actually block us.

We can configure the grace periods for both and it first will add a message to the issue/PR mentioning it is stale for a while, and then run another grace period before actually closing. If someone comments, makes changes or act upon any of those, the first grace period resets and it becomes "not stale anymore".

Furthermore, we can also add an ignore list, so some things will never be marked as stale.

I think it is valid to keep, but maybe review the grace periods...

Right now it is:
- Mark issue as stale after 120 (4 months) days of no activity at all.
- Close issue marked as stale after 30 days (1 month) of inactivity. (Totaling 5 months).
- Mark PR as stale after 15 days of no activity at all.
- Close PR marked as stale after 15 days of inactivity. (Totaling 1 month).

docs/contributing-guide/testing/03-integration-testing.md
madcampos: It was semi-intentional... I think we can describe the testing tools and process when it is more mature. Like when your PR lands for API testing.

docs/contributing-guide/tools.md
madcampos: I think sorta...

The `.nvmrc` file helps with cloudflare as it reads that to determine the node version. And it is set to use the latest LTS version, so both when building on cloudflare and if you use nvm it will always point to the latest LTS.

Volta keeps track using the `volta` field on `package.json`. It is set to a version, and not a moving target, like "LTS".

So, we can say we support both tools, but they should not be both installed in the user's system, and we should also update the versions so both match (and move the nvm version to a number instead of the moving target).

src/components/ProfileCard/ProfileCard.tsx
madcampos: This was the bug on my side, thank you for pointing out. I'll remove those changes as they are fixed in `dev`

docs/contributing-guide/tools.md
madcampos: I've reworded the docs, see what you think

docs/contributing-guide/testing/03-integration-testing.md
madcampos: Expanded the TODO to add a comment on it, see what you think

.github/.workflows/close-stale-issues.yml
kbventures: I fully support stale status and closing stale PRs, but closing "issues" will have negative consequences. 

Good examples of this could be our issues around moving to SSR and upgrading our authentication process to industry standards.  A relevant issue might take a year before it's completed.

kbventures (dismissed):

.github/.workflows/close-stale-issues.yml
madcampos: Okay, that makes sense. I've removed the auto closing for issues and added more time for PRs

manvinderjit: @madcampos Is there a way to ensure that the issues created by the core team cannot be marked stale or closed if stale while those opened by others can be?

kbventures: @manvinderjit @madcampos or a label we can add that prevents it from being deleted.

madcampos: I think this is what you are talking about: [`exempt-issue-label`](https://github.com/marketplace/actions/close-stale-issues#exempt-issue-labels)

madcampos: @manvinderjit @kbventures added that change, see what you think

kbventures (dismissed):

PR #151, opened 2025-06-14
Fix: seed data

madcampos:
## Summary

Recreate seed data.

## Tests

Manually tested inserting to the database and sending requests to the backend.

## Additional information

There is a script to generate the seed sql file from some json files.

### DIFF
--- a/api/src/routes/team-members/data.ts
+++ b/api/src/routes/team-members/data.ts
@@ -12,13 +12,13 @@ export async function addTeamMembers(database: D1Database, teamId: string, data:
 				database.prepare(`
 					INSERT INTO ${DBTables.ROLE} (
 						id, schemaVersion, happenedAt, insertedAt,
-						name, profileId, teamId,
-						descrition
+						teamId, name, description,
+						profileId
 					)
 					SELECT
 						?, ?, ?, ?,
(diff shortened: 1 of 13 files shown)


api/src/routes/team-members/data.ts
```diff
@@ -12,13 +12,13 @@ export async function addTeamMembers(database: D1Database, teamId: string, data:
 				database.prepare(`
 					INSERT INTO ${DBTables.ROLE} (
 						id, schemaVersion, happenedAt, insertedAt,
-						name, profileId, teamId,
-						descrition
+						teamId, name, description,
+						profileId
```
bcha92, line 16: Not really a blocker, but is there a particular reason `profileId` is moved to the last column?

bcha92 (dismissed):
Aside that one little non-blocker, I think it looks good!

api/src/routes/team-members/data.ts
madcampos: Yes, I tried adding team members using API calls and it was failing.

I reorganized the parameters on the SQL query to fix it as the `teamId` and `profileId` where being mixed up.

kbventures (approved):

PR #159, opened 2025-06-24
create passwordCheck utility function and unit test

### LINKED ISSUE #157: Fix API allowing weak passwords without validation-BE
Subtask for Backend work

bcha92:
Created utility function for passwordCheck


api/src/routes/auth/index.ts
```diff
@@ -29,11 +30,21 @@ authRoutes.openapi(
 			[StatusCodes.OKAY]: {
 				description: 'Created a new profile and sent an email for confirmation',
 				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.UNPROCESSABLE_CONTENT]: {
```
madcampos, line 34: **praise:** Nice status code

api/src/routes/auth/index.ts
```diff
@@ -29,11 +30,21 @@ authRoutes.openapi(
...
+				description: 'Weak Password found',
+				content: { 'application/json': { schema: StatusResponseSchema } }
 			}
 		}
 	}),
 	async (context) => {
 		const { email, password, name } = context.req.valid('json');
+
+		// Password Check Rules Here
```
madcampos: **suggestion:** I think the comment is redundant here... How about we add some do documentation to the function istelf saying what it does?

api/src/routes/auth/index.ts
```diff
@@ -29,11 +30,21 @@ authRoutes.openapi(
...
+		if (passwordCheck(password).score === 0) {
```
madcampos: **suggestion:** Let's move the test inside the function and return only a boolean. Treat the test function as a black box thato nly returns `true` or `false`.

That way the logic related to the function is all self contained.

api/src/utils/passwordCheck.test.ts
```diff
@@ -0,0 +1,29 @@
+import { describe, expect, test } from 'vitest';
+import { passwordCheck } from './passwordCheck.ts';
+
+describe('Weak password check', () => {
```
madcampos: **suggestion:** We should add those passwords to the test LOL
https://xkcd.com/936/

api/src/utils/passwordCheck.ts
```diff
@@ -0,0 +1,39 @@
+import zxcvbn from 'zxcvbn';
+
+export const passwordCheck = (password: string) => {
```
madcampos:
```suggestion
/**
 * Checks if the given password is strong enough.
 * Returns `true` if the password passes the strength test.
 */
export function passwordStrengthCheck(password: string) {
```

Couple things here:
- A nitpick: please prefer function declarations instead of arrow functions. Specially if the functions are named.
- The JSDoc comment (comment starting with `/**`) adds documentation to the function and the ts language server will provide nice hints to it.
- This function should have the logic self contained, as explained in another comment.

About JSDoc:
Here is the main documentation: https://jsdoc.app/
Here is what typescript supports: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html
And it supports inline markdown in the comments.

api/src/utils/passwordCheck.ts
```diff
@@ -0,0 +1,39 @@
...
+	const passwordHasUppercase = /[A-Z]/u.test(password);
+	const passwordHasLowercase = /[a-z]/u.test(password);
+	const passwordHasNumber = /[0-9]/u.test(password);
+	const passwordHasSymbol = /[!@#$%^&*/?~]/u.test(password);
+	const noAmbiguousChar = !/[^a-zA-Z0-9!@#$%^&*/?~]/u.test(password);
```
madcampos: **question:** I think those cases are already covered by the library, no?

madcampos: **question:** Can you explain what "ambiguous characters" mean here? Like, what cases are you covering?

api/src/routes/auth/index.ts
kbventures: Clean code principle.  Good code is self-explanatory.

api/src/utils/passwordCheck.ts
kbventures: The library allows passwords with no numbers, has no combination of lower or upper case characters, and no special symbols.

This, for example, would be perceived as strong aaaaabbbbbc by the library.

I think having these additional tests for upper/lower case, number, and symbol are a good idea.

kbventures (changes requested):
UPDATED: 

After the library specification has been clarified and the function is updated. 

We need api tests update for the auth hurl file.

We can create a separate issue you can work on after this one is approved, or you can take what you need from:

https://github.com/torontojs/communityhub/pull/117

api/src/utils/passwordCheck.ts
madcampos: I'll use my powers of pedantry [^1] here ans say: no.

I will be very annoying and cite the [most recent recommendations for password from NIST here](https://pages.nist.gov/800-63-4/sp800-63b.html). Those are the standards and guidance followed by governments and the people who make things like the crypto algorithms everyone uses.

The recommendations say on section [3.1.1.2. Password Verifiers](https://pages.nist.gov/800-63-4/sp800-63b.html):
> The following requirements apply to passwords:
> 1. Verifiers and CSPs SHALL require passwords to be a minimum of eight characters in length and SHOULD require passwords to be a **minimum of 15 characters in length**.
> 2. Verifiers and CSPs SHOULD permit a **maximum password length of at least 64 characters**.
> 3. Verifiers and CSPs SHOULD **accept all printing ASCII [[RFC20]](https://pages.nist.gov/800-63-4/sp800-63b.html#ref-RFC20) characters and the space character in passwords**.
> 4. Verifiers and CSPs **SHOULD accept Unicode [[ISO/ISC 10646]](https://pages.nist.gov/800-63-4/sp800-63b.html#ref-ISOIEC10646) characters in passwords**. Each Unicode code point SHALL be counted as a single character when evaluating password length.
> 5. Verifiers and CSPs **SHALL NOT impose other composition rules (e.g., requiring mixtures of different character types) for passwords**.
> 6. Verifiers and CSPs SHALL NOT require users to change passwords periodically. However, verifiers SHALL force a change if there is evidence of compromise of the authenticator.
> 7. Verifiers and CSPs SHALL NOT permit the subscriber to store a hint that is accessible to an unauthenticated claimant.
> 8. Verifiers and CSPs SHALL NOT prompt subscribers to use knowledge-based authentication (KBA) (e.g., “What was the name of your first pet?”) or security questions when choosing passwords.
> 9. Verifiers SHALL verify the entire submitted password (i.e., not truncate it).

All the bolded parts are mine. It basically means that:
- A password should have at least 15 characters of length.
- It should be at most 64 characters, more is acceptable.
- All printable characters, in all Unicode, are valid [^2].
- We should not enforce bogus rules like require uppercase, lowercase, numbers, special characters, or other arbitrary rules.

The TLDR: for passwords _size matters_, larger passwords are better. The contents of said password have no change in how secure it is.

[^1]: Not to be confused with the powers of _pederasty_, those are my other powers...
[^2]: This is easily regex-able with this regex: `/\p{Cc}/iug`

kbventures: About the library...

@madcamposthe The zxcvbn library on the frontend considers a password like aaaabbbbbc to be "strong." and it has 11 characters. 

Currently, the library evaluates strength based on multiple factors—like character variety, casing, symbols,  length, and pattern, which you rightly pointed out as many recognized standard flaws. For example, a 9-character password like !@#$aaaa0 might be rated as "strong," while a longer one like aaaabbbbc could also get the same rating.

So, should we enforce a minimum of 15 characters? That requirement doesn’t align with how the library evaluates password strength.

The library seems to go against many industry standards, but building our own library/checks seems impractical.  

We seem to have a misalignment we what we want and what the library does.

madcampos: Yes, we should enforce a minimum length and have a recomended length. I would say anything below 10~12 characters is not very strong, and 15~20 characters is a good enough size.

As for character variety, _technically_ it is okay to have a password like "`AAAAAAAAAAAAAAA`". It will be as hard to break with a brute force attack as "`0#4${alGBxi.?"`".

What makes the second password _better_ is it's **_randomness_**, which makes it more resilient agains [dictionary attacks](https://en.wikipedia.org/wiki/Dictionary_attack).

I'm down for using [another library](https://www.npmjs.com/search?q=password%20strength) for password validation. All of them will be kinda flawed though as it would require an actual dictionary to compare against and that is well, _heavy_ to embed in a script.

An option would be to send requests to the [Have I Been Pwned API](https://haveibeenpwned.com/API/v3). That checks for passwords that show up in known breaches and tests against a dictionary, which helps a lot to provide a more secure password. That with a library that does some clever analysis of frequency and repetition will mitigate most issues.

The only thing left to do then is add a rate limit to the API endpoints so an attacker won't be able to spam the server with requests to try and figure something out.

Our passwords are already hashed and salted, which solves the other issue of storage, so even if the database leaks, it would take an infeasible ammount of time to break the hashing and reverse the saved passwords. And it protects against [rainbow table attacks](https://en.wikipedia.org/wiki/Rainbow_table).

kbventures: As for rate limiting rule we have one free one: 

https://developers.cloudflare.com/waf/rate-limiting-rules/

madcampos: Yeah, definitely use Cloudflare's rate limiter

bcha92: It was something I picked up from when doing some open source contribution for BitWarden (which I happen to use as a password manager myself)...

But I'm also a bit perplexed what should be "considered" an ambiguous character like mistaking `l` (lowercase L) for `1` (number one) or `I` (uppercase i), or `0` (number zero) for `O` (uppercase o), among others. It was used for "random" password generator function in their repo.

Source: https://community.bitwarden.com/t/list-of-characters-for-avoid-ambiguous-characters/13891

Again, it is merely a suggestion, which I can remove if it's pointless to over-engineer this sort of thing.

api/src/routes/auth/index.ts
bcha92: removed comment

api/src/routes/auth/index.ts
bcha92: should the password check be before the email check, or afterward? But I'm certain it should be before the hashing/salting.

api/src/utils/passwordCheck.test.ts
bcha92: will add more tests

api/src/utils/passwordCheck.ts
bcha92: changed to function declaration and naming to strengthCheck, should return a boolean value

api/src/utils/passwordCheck.ts
bcha92: removed as it may not be necessary for our use case

api/src/routes/auth/index.ts
bcha92: changed `passwordStrengthCheck(password: string)` to return `true` or `false`

api/src/utils/passwordCheck.test.ts
bcha92: added more tests

api/src/utils/passwordCheck.ts
bcha92: Currently pared the function down to check for the essentials:
1. password strength score (via library) is at least strong (3 or higher)
2. min character of password is 15
3. should not use non-unicode characters (couldn't find any on google, but found non-UTF-8 characers which I know is completely different and passes the checks)

madcampos: The "non-unicode" part are the [C0 and C1 control characters](https://en.wikipedia.org/wiki/C0_and_C1_control_codes), like the `null` character.

To create a string with those in javascript you can do this:
```js
const controlStringNullChar = '\x00';
```

Basically escaping them as: `\x<two digit hex>`

madcampos: This is way more complex... Like, it is both character and font dependant.

For example: ["В"](https://en.wikipedia.org/wiki/Ve_(Cyrillic)) (Cyrillic letter "Ve") is usually rendered very similar to "B" (Uppercase "b").

There is a huge list of characters that are like this, and with unicode updating every year, this is also a moving target 🥲

api/src/utils/passwordStrengthCheck.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function passwordStrengthCheck(password: string): boolean {
+	const result = zxcvbn(password);
+	const guessable = result.score < 3;
+	const minLength = password.length < 15;
+	const nonUnicodeCharFound = password.match(/\p{Cc}/iug);
+
+	if (guessable || minLength || nonUnicodeCharFound) {
+		return false;
+	}
```
madcampos:
```suggestion
	const hasControlCharacter = /\p{Cc}/iug.test(password);

	if (guessable || minLength || hasControlCharacter) {
		return false;
	}
```

Two nitpicks:
- Prefer `RegExp.test` for boolean checks as it just checks if the regex matches or not and don't create the extra things that `string.match` creates... Confusing, I know...
- Rename the variable to better represent it is a boolean.

api/src/utils/passwordStrengthCheck.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+	test('"m8G!7zdPlAy4o9b$vT9f" should pass', () => {
+		expect(passwordStrengthCheck('m8G!7zdPlAy4o9b$vT9f')).toEqual(true);
+	});
+
+	test('"Tr0ub13M4k3R1ntH3h0u53" should pass', () => {
+		expect(passwordStrengthCheck('Tr0ub13M4k3R1ntH3h0u53')).toEqual(true);
+	});
+
+	test('"あڥA3+]?áàâãäçéèêëéèêëñóòôõöúùûüÁÀÂÃÄÇÉÈÊËÍÌÎÏÑÓÒÔÕÖÚÙÛÜßẞ" should not pass', () => {
```
madcampos:
```suggestion
	test('"あڥA3+]?áàâãäçéèêëéèêëñóòôõöúùûüÁÀÂÃÄÇÉÈÊËÍÌÎÏÑÓÒÔÕÖÚÙÛÜßẞ" should pass', () => {
```

api/src/utils/passwordStrengthCheck.test.ts
```diff
@@ -0,0 +1,37 @@
...
+		expect(passwordStrengthCheck('あڥA3+]?áàâãäçéèêëéèêëñóòôõöúùûüÁÀÂÃÄÇÉÈÊËÍÌÎÏÑÓÒÔÕÖÚÙÛÜßẞ')).toEqual(true);
+		// NOTE: Non UTF-8 Character Example Here, but since we are only checking for non-unicode characetr, this string passes the test
```
madcampos:
```suggestion
		test('C0 control characters should not pass', () => {
			 const controlCharacters = '\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0a\x0b\x0c\x0d\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f\x7f';
			expect(passwordStrengthCheck(controlCharacters)).toEqual(false);
		}
```

madcampos (changes requested):

api/src/utils/passwordStrengthCheck.test.ts
bcha92: oopsie 😳

api/src/utils/passwordStrengthCheck.test.ts
bcha92: I'll have to look up the C0 control characters in more detail.

madcampos: @bcha92 I think I linked to a [wikipedia article](https://en.wikipedia.org/wiki/C0_and_C1_control_codes) about them in one of the comments. Basically, they are old stuff from then computers looked more like typewriters than black glass slabs.

I think the most dangerous one is `\0`, it is literally the `null` character. Strings in C for most applications are "null terminated". That means a null character is considered the end of a string, but if the length doesn't match the content it can lead to some memory leaks and other interesting attacks and hacks...

Another cool one is the [bell character](https://en.wikipedia.org/wiki/Bell_character). Typing it on a terminal makes it produce an audible beep 🤣

madcampos (dismissed):

kbventures (approved):

kbventures:
@bcha92 great work.  It's ready, but you need to merge dev into it first(some conflicts) and add / fix API tests.

kbventures (changes requested):
You just need to merge dev into your branch and update / add API tests. Great work!

api/api_tests/01_auth_sign-up.hurl
```diff
@@ -12,7 +12,7 @@ Content-Type: application/json
 HTTP 200
```
kbventures: error: Assert status code
   --> api/api_tests/01_auth_sign-up.hurl:26:6
    |
    | POST {{host}}/api/auth/sign-up
    | ...
 26 | HTTP 422
    |      ^^^ actual value is <200>
    |

error: Assert status code
   --> api/api_tests/01_auth_sign-up.hurl:41:6
    |
    | POST {{host}}/api/auth/sign-up
    | ...
 41 | HTTP 422
    |      ^^^ actual value is <200>
    |

```
		if (emailExists) {
			// INFO: Hide non existing emails to reduce attack surface from guessing registered emails.
			return context.json(response satisfies StatusResponse, StatusCodes.OKAY);
		}
```

```
# Sign-up duplicate email with different casing UNPROCESSABLE CONTENT
# FIXME: https://github.com/torontojs/communityhub/issues/97
```

For the first one, it's supposed to return 200 for security reasons to obfuscate the real info.  As for the second,d there's a GitHub issue that needs to stop someone from opening an account with the same email utilizing different casing for email.

api/src/utils/passwordStrengthCheck.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Checks if the given password is strong enough
+ * Returns `true` if the password passes the strength test
+ */
+
+export function passwordStrengthCheck(password: string): boolean {
+	const result = zxcvbn(password);
+	const guessable = result.score < 3;
+	const minLength = password.length < 15;
+	const hasControlCharacter = /\p{Cc}/iug.test(password);
```
kbventures, line 18: Question: Why are we checking for control characters? .bind() treats them as raw strings.

kbventures (dismissed):
Great work.  I left some notes to action/respond to. 

Nitpic: Is there no room for new tests here?

api/src/utils/passwordStrengthCheck.ts
bcha92: @kbventures non-unicode characters

api/api_tests/01_auth_sign-up.hurl
bcha92: Didn't know about the first one using 200 for security reasons, but if we need something "neutral" so the request is received but not an OKAY to confirm an email exists or not, shouldn't we use either 202 (ACCEPTED) or 204 (NO CONTENT) for this instead?

bcha92: Should add new tests but trying to find which one to add (or duplicate the vitest ones?).

api/src/utils/passwordStrengthCheck.ts
kbventures: @bcha92 Please expand on this rationale.

api/api_tests/01_auth_sign-up.hurl
kbventures: @bcha92 We are obfuscating to not allow fishing for information.   

As for 202 vs 204.  Both are fine to me. Feel free to apply a solution or create an issue to investigate / discuss it further. 

As for the tests,  focus on hurl tests when possible for routes and vitest for utilities.  We're slowing down on unit tests until a Cloudflare vitest dependency problem is resolved.

api/src/utils/passwordStrengthCheck.ts
bcha92: @kbventures I'm open to another idea if you have suggestions.

TLDR: Passwords should only be limited to ASCII characters.

The initial idea here is while it is possible to use non-ASCII characters before hashing a password, we should not be using them at all: for example cyrillic, greek, indo-dravidian (Hindi, Bengali, etc.), japanese because such characters would be represented differently in each unicode setting and servers without explicit encoding setting would naturally default to ASCII, which may cause incorrect conversion of passwords on the other side. In such cases, sticking with traditional ASCII is the best bet, especially for legacy UIs and infrastructures without explicit unicode settings.

kbventures: @bcha92 I'm not 100% sure.  

https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html

The above is on my todo list. :)

kbventures: Summoning a wizard @madcampos

bcha92: @kbventures naturally the FE `<input>` element already enforces ASCII-string-only passwords by default, but server-side enforcement should be used as a precaution given SQL injection might be the least of our problems (see @madcampos for that "thing" they discovered during testing [can't remember which PR now lol]).

kbventures: I might be hanging myself here, but for me, this is a suspicious check. 

image

D1 is using prepared statements, and their main use case, aside from making reusable queries, is preventing SQL injections. 

Before we even get to the prepared statements,  we have strict query and json checks https://hono.dev/docs/api/request#valid ,and then at the data level, we have zod schemas so any incompatible character set or some binary shenanigans won't fly here either. 

I didn't know input enforced ASCII characters so that's a very interesting point.

api/api_tests/01_auth_sign-up.hurl
madcampos: > Didn't know about the first one using 200 for security reasons, but if we need something "neutral" so the request is received but not an OKAY to confirm an email exists or not, shouldn't we use either 202 (ACCEPTED) or 204 (NO CONTENT) for this instead?

Any of those work here... 🤷‍♂️

The important part is to return the same response regardless of result, if the account exists or not. To obfuscate potential attacks we should always return the same thing.

As far as semantic precision, the "best" status code would be actually `201` (Created). As it means a new resource was created.

`202` Accepted is for asynchronous things, like "I got it, put on a queue, will deal with later".
`204` No content is for when the headers are important but there is actually no content. We do return a status message here, so there is actually content.

api/src/utils/passwordStrengthCheck.ts
madcampos: > @kbventures I'm open to another idea if you have suggestions.
> 
> TLDR: Passwords should only be limited to ASCII characters.
> 
> The initial idea here is while it is possible to use non-ASCII characters before hashing a password, we should not be using them at all: for example cyrillic, greek, indo-dravidian (Hindi, Bengali, etc.), japanese because such characters would be represented differently in each unicode setting and servers without explicit encoding setting would naturally default to ASCII, which may cause incorrect conversion of passwords on the other side. In such cases, sticking with traditional ASCII is the best bet, especially for legacy UIs and infrastructures without explicit unicode settings.

You are semi-correct here. Yes, there may be diferences in how unicode creates some "characters", but no, we should not block unicode characters at all.

## Blocking characters

If we block a character we reduce the number accepted characters, thus making it easier for a brute force attack to succeed. Less characters, less combinations to check.

We should definitely allow users to type whatever, it is not an issue at all, and working with a high level language like js is totally fine.

## Languages and support for unicode

In the past, we really would have to worry about character encoding, but nowadays, working with a high level language such as javascript we don't need to worry.

JS, SQLite, and browsers all understand and use unicode by default or with minimal configuration.

## Character representation

So, characters in unicode are variable length, that means you need a minimum of 1 byte (8 bits) to represent a character, but some characters can be represented by more than 1 byte. E.g.: "💩" is two bytes long. The amount of bytes is variable and some emojis can have some wild number like 16 bytes. I gave a [talk about unicode](https://madcampos.dev/talks/tojs-unicode/) a while back, check it out for more details.

## Unicode normalization

The problem with representation is for some characters that can be represented in multiple ways. Here I'm not talking about homographs, characters that _look_  the same. I'm talking about characters that are _literally_ the same.

One example of that is the character "ñ". It can be represented two ways: "~" + "n" (a tilde plus an "n") or "ñ" (an "n" with a tilde).

This is part of the unicode spec and how it works is complex for a number of reasons that are not worth getting into the details.

So, to solve this problem we have a string method: [`String.normalize`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize).
It will get a string and normalize it, so if we pick a normalization that combines characters this method will do some unicode sorcery and transform the string with the split characters into the string with combined characters.

As long as we use the same normalization throughout, it doesn't matter the user input, we will always get the same result.

## Blocking characters 2: electric boogaloo

Now, there is one class of characters that needs care and blocking: C0 and C1 control characters. I linked to the wikipedia article on them about 4 times now, so I won't again.

Those characters are part of ASCII, are also part of unicode, and they are problematic because they include things like the null byte (`\0`).

Those we should block, and answering to @kbventures' question: this is what this regex blocks, C0 and C1 characters.

## Password inputs

No, password inputs don't block non ASCII characters:
screenshot of a password input and the console value of it with the string 'ñ'.

To wrap up: that is like the 4th time I'm typing the same things, please don't make me repeat the same thing about passwords and unicode for a 5th time.

madcampos: > .bind() treats them as raw strings

@kbventures you mean [`db.bind()`](https://developers.cloudflare.com/d1/worker-api/prepared-statements/#bind) or [`Function.bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind)

Those are two completely different things.

madcampos (changes requested):
Ditto to Ken's comment about the hurl tests.
Please run the tests using curl, postman, or whatever tool to check if they are giving a `422` status code due to the weak password. If so, we need to change the passwords on the tests to make them stronger.

madcampos (dismissed):

api/src/utils/passwordStrengthCheck.ts
```diff
@@ -0,0 +1,19 @@
+import zxcvbn from 'zxcvbn';
+
+/**
+ * Checks if the given password is strong enough
+ * Returns `true` if the password passes the strength test
+ */
+
```
kbventures, line 13: Maybe we shoulld add a note here as to which control character we are talking about there @bcha92 

```
The regex /\p{Cc}/iug checks for any Unicode control characters (category Cc), including characters like:

\x00–\x1F (e.g., NULL, BEL, TAB, LF, CR)

\x7F (DEL)

and other non-printable Unicode control codes beyond ASCII.

These are generally disallowed in passwords for security and usability.
```

kbventures (dismissed):
Approve. Great work.  Please add a note in regards to the control character check.  Surely someone else will wonder what they are again in the future! :)

madcampos (approved):

PR #163, opened 2025-07-02
feat: add Design System Text Input Component

### LINKED ISSUE #162: [Design System] Text Input Component
## Description

We have HTML input text tags being used multiple times in and across the pages. Creating a design component will allow re-usability.

## Proposed solution

Create an Input Text JSX Component as a part of the system design for re-usability.

## Additional context

Image

manvinderjit:
## Summary

Add a design system component for text inputs and labels

## Tests

Pending

## Additional information

Based on the following figma design

Screenshot from 2025-07-01 17-58-03

closes #162

### DIFF
--- a/src/components/TextInputComponent/TextInputComponent.css
+++ b/src/components/TextInputComponent/TextInputComponent.css
@@ -15,8 +15,6 @@
 			font-size: var(--text-sm);
 			font-weight: var(--weight-semibold);
 			justify-content: space-between;
-
-			&[aria-disabled='true'] { color: var(--color-helper); }
 		}
 	}
 
@@ -47,3 +45,5 @@
 		text-transform: uppercase;
 	}
 }
+
+.text-input-component-container:has(input[aria-disabled]) label { color: var(--color-helper); }
--- a/src/components/TextInputComponent/TextInputComponent.css
+++ b/src/components/TextInputComponent/TextInputComponent.css
@@ -1,0 +1,49 @@
+.text-input-component-container {
+	display: flex;
+	flex-direction: column;
+	gap: var(--spacing-xs);
+
+	.input-label-container {
+		align-items: center;
+		display: flex;
+		justify-content: space-between;
+
+		& > label {
+			color: var(--color-foreground);
+			display: flex;
+			flex-grow: 1;
+			font-size: var(--text-sm);
+			font-weight: var(--weight-semibold);
+			justify-content: space-between;
+
+			&[aria-disabled='true'] { color: var(--color-helper); }
+		}
+	}
+
+	& > input {
+		background-color: var(--color-card);
+		border: var(--border);
+		border-radius: var(--rounded-xs);
+		gap: var(--spacing-xs);
+		height: auto;
+		padding: var(--spacing-xs);
+		width: 100%;
+
+		&[data-error='true'] { border-color: var(--color-error); }
+
+		&[aria-disabled='true'] {
+			background-color: var(--color-background);
+			cursor: not-allowed;
+			pointer-events: none;
+		}
+	}
+
+	&:has(input:required) > span.input-label-container > label::after {
+		color: var(--color-accent);
+		content: 'Required' / '';
+		font-size: 0.75rem;
+		font-weight: var(--weight-semibold);
+		letter-spacing: 0;
+		text-transform: uppercase;
+	}
+}
--- a/src/components/TextInputComponent/TextInputComponent.css
+++ b/src/components/TextInputComponent/TextInputComponent.css
@@ -44,6 +44,5 @@
 		letter-spacing: 0;
 		text-transform: uppercase;
 	}
+	&:has(input[aria-disabled]) label { color: var(--color-helper); }
 }
-
-.text-input-component-container:has(input[aria-disabled]) label { color: var(--color-helper); }
--- a/src/components/TextInputComponent/TextInputComponent.css
+++ b/src/components/TextInputComponent/TextInputComponent.css
@@ -1,51 +1,51 @@
 @layer components {
-  .text-input-component-container {
-	  display: flex;
-	  flex-direction: column;
-	  gap: var(--spacing-xs);
-  
-	  .input-label-container {
-		  align-items: center;
-		  display: flex;
-		  justify-content: space-between;
-  
-		  & > label {
-			  color: var(--color-foreground);
-			  display: flex;
-			  flex-grow: 1;
-			  font-size: var(--text-sm);
-			  font-weight: var(--weight-semibold);
-			  justify-content: space-between;
-		  }
-	  }
-  
-	  & > input {
-		  background-color: var(--color-card);
-		  border: var(--border);
-		  border-radius: var(--rounded-xs);
-		  gap: var(--spacing-xs);
-		  height: auto;
-		  padding: var(--spacing-xs);
-		  width: 100%;
-  
-		  &[data-error='true'] { border-color: var(--color-error); }
-  
-		  &[aria-disabled='true'] {
-			  background-color: var(--color-background);
-			  cursor: not-allowed;
-			  pointer-events: none;
-		  }
-	  }
-  
-	  &:has(input:required) > span.input-label-container > label::after {
-		  color: var(--color-accent);
-		  content: 'Required' / '';
-		  font-size: 0.75rem;
-		  font-weight: var(--weight-semibold);
-		  letter-spacing: 0;
-		  text-transform: uppercase;
-	  }
-  }
-  
-  .text-input-component-container:has(input[aria-disabled]) label { color: var(--color-helper); }
+	.text-input-component-container {
+		display: flex;
+		flex-direction: column;
+		gap: var(--spacing-xs);
+
+		.input-label-container {
+			align-items: center;
+			display: flex;
+			justify-content: space-between;
+
+			& > label {
+				color: var(--color-foreground);
+				display: flex;
+				flex-grow: 1;
+				font-size: var(--text-sm);
+				font-weight: var(--weight-semibold);
+				justify-content: space-between;
+			}
+		}
+
+		& > input {
+			background-color: var(--color-card);
+			border: var(--border);
+			border-radius: var(--rounded-xs);
+			gap: var(--spacing-xs);
+			height: auto;
+			padding: var(--spacing-xs);
+			width: 100%;
+
+			&[data-error='true'] { border-color: var(--color-error); }
+
+			&[aria-disabled='true'] {
+				background-color: var(--color-background);
+				cursor: not-allowed;
+				pointer-events: none;
+			}
+		}
+
+		&:has(input:required) > span.input-label-container > label::after {
+			color: var(--color-accent);
+			content: 'Required' / '';
+			font-size: 0.75rem;
+			font-weight: var(--weight-semibold);
+			letter-spacing: 0;
+			text-transform: uppercase;
+		}
+	}
+
+	.text-input-component-container:has(input[aria-disabled]) label { color: var(--color-helper); }
 }
--- a/src/components/TextInputComponent/TextInputComponent.css
+++ b/src/components/TextInputComponent/TextInputComponent.css
@@ -44,5 +44,6 @@
 		letter-spacing: 0;
 		text-transform: uppercase;
 	}
-	&:has(input[aria-disabled]) label { color: var(--color-helper); }
 }
+
+.text-input-component-container:has(input[aria-disabled]) label { color: var(--color-helper); }
--- a/src/components/TextInputComponent/TextInputComponent.css
+++ b/src/components/TextInputComponent/TextInputComponent.css
@@ -1,49 +1,51 @@
-.text-input-component-container {
-	display: flex;
-	flex-direction: column;
-	gap: var(--spacing-xs);
-
-	.input-label-container {
-		align-items: center;
-		display: flex;
-		justify-content: space-between;
-
-		& > label {
-			color: var(--color-foreground);
-			display: flex;
-			flex-grow: 1;
-			font-size: var(--text-sm);
-			font-weight: var(--weight-semibold);
-			justify-content: space-between;
-		}
-	}
-
-	& > input {
-		background-color: var(--color-card);
-		border: var(--border);
-		border-radius: var(--rounded-xs);
-		gap: var(--spacing-xs);
-		height: auto;
-		padding: var(--spacing-xs);
-		width: 100%;
-
-		&[data-error='true'] { border-color: var(--color-error); }
-
-		&[aria-disabled='true'] {
-			background-color: var(--color-background);
-			cursor: not-allowed;
-			pointer-events: none;
-		}
-	}
-
-	&:has(input:required) > span.input-label-container > label::after {
-		color: var(--color-accent);
-		content: 'Required' / '';
-		font-size: 0.75rem;
-		font-weight: var(--weight-semibold);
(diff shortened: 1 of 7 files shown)


src/components/TextInputComponent/TextInputComponent.tsx
```diff
@@ -0,0 +1,67 @@
...
+				defaultValue={value || ''}
+				data-error={error ? true : false}
+				data-helper={helper ? true : false}
+				placeholder={placeholder}
+				readOnly={isReadOnly}
+				{...rest}
+			/>
+		</div>
+	);
+});
```
madcampos:
```suggestion
interface Props extends InputHTMLAttributes<HTMLInputElement> {
	label: string;
	error?: string;
	helper?: string | null;
	labelSlot?: React.JSX.Element;
}

const TextInputComponent = forwardRef<HTMLInputElement, Props>(({
	id,
	name,
	label,
	disabled,
	error,
	helper,
	value,
	labelSlot,
	onKeyDown,
	...rest
}, ref) => {
	// Conditional flag checks
	if (!label) {
		console.warn('Custom Warning: input must always have a label. Otherwise it fails the WCAG SC 3.3.2 standard.');
	}
	
	const elementId = useMemo(() => id ?? `input-${Math.trunc(Math.random() * 10000).toString(16)}`, [id]);

	const keydownHandler: KeyboardEventHandler<HTMLInputElement> = (evt) => {
		const allowedKeys = [
			'Tab',
			'Home',
			'PageUp',
			'PageDown',
			'End',
			'Enter',
		];

		// Disables most keypresses on disabled input.
		if (disabled && !allowedKeys.includes(evt.key)) {
			evt.preventDefault();
			return;
		}

		// Execute the provided event handler, if it exists.
		onKeyDown?.(evt);
	};

	return (
		<div className='text-input-component-container'>
			<span className='input-label-container'>
				<label htmlFor={elementId}>{label}</label>
				{labelSlot}
			</span>
			<input
				ref={ref}
				id={elementId}
				name={name}
				type='text'
				aria-disabled={disabled}
				defaultValue={value || ''}
				data-error={error ? true : false}
				data-helper={helper ? true : false}
				{...rest}
			/>
		</div>
	);
});
```

Notes here:
- Most of the props should already be covered by the `extends` there, as they use the same names and types as the ones already on an input element, so adding them is just duplicating existing code.
- `id` is optional, but for connecting an input with it's label it is required. Here the id will be used if existing, or else a random id will be created.
- Applying the `aria-disabled` attribute on a label has no effect.
- A read only input with no value is not an issue.
- I think a required input with something on the label slot should not be an issue as well.
- The key down event should be handled here.

src/pages/text-input-usage/sample-text-inputs.tsx
```diff
@@ -0,0 +1,150 @@
...
+		'Control',
+		'Alt',
+		'Meta'
+	];
+
+	// Block typing and other keys
+	if (!allowedKeys.includes(e.key)) {
+		e.preventDefault();
+	}
+};
```
madcampos: This is better handled from inside the component.

src/pages/text-input-usage/sample-text-inputs.tsx
```diff
@@ -0,0 +1,150 @@
...
+			<div className='App'>
+				{/* NOTE: This is temp UI to test input component design */}
+
+				<h2 style={{ textAlign: 'center' }}>Render all the Design System's Text Input Component Cases</h2>
+				<h2 style={{ textAlign: 'center' }}>
+					<span style={{ color: 'hsla(2, 84%, 56%, 1)' }}>NOTE:</span> The input must always have a label to meet the WCAG SC 3.3.2 standard.
+				</h2>
+
+				<div style={{ display: 'flex', flexDirection: 'column', gap: '1rem', padding: '1rem', width: '400px', border: '2px dashed', borderRadius: '5px', margin: 'auto' }}>
+					<h5 style={{ padding: '0', margin: '20px 0 0 0', fontSize: '1.1rem', fontStyle: 'italic', fontWeight: '500', borderBottom: '1px grey dashed' }}>
```
madcampos: Please don't use an `h5` tag without using all of the other headers before it.

Here is it better to use multiple `fieldset`s

madcampos (changes requested):

src/components/TextInputComponent/TextInputComponent.tsx
manvinderjit: The aria-disabled was for changing the text color of the label. I didn't want to create a new data attribute, so used the same one.

src/pages/text-input-usage/sample-text-inputs.tsx
manvinderjit: Initially I did create a fallback method inside the component, but then i moved it outside in case we want to give the user the option to control the disable behaviour.

src/pages/text-input-usage/sample-text-inputs.tsx
manvinderjit: This is a non-production page just to show how to use the inputs. It will not be pushed to production.

src/components/TextInputComponent/TextInputComponent.tsx
madcampos: You can use a `:has` selector instead for the label, like this:
```css
.input-wrapper:has(input[aria-disabled]) label {}
```

src/pages/text-input-usage/sample-text-inputs.tsx
madcampos: It is better to keep inside the component, as this behavior is expected to be a default, and anything other than that is the exception. We can treat the exception when it arises

src/pages/text-input-usage/sample-text-inputs.tsx
madcampos: Still, it helps to build the habit of writing accessible code, so it becomes second nature ;)

manvinderjit: Okie dokie!

src/components/TextInputComponent/TextInputComponent.tsx
manvinderjit: @madcampos I was not using the default disabled because using it applies the default disabled styles and makes the button disabled. If we want to leverage `aria-disabled`, we need to have a separate prop. 
Additionally, it will give us the option of leveraging both disabled and aria-disabled separately.

madcampos: That is fine, let's use `disabled` on the component as most of the time the intended outcome is to use `aria-disabled` on the input.

src/components/TextInputComponent/TextInputComponent.tsx
```diff
@@ -0,0 +1,72 @@
+import { forwardRef, type InputHTMLAttributes, type KeyboardEventHandler, useMemo } from 'react';
```
kbventures, line 1: forwardRef is deprecated. Please "ref"

https://react.dev/reference/react/forwardRef

manvinderjit, line 1: @kbventures forwardRef is deprecated version 19+ onward. We are on 18.xx. Using ref on our current version throws errors.

src/pages/text-input-usage/sample-text-inputs.tsx
```diff
@@ -0,0 +1,128 @@
...
+	return (
+		<button
+			onClick={onClick}
+			aria-label={ariaLabel}
+			style={buttonStyle}
+		/>
+	);
+};
+
+createRoot(root).render(
```
manvinderjit, line 32: @kbventures This is a just an example / guide file that shows how to use the text input. It will not be pushed to production. If you want I can move these to CSS file.

src/pages/text-input-usage/sample-text-inputs.tsx
```diff
@@ -0,0 +1,128 @@
...
+	(
+		<StrictMode>
+			<div className='App'>
```
kbventures, line 35: Please move the inline CSS to their respective CSS Files or CSS Layers.

src/components/TextInputComponent/TextInputComponent.tsx
```diff
@@ -0,0 +1,72 @@
...
+	labelSlot,
+	onKeyDown,
+	...rest
+}, ref) => {
+	// Conditional flag checks
+	if (!label) {
+		console.warn('Custom Warning: input must always have a label. Otherwise it fails the WCAG SC 3.3.2 standard.');
+	}
+
+	const elementId = useMemo(() => id ?? `input-${Math.trunc(Math.random() * 10000).toString(16)}`, [id]);
```
kbventures, line 30: The primary purpose of useMemo is "performance optimization".  I understand that's not why we're using it here, but would it not be a better case here for something like useRef, which doesn't need a dependency?

src/components/TextInputComponent/TextInputComponent.tsx
kbventures: @manvinderjit I had deleted that comment. I noticed we needed React 19, and we had decided not to upgrade at this point after writing it. ;)

src/components/TextInputComponent/TextInputComponent.css
```diff
@@ -0,0 +1,49 @@
...
+		&[data-error='true'] { border-color: var(--color-error); }
+
+		&[aria-disabled='true'] {
+			background-color: var(--color-background);
+			cursor: not-allowed;
+			pointer-events: none;
+		}
+	}
+
+	&:has(input:required) > span.input-label-container > label::after {
```
kbventures: Nitpick:

Is there a reason why we're switching between a pseudo-class select and a full attribute selector?

There's no functional difference.  It hinders readability and consistency.

kbventures (changes requested):
Great work. 

I left some notes and actionable items.

These inputs will likely be used across the project (Create an issue for that), and a component layer CSS page would also make sense here for the input for that reason.

src/components/TextInputComponent/TextInputComponent.tsx
madcampos: That was a change I made. We do depend on `id` because it can potentially be set by the user or be changed by the user.

The use of `useMemo` here is not for "expensive computation" but more for caching the `Math.random` result.

We could in theory have the variable just like this:
```ts
const elementId = id ?? `input-${Math.trunc(Math.random() * 10000).toString(16)}`;
```

This means if the consumer of the component did not provide any id, we generate a random string to use. The problem then is that every time the component re-renders the random string will change.

With `useMemo`, the random part is cached and will not be recomputed.

src/components/TextInputComponent/TextInputComponent.css
madcampos: @kbventures I don't understand what you mean with your question...
What you mean by "switching between pseudo-class select and full attribute selector"?

src/components/TextInputComponent/TextInputComponent.css
```diff
@@ -0,0 +1,49 @@
...
+		color: var(--color-accent);
+		content: 'Required' / '';
+		font-size: 0.75rem;
+		font-weight: var(--weight-semibold);
+		letter-spacing: 0;
+		text-transform: uppercase;
+	}
+}
+
+.text-input-component-container:has(input[aria-disabled]) label { color: var(--color-helper); }
```
madcampos:
```suggestion
	&:has(input[aria-disabled]) label { color: var(--color-helper); }
}
```

src/pages/text-input-usage/sample-text-inputs.tsx
madcampos: I will be very picky again: even if it is a test file, it helps to build the habit of doing things write. It is like flossing your teeth, very annoying, but if you do it every day it becomes muscle memory, and you get to a point where you don't even notice you are doing it.

Same with code, it may be annoying to write things "the right way", but if we do it even for throw away pieces of code, we develop a habit and writing better code becomes muscle memory.

src/pages/text-input-usage/sample-text-inputs.tsx
```diff
@@ -0,0 +1,128 @@
...
+				{/* NOTE: This is temp UI to test input component design */}
+
+				<h2 style={{ textAlign: 'center' }}>Render all the Design System's Text Input Component Cases</h2>
+				<h2 style={{ textAlign: 'center' }}>
+					<span style={{ color: 'hsla(2, 84%, 56%, 1)' }}>NOTE:</span> The input must always have a label to meet the WCAG SC 3.3.2 standard.
+				</h2>
```
madcampos:
```suggestion
				<p>
					<em>NOTE:</em> The input must always have a label to meet the WCAG SC 3.3.2 standard.
				</p>
```

madcampos (changes requested):

src/components/TextInputComponent/TextInputComponent.css
madcampos:
```suggestion
@layer components {
  .text-input-component-container {
	  display: flex;
	  flex-direction: column;
	  gap: var(--spacing-xs);
  
	  .input-label-container {
		  align-items: center;
		  display: flex;
		  justify-content: space-between;
  
		  & > label {
			  color: var(--color-foreground);
			  display: flex;
			  flex-grow: 1;
			  font-size: var(--text-sm);
			  font-weight: var(--weight-semibold);
			  justify-content: space-between;
		  }
	  }
  
	  & > input {
		  background-color: var(--color-card);
		  border: var(--border);
		  border-radius: var(--rounded-xs);
		  gap: var(--spacing-xs);
		  height: auto;
		  padding: var(--spacing-xs);
		  width: 100%;
  
		  &[data-error='true'] { border-color: var(--color-error); }
  
		  &[aria-disabled='true'] {
			  background-color: var(--color-background);
			  cursor: not-allowed;
			  pointer-events: none;
		  }
	  }
  
	  &:has(input:required) > span.input-label-container > label::after {
		  color: var(--color-accent);
		  content: 'Required' / '';
		  font-size: 0.75rem;
		  font-weight: var(--weight-semibold);
		  letter-spacing: 0;
		  text-transform: uppercase;
	  }
  }
  
  .text-input-component-container:has(input[aria-disabled]) label { color: var(--color-helper); }
}
```

manvinderjit: Is applied with layers.

src/components/TextInputComponent/TextInputComponent.css
kbventures: @madcampos, here we are using two different ways for doing the SAME thing.  Why?

src/pages/text-input-usage/sample-text-inputs.tsx
manvinderjit: Done.

manvinderjit:
@madcampos @kbventures made the requested changes. Please let me know if anything else needs to be changed. Thanks.

madcampos:
@manvinderjit formatter is failing, please format the failing file: https://github.com/torontojs/communityhub/actions/runs/16277118910/job/45958590008?pr=163

manvinderjit:
> @manvinderjit formatter is failing, please format the failing file: https://github.com/torontojs/communityhub/actions/runs/16277118910/job/45958590008?pr=163

@madcampos Formatting resolved. It was because I committed a change from github itself. 

On a side note, I have found that running `"format": "dprint fmt --staged"` locally does not work for me. I have to run `dprint fmt` without the flag to make it work.

kbventures (approved):
Great work.

madcampos (approved):

PR #165, opened 2025-07-06
feat: add new password rules to enhance security

### LINKED ISSUE #158: Fix API allowing weak passwords without validation-FE
Subtask for FE work

bcha92:
## Summary

Add new password rules to enhance security
1. Password must be at least 15 characters long
2. Password strength must be 3 or higher

## Tests
Unknown?


src/components/SignUp/SignUpForm.tsx
```diff
@@ -13,16 +13,30 @@ const SignUpForm = (): React.JSX.Element => {
...
 	const [isLoading, setIsLoading] = useState<boolean>(false);
+	const [invalidPasswordMessage, setInvalidPasswordMessage] = useState<string>('');
 
 	const handleOnInput = (event: React.ChangeEvent<HTMLInputElement>) => {
 		const password = event.target.value;
 		const result = zxcvbn(password);
-		setStrength(result.score);
-		setFeedback(result.feedback.suggestions.join(', '));
+		setStrength(result.score < 3 ? result.score : password.length < 15 ? 2 : result.score);
+		setFeedback([...result.feedback.suggestions, password.length < 15 ? 'Password is shorter than 15 characters' : ''].join(', '));
```
madcampos: **nitpick:** Can we use regular ifs here instead of nested ternaries please? Nested ternaries can get hard to read very quickly

madcampos (changes requested):

src/components/SignUp/SignUpForm.tsx
kbventures: I think we should have this in our contribution.MD. No nested ternary.

madcampos: @kbventures it is part of the eslint rules and I think it is already on the contribution guide

kbventures (changes requested):
The submit button should be disabled until a password meeting our requirements is selected.

src/components/SignUp/SignUpForm.tsx
```diff
@@ -13,16 +13,40 @@ const SignUpForm = (): React.JSX.Element => {
...
+		if (result.score < 3) {
+			setStrength(result.score);
+		} else if (password.length < 15) {
+			setStrength(2);
+			feedback.push('Password is shorter than 15 characters');
+		} else {
+			setStrength(result.score);
+		}
+
+		setFeedback(feedback.join(', '));
```
madcampos, line 32: **suggestion**: what about we use isomorphic JS here?

Expanding on that: this kind of function is used by both the backend and frontend. So it should be in a shared utils folder and be called from both sides.

bcha92: I think that's a much better idea to have a shared util function logic used for both BE and FE.

madcampos (approved):
Approving with the caveat that we refactor the code to reuse parts on both back and frontend in the future.

kbventures (approved):
Approved. Great work.  Please move the isomorphic sharing of functions to a separate issue(low priority)

PR #173, opened 2025-07-11
Feat/172 helper message component

### LINKED ISSUE #172: [Design System] Create Helper Message Component
## Description

Need helpers for text inputs, selects or other interactive elements, to provide additional information and, optionally, status regarding the placed content. Helpers can give input-related hints and also indicate whether the user has successfully filled the input, or if there's still any mistake yet to fix.

## Proposed solution

Add a helper message component that can be added to different input and interactive elements. 

## Additional context

The following image shows the design from the design system.

Image

manvinderjit:
## Summary

Need helpers for text inputs, selects or other interactive elements, to provide additional information and, optionally, status regarding the placed content. Helpers can give input-related hints and also indicate whether the user has successfully filled the input, or if there's still any mistake yet to fix.

## Tests

Pending

## Additional information

Closes #172 

Adds helpers based on the following figma design:

image


src/components/HelperMessageComponent/HelperMessageComponent.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface Props {
+	variant?: HelperMessageComponentVariants;
+	labelText: string;
+}
+
+const HelperMessageComponent = ({ variant = 'default', labelText }: Props) => {
+	return (
+		<div className='helper-container' data-variant={variant}>
+			<span className='helper-icon' />
```
bcha92: Not a blocker per se, but given we have `svg` icons to load for the `.helper-icon` elements and for the most part they don't necessarily need to be loaded at the start of the page, would implementing lazy loading be useful or beneficial in terms of speed and FCP (First Contentful Paint) in a reusable component like this one?
```jsx
<span className="helper-icon" loading="lazy" />
```

But otherwise I think it looks good to go 👍

madcampos: We are using react, FCP is already an issue on itself 🙃

That being said, I don't think it will particularly help as there is no easy way to add lazy loading to SVG elements, there is a native attribute for both images and iframes, but not for SVGs.

If we would implement it ourselves, there would be added complexity in the form of creating an intersection observer, then removing it when the svg was loaded, then adding some request for the svg content, all while keeping track of things using react and state.

One very viable option instead would be to have a shared svg definition added to the page, and then each icon could be only a `use` tag. That removes duplication, while still allowing for flexibility.

src/components/HelperMessageComponent/HelperMessageComponent.css
```diff
@@ -0,0 +1,59 @@
+@layer components {
```
kbventures, line 1: Nitpick:  Add a simple one to a page so we can see what it looks like.  Great work btw.

kbventures (dismissed):

src/components/HelperMessageComponent/HelperMessageComponent.css
madcampos, line 1: **nitpick/question**: Why don't we use component icons instead of the background images?

With live SVGs inside the DOM we can target the icons with CSS and change things like color and animate them.

manvinderjit: @madcampos We used to add component icons but you recommended to use them as background images. Where do you recommend we use the former vs the latter?

src/components/HelperMessageComponent/HelperMessageComponent.tsx
manvinderjit: @madcampos You mean like this 
`<svg className="icon">
    <use href="" />
  </svg>`

src/components/HelperMessageComponent/HelperMessageComponent.css
manvinderjit: @kbventures Do you want me to add a separate page for these or include them in the inputs page?

src/components/HelperMessageComponent/HelperMessageComponent.css
madcampos: Background images work for things like the arrow on a select, or as the list indicator for a list item. Those are situations where the image is an enhancement to a single element.

For components with more than one element, then using an SVG is usually better.

The other point is styling, if we need to control or change how the icon will be styled (which is almost always), then using SVG is better as we can style it with CSS. An SVG image cannot have dynamic styles.

src/components/HelperMessageComponent/HelperMessageComponent.tsx
madcampos: Yes. It is a combo:

First the defs as a global element.
```html
<html>
  <body>
    <svg id="icon-definitions" aria-hidden="true" role="none">
      <defs>
        <!-- All icon defs go here -->
      </defs>
    </svg>
    ...
  </body>
</html>
```

Then the component will use the definitions as you said:
```html
<svg class="icon" data-icon="..."><use href="#..." /></svg>
```

The caveats to that are:
- The first SVG still takes space in page, so we need to hide it with CSS.
- The first SVG cannot be completely hidden with things like `display: none` or else effects, filters, masks, and other things will not work, so it needs to be hidden using a technique similar to the sr only class.
- It doesn't work inside web components as the id will not exist inside the Shadow DOM.

manvinderjit: @madcampos Considering we have a multi page app, should we build a separate component for svg definitions and reuse it on each page?

madcampos: That works, yes.

The ideal solution would be to statically include the svg, but for now that works.

src/components/HelperMessageComponent/HelperMessageComponent.css
manvinderjit: Done

src/components/HelperMessageComponent/HelperMessageComponent.tsx
manvinderjit: Added the definitions as a Component.

src/components/HelperMessageComponent/HelperMessageSvgDefinitions.tsx
```diff
@@ -0,0 +1,74 @@
...
+						strokeLinecap='round'
+						strokeLinejoin='round'
+					/>
+					<path d='M8 6V8.66667' stroke='#DE9502' strokeWidth='2' strokeLinecap='round' strokeLinejoin='round' />
+					<path d='M8 11.3335H8.00667' stroke='#DE9502' strokeWidth='2' strokeLinecap='round' strokeLinejoin='round' />
+				</symbol>
+			</defs>
+		</svg>
+	);
+};
```
madcampos: **suggestion**: we can make it a simple script, react doesn't have to _know_ about this so it won't have it's feelings hurt.

```typescript
export function initializeIconDefinitions() {
  if (!document.querySelector('#icon-definitions')) {
    document.body.insertAdjacentHtml('afterbegin',
      // SVG string goes here
    );
  }
}
```
 
Then on the component we `import` this function and call it:
```typescript
import { initializeIconDefinitions } from './icon-definitions.ts';

export function RandomIconComponent() {
  initializeIconDefinitions();

  //...
}
```

If you want to even go further, there is another option to use the module import as a side effect and take advantage of the top level `await` in modules.

The definition module would be rewritten as such:
```typescript
// No export here as it will be called and used as a side effect.
// Note the function is also now async.
async function initializeIconDefinitions() {
  if (!document.querySelector('#icon-definitions')) {
    const result = await fetch('/icon-defs.svg');
    const svgText = await result.text();

    document.body.insertAdjacentHtml('afterbegin', svgText);
  }
}

// Top level await in modules is supported in browsers for a while.
await initializeIconDefinitions();
```

Then the import inside the component would just be:
```typescript
// This import definition is executed for side effects.
// Modules are async due to network fetching and all, so it Just Works™
import './icon-definitions.ts';

// ...
```

src/pages/usage-helper-message/sample-helper-message.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+createRoot(root).render(
+	(
+		<StrictMode>
+			{/* IMPORTANT: Must import the SVG Definitions Component - HelperMessageSvgDefinitions, to use the helper message */}
+			<HelperMessageSvgDefinitions />
+			<div className='App'>
+				{/* NOTE: This is temp UI to test the helper message usage with text input or other components */}
+
+				<h2 style={{ textAlign: 'center' }}>Render the Design System's Helper Message Input Component Cases</h2>
```
madcampos: **issue**: there is no `h1` on this page.

src/pages/usage-helper-message/sample-helper-message.tsx
```diff
@@ -0,0 +1,80 @@
...
+				<p>
+					<em>IMPORTANT:</em> Must import the SVG Definitions Component - HelperMessageSvgDefinitions, to use the helper message
+				</p>
+
+				<div className='component-container'>
+					<h3>
+						Default Variant:
+					</h3>
+					<HelperMessageComponent variant='default' labelText='Helper Message' />
```
madcampos:
```suggestion
					<fieldset>
						<legend>Default Variant</legend>
						<HelperMessageComponent variant='default' labelText='Helper Message' />
					</fieldset>
```

For those scenarios where you are exposing sets of components and there is a need for grouping it is better to use the [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/fieldset) or the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details) (with the `open` attribute set) elements.

The `<details>` element is preferred for longer lists or where the section contains a lot of information, so it can be collapsed. But for just laying out multiple states and scenarios, prefer the `<fieldset>` element

src/pages/usage-helper-message/style.css
```diff
@@ -0,0 +1,28 @@
+@layer components {
+	p { text-align: center; }
+
+	em {
+		font-style: italic;
+		font-weight: var(--weight-bold);
+	}
+
+	.component-container {
```
madcampos: **issue**: please use a more descriptive name to avoid clashes and improve clarity.

madcampos (changes requested):

src/pages/usage-helper-message/sample-helper-message.tsx
manvinderjit: I generally leave the h1 for the index.html page instead of the component, but I can change it because we are not going to use it here.

src/components/HelperMessageComponent/HelperMessageSvgDefinitions.tsx
manvinderjit: @madcampos Updated the code using the latter.

madcampos (approved):

PR #184, opened 2025-07-14
feat: add Date Selector component

### LINKED ISSUE #31: Reusable Date Selector Component

manvinderjit:
## Summary

Add a Reusable Date Selector component for the design system

## Tests

Pending

## Additional information

Closes #31 

Screenshot 2025-07-14 at 16-05-48 User Profile


src/components/DateSelector/DateSelector.css
```diff
@@ -0,0 +1,22 @@
+@layer components {
+	.dob-container {
```
kbventures: .dob-container

Since I don't know what .dop is, it's safe to assume naming can be improved here?

src/components/DateSelector/DateSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+ * Checks whether a date string is in the valid MM-DD format.
+ *
+ * A valid format must follow the pattern: two-digit month (01–12)
+ * followed by a dash and a two-digit day (01–31).
+ * This function does not validate actual calendar dates (e.g., "02-30" would be considered valid).
+ *
+ * @param date - The date string to validate, expected in MM-DD format.
+ * @returns `true` if the format is valid, otherwise `false`.
+ */
+const isValidDateFormat = (date: string): boolean => {
```
kbventures, line 22: Handling dates is notoriously prone to errors.  

Why are we not considering a library instead?

https://www.npmjs.com/package/date-fns

date-fns lets you import only the specific functions you need.

kbventures (changes requested):

src/components/DateSelector/DateSelector.tsx
madcampos: I'd rather use the [temporal polyfill](https://github.com/js-temporal/temporal-polyfill). But yes, handling dates is hard and error prone

src/components/DateSelector/DateSelector.css
madcampos: Date of birth? 👀

src/components/DateSelector/DateSelector.tsx
```diff
@@ -0,0 +1,179 @@
+import { type ChangeEvent, useEffect, useState } from 'react';
+import './DateSelector.css';
+
+interface Props {
+	labelContent?: string;
+	setDateValue: React.Dispatch<React.SetStateAction<string>>;
```
madcampos: **issue**: changes to a component should be _events_ and not passing state setters down.

Please change it to a `onChange` event handler, and then the consumers for this component should set the state when the function is called.

src/components/DateSelector/DateSelector.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+	useEffect(() => {
+		if (!isRealDate(tempDateValue)) {
+			setDateValue('');
+			return;
+		}
+		if (isRealDate(tempDateValue)) {
+			setDateValue(tempDateValue);
+		}
+	}, [tempDateValue]);
```
madcampos: This `useEffect` is not needed. When the component prop changes it will get re-rendered anyways. We should instead use a function that receives the prop as input and does this validation.

_If, and only if_, the computation is expensive use a `useMemo` instead.

src/components/DateSelector/DateSelector.css
manvinderjit: I copied this from `CompleteProfile` where it was originally used for date of birth. But since it is reusable, i should just name it date.

src/components/DateSelector/DateSelector.tsx
manvinderjit: @madcampos I am using this useEffect for two reasons:

1. It only updates the `date` state of the parent component when date is valid, i.e., both Month and Day are selected. It keeps the state in the child component until the date is valid.
2. It resets the `date` state of the parent component when the date is invalid, e.g., a user goes from a Month like `March 31` to `Feb`. As switching to Feb won't have the 30 and 31 date, we need to reset it.

src/components/DateSelector/DateSelector.tsx
manvinderjit: @madcampos Should we use the temporal polyfill library or leave the code as is for now?

src/components/DateSelector/DateSelector.tsx
manvinderjit: I have changed it to a handleChange function because I only want to update the parent's state value when both day and month are selected and the date is valid. Not partially. 

More insights here:

https://github.com/torontojs/communityhub/pull/184#discussion_r2217361369

src/components/DateSelector/DateSelector.tsx
madcampos: Leave it as is, we can revisit later

src/components/DateSelector/DateSelector.tsx
madcampos: So, let's take the input element as an inspiration here. We can expose three pieces then:
- `onChange` event, when any of the things are changed. (`onInput` would be redundant here, think of this component similar to a checkbox that doesn't fire the `input` event.)
- The value of the component.
- The validation state of the component.

We still should expose the way to modify internal state to the component as "event handlers" instead of passing the setter function down.

src/components/DateSelector/DateSelector.tsx
madcampos: Thinking about the behaviour of elements like `input`, validation is lazy. We should do the same here and always update the value, but only validate when needed to.

Furthermore, the input itself should work to prevent users from inputting the wrong date.

manvinderjit: @madcampos Since are using two different selects

image

I want to prevent situations wherein user selects only either day or month, not both and then clicks submit

In such a case the state value for the date in parent component will be '09-' or '-12'. 

Should we just add this check at form submission then and not onChange?

madcampos: We should add both: A message when the date is invalid, and then a check when submitting to block subissions on invalid date

src/components/DateSelector/DateSelector.tsx
kbventures: @manvinderjit @madcampos Another reason I mentioned considering a library: 29 will always be selectable—even in non-leap years. date-fns is highly modular, so we could use just what we need and ignore the rest. Create an issue with "Low priority" label.

kbventures (dismissed):
Approved.  Great work .Please create low priority issue so we can deal with updating this  later.

src/components/DateSelector/DateSelector.tsx
madcampos: https://github.com/torontojs/communityhub/issues/197

manvinderjit:
@madcampos Updated the component based on the feedback. Please review. Thanks.

bcha92 (dismissed):
Day selector appears to be fixed, only appearing the appropriate date (i.e. September should have 30 days, not 31).
Only need to merge the base branch.

src/components/DateSelector/DateSelector.tsx
```diff
@@ -1,9 +1,13 @@
-import { type ChangeEvent, useEffect, useState } from 'react';
+import { type ChangeEvent, useEffect } from 'react';
 import './DateSelector.css';
+import HelperMessageComponent from '../HelperMessageComponent/HelperMessageComponent.tsx';
 
 interface Props {
+	dateValue: string;
+	isDateValid: boolean | null;
```
madcampos: **question**: Isn't this prop redundant in the sense that this can be derived here from the provided date and the validation being ran?

Like, the component already runs some validation and calls two handlers:
- When the value updates
- When the value is valid/invalid

src/components/DateSelector/DateSelector.tsx
```diff
@@ -59,9 +63,8 @@ const getDaysInMonth = (month: string) => {
 	}
 };
 
-const DateSelector = ({ handleSetDateValue, labelContent = 'Select Date' }: Props) => {
-	const [tempDateValue, setTempDateValue] = useState<string>('');
-	const [month, day] = tempDateValue ? tempDateValue.split('-') : ['', ''];
+const DateSelector = ({ dateValue, isDateValid, handleSetDateValue, handleSetDateValidity, labelContent = 'Select Date' }: Props) => {
+	const [month, day] = dateValue ? dateValue.split('-') : ['', ''];
```
madcampos:
```suggestion
	const [month = '', day = ''] = dateValue?.split('-') ?? [];
```

src/components/DateSelector/DateSelector.tsx
```diff
@@ -96,42 +99,36 @@ const DateSelector = ({ handleSetDateValue, labelContent = 'Select Date' }: Prop
...
-			handleSetDateValue('');
-			return;
-		}
-		if (isRealDate(tempDateValue)) {
-			handleSetDateValue(tempDateValue);
-		}
-	}, [tempDateValue]);
+		if (dateValue.trim() === '') { return; }
+		handleSetDateValidity(isRealDate(dateValue));
+	}, [dateValue]);
```
madcampos: This should be called after the date value change handler is called, not as an effect. It makes the code cleaner.

src/components/DateSelector/DateSelector.tsx
manvinderjit: @madcampos But the parent needs to know if the date is valid, to prevent form submission, whose state `isDateValid` will be in the parent.

src/components/DateSelector/DateSelector.tsx
manvinderjit: @madcampos How will this impact scenarios wherein the parent component clears all fields, like a reset button?

madcampos: I think it won't be a problem as validation should only be triggered after user input, so the parent setting the value won't "count" as the user has not changed the input yet.

src/components/DateSelector/DateSelector.tsx
madcampos: Yes, the state will live on the parent, and is set by the validation handler. It does not need to be a prop on this component as it can be derived from the `dateValue`.

manvinderjit: @madcampos Addressed, please review.

src/components/DateSelector/DateSelector.tsx
manvinderjit: Addressed, please review.

src/components/DateSelector/DateSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+};
+
+/**
+ * Determines whether a date string represents a real calendar date.
+ *
+ * The input must be in MM-DD format and represent a valid calendar day.
+ * This function uses a placeholder year (2000, a leap year) to account for month lengths,
+ * including February 29th. It assumes no year context is needed and is used primarily for validation.
+ *
+ * @deprecated We will remove update this when Temporal becomes stable
```
madcampos, line 34: **praise:** Nice!

madcampos (approved):

src/components/DateSelector/DateSelector.tsx
manvinderjit: Thank you for the valuable feedback that made it nice.

PR #187, opened 2025-07-20
feat: add useProfileRedirect hook

### LINKED ISSUE #169: [Integration] Integration Complete Profile page with Profile States
## Description

Complete profile must check user's profile state and authorize / redirect accordingly.

## Proposed solution

Need to integrate complete profile page with profile states.

## Additional context

Blocked by #142

manvinderjit:
## Summary

add useProfileRedirect hook that redirects user based on their current profile status.

## Tests

Pending

### DIFF
--- a/src/hooks/useProfileRedirect.ts
+++ b/src/hooks/useProfileRedirect.ts
@@ -12,11 +12,7 @@ interface ProfileStatusResponse {
 
 const REDIRECT_PATHS = {
 	signIn: '/pages/sign-in/',
-	// TODO: Determine how to redirect uses who created account but failed to activate immediately
-	activate: '/pages/activate/',
 	checkSteps: '/pages/check-steps/',
-	// TODO: Determine how to redirect uses who with following profile status
-	leadsToNowhere: '/pages/leads-to-nowhere',
 	completeProfile: '/pages/complete-profile/',
 	home: '/pages/home/'
 };
--- a/src/hooks/useProfileRedirect.ts
+++ b/src/hooks/useProfileRedirect.ts
@@ -23,21 +23,17 @@ const REDIRECT_PATHS = {
 
 const getRedirectPathForStatus = (status: ProfileStatus): string => {
 	switch (status) {
-		case 'created':
-			return REDIRECT_PATHS.activate;
-
 		case 'activated':
 			return REDIRECT_PATHS.checkSteps;
 
-		case 'social-handle-provided':
-			return REDIRECT_PATHS.leadsToNowhere;
-
 		case 'tos-accepted':
+		case 'social-handle-provided':
 			return REDIRECT_PATHS.completeProfile;
 
 		case 'profile-completed':
 			return REDIRECT_PATHS.home;
 
+		case 'created':
 		case 'deleted':
 		case 'error':
 		default:
--- a/src/hooks/useProfileRedirect.ts
+++ b/src/hooks/useProfileRedirect.ts
@@ -1,0 +1,96 @@
+import { useEffect } from 'react';
+
+type ProfileStatus = 'activated' | 'created' | 'deleted' | 'error' | 'profile-completed' | 'social-handle-provided' | 'tos-accepted';
+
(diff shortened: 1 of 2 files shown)


src/hooks/useProfileRedirect.ts
```diff
@@ -0,0 +1,96 @@
+import { useEffect } from 'react';
+
+type ProfileStatus = 'activated' | 'created' | 'deleted' | 'error' | 'profile-completed' | 'social-handle-provided' | 'tos-accepted';
```
madcampos, line 3: **nitpick**: We should reference the same type on the frontend and on the backend

src/hooks/useProfileRedirect.ts
```diff
@@ -0,0 +1,96 @@
...
+
+interface ProfileStatusResponse {
+	id: string;
+	name: string;
+	access: string;
+	avatar: string | null;
+	status: ProfileStatus;
+}
```
madcampos, line 11: **nitpick**: this should also be referenced from the same place in the frontend and backend.

src/hooks/useProfileRedirect.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const REDIRECT_PATHS = {
+	signIn: '/pages/sign-in/',
+	// TODO: Determine how to redirect uses who created account but failed to activate immediately
```
madcampos: They should be sent to the sign-in page. Until they confirm their email, they will be locked out of the platform.

src/hooks/useProfileRedirect.ts
```diff
@@ -0,0 +1,96 @@
...
+	activate: '/pages/activate/',
+	checkSteps: '/pages/check-steps/',
+	// TODO: Determine how to redirect uses who with following profile status
+	leadsToNowhere: '/pages/leads-to-nowhere',
```
madcampos: **question**: what page is this?

src/hooks/useProfileRedirect.ts
```diff
@@ -0,0 +1,96 @@
...
+			return REDIRECT_PATHS.completeProfile;
+
+		case 'profile-completed':
+			return REDIRECT_PATHS.home;
+
+		case 'deleted':
+		case 'error':
+		default:
+			return REDIRECT_PATHS.signIn;
+	}
```
madcampos:
```suggestion
	switch (status) {
		case 'activated':
			return REDIRECT_PATHS.checkSteps;

		case 'tos-accepted':
		case 'social-handle-provided':
			return REDIRECT_PATHS.completeProfile;

		case 'profile-completed':
			return REDIRECT_PATHS.home;

		case 'created':
		case 'deleted':
		case 'error':
		default:
			return REDIRECT_PATHS.signIn;
	}
```

src/hooks/useProfileRedirect.ts
manvinderjit: None. I meant to add a dummy page for the following case.
`case 'social-handle-provided':
			return REDIRECT_PATHS.leadsToNowhere;`

src/hooks/useProfileRedirect.ts
manvinderjit: @madcampos This leads to errors because there are separate tsconfig files for frontend and backend, and their `include` has respective folders only.

When i update the `frontend` tsconfig file to include the backend config ...
image

...I can use the backend types but then i get a bunch of errors when pushing

image

src/hooks/useProfileRedirect.ts
manvinderjit: same issue as above

src/hooks/useProfileRedirect.ts
madcampos: That's okay, we can move the files to a shared folder and then add the needed configuration so it is visible.

madcampos: I will mark this as resolved and we can look into combining back and frontend in the future.

madcampos (approved):

PR #189, opened 2025-07-20
Feature: trim inputs and lowercase emails

madcampos:
## Summary

Trim all user inputs that are "open strings". Types that are things like a UUID or a Date are not trimmed as they already should be provided trimmed or will error out.

All emails will be cast to lowercase.

## Tests

Added a test for signup with mixed case email.

## Additional information

Closes: #97 #124

PR #199, opened 2025-07-21
Normalize emails to lower case back end

kbventures:
## Summary

Normalize emails to lower case back end


madcampos:
This was already covered here: https://github.com/torontojs/communityhub/pull/189

PR #219, opened 2025-07-29
Chore/fix: enable JSX/TSX linting

### LINKED ISSUE #218: estling config shoudl not all allow fetch without await
### Description

This should not pass the ESLint check because fetch is asynchronous but not awaited, so "result" is an unhandled Promise, not the actual response object
```
useEffect(() => {
		try {
			const result = fetch('/api/auth/valid-reset-pw-token', {
				method: 'POST',
				headers: {
				},
				body: JSON.stringify(token)
			});
			if (!result) {
				setIsValidToken(false);
			}

			setIsValidToken(true);
		} catch (error) {
			console.error(error);
		}
		
	}, []);
```

madcampos:
## Summary

Enables linting for `.jsx` and `.tsx` files that was disabled before.

## Tests

Re-linted the project and fixed issues.

## Additional information

I've fixed all of the linting errors and some of the warnings as well.

### DIFF
--- a/src/components/CompleteProfile/CompleteProfile.tsx
+++ b/src/components/CompleteProfile/CompleteProfile.tsx
@@ -49,7 +49,7 @@ const platformEnum = ['site', 'slack', 'linkedin', 'github', 'portfolio', 'codep
  * @returns `true` if the format is valid, otherwise `false`.
  */
 const isValidBirthdayFormat = (birthday: string): boolean => {
-	const regex = /^(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])$/;
+	const regex = /^(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])$/iu;
 	return regex.test(birthday);
 };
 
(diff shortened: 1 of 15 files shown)


src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -72,7 +72,8 @@ const isRealDate = (birthday: string): boolean => {
 	const month = Number(monthStr);
 	const day = Number(dayStr);
 
-	const date = new Date(2000, month - 1, day);
+	const REFERENCE_YEAR = 2000;
+	const date = new Date(REFERENCE_YEAR, month - 1, day);
```
bcha92, line 76: Is a separate `REFERENCE_YEAR` variable needed for any reason?
Or is it just for naming convention clarity?

bcha92 (commented):
not a blocker

sounmind (approved):
LGTM!

src/components/CompleteProfile/CompleteProfile.tsx
madcampos: Just for clarity. It shuts up the linting warning for magic numbers. Random numbers without an explanation of what they mean are usually a bad idea as the context of what something like "1234" means

kbventures (approved):
Great work. Ready to be merged!

PR #222, opened 2025-08-06
Feat/98 file upload api

### LINKED ISSUE #98: File upload API

manvinderjit:
## Summary

Add file upload api for uploading files to R2 buckets.

## Tests
WIP - Pending

## Additional information

Closes #98


api/src/routes/files/data.ts
```diff
@@ -0,0 +1,45 @@
+import { DBTables, SCHEMA_VERSION } from '../../utils/db.ts';
+import type { AllowedFileMimeType, FileAccessLevelType, UploadedFileType } from './validation.ts';
+
+export const fetchFileInfo = async (database: D1Database, fileNameAsId: string, fileAccessLevel: FileAccessLevelType = 'public', isAuthenticated = false) => {
+	// Search all files if requesting protected file and authenticated, otherwise only search public files
+	const dbQuery = fileAccessLevel === 'protected' && isAuthenticated ?
+		`SELECT id AS fileName, mimeType, accessLevel
```
madcampos: **thought:** ID and file name should be different if we want to retain the file name.

The ID has to be unique and is use to reference the file. The file name can be anything, and may even be the same name from other files.

Imagine I have `profile-picture.jpg` and you also have a file with the same name. They are two diferent files, but the name is the same.

api/src/routes/files/data.ts
```diff
@@ -0,0 +1,45 @@
...
+			WHERE
+				id = ?
+			LIMIT 1
+			` :
+		`SELECT id AS fileName, mimeType, accessLevel
+			FROM ${DBTables.UPLOADS}
+			WHERE
+				id = ?
+				AND accessLevel = 'public'
+			LIMIT 1`;
```
madcampos:
```suggestion
	const accessLevelCheck = fileAccessLevel === 'protected' && isAuthenticated ?  "AND accessLevel = 'public'" : '';
	const dbQuery = `
		SELECT id AS fileName, mimeType, accessLevel
		FROM ${DBTables.UPLOADS}
		WHERE
			id = ?
			${accessLevelCheck}
		LIMIT 1
	`;
```

api/src/routes/files/data.ts
```diff
@@ -0,0 +1,45 @@
...
+	profileId: string,
+	fileNameAsId: string,
+	fileType: UploadedFileType,
+	fileMimeType: AllowedFileMimeType,
+	accessLevel: FileAccessLevelType = 'public'
+) => {
+	const now = new Date().toISOString();
+
+	const { success } = await database.prepare(`
+		INSERT INTO ${DBTables.UPLOADS} (id, schemaVersion, profileId, type,  insertedAt, mimeType, accessLevel) VALUES (?, ?, ?, ?, ?, ?, ?)
```
madcampos: **question:** What happens when the user uploads a file twice? Will it overwrite the previous one?

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+			}
+		}
+	}),
+	async (context) => {
+		try {
+			const { filename } = context.req.valid('param');
+
+			if (!filename) {
+				return context.json({ message: 'Invalid or no filename provided!' }, StatusCodes.BAD_REQUEST);
+			}
```
madcampos: Zod should already check for missing parameters

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+		if (!file || !(file instanceof File)) {
+			return context.json({ message: 'No or invalid file uploaded. Please provide a file!' } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
+		}
+
+		const fileMimeType = file.type as AllowedFileMimeType;
+
+		// Validate file extension
+		if (!validateFileExtension(file.name, fileMimeType)) {
+			return context.json({ message: 'Invalid image file extension. Only JPEG and PNG are allowed!' } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
+		}
```
madcampos, line 133: This check should be before the public/private one

api/src/routes/files/validation.ts
```diff
@@ -0,0 +1,72 @@
+import { z } from 'zod';
+
+export const FileNameParamSchema = z.object({
+	filename: z.string('Invalid file name!')
+});
+
+export type FileNameParam = z.infer<typeof FileNameParamSchema>;
+
+export const AllowedFileMimeTypesOptions = ['image/png', 'image/jpeg', 'image/jpg'] as const;
```
madcampos:
```suggestion
export const AllowedFileMimeTypesOptions = ['image/png', 'image/jpeg'] as const;
```

The mime type is `image/jpeg`, it applies to two extensions: `jpg` and `jpeg`

api/src/routes/files/validation.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const AllowedFileMimeTypeSchema = z.enum(AllowedFileMimeTypesOptions).describe('The allowed MIME types for the uploaded documents.');
+
+export type AllowedFileMimeType = z.infer<typeof AllowedFileMimeTypeSchema>;
+
+export const mimeToExtensionsMap: Record<AllowedFileMimeType, string[]> = {
+	'image/jpeg': ['jpg', 'jpeg'],
+	'image/jpg': ['jpg', 'jpeg'],
+	'image/png': ['png']
+};
```
madcampos, line 12: I think it is better to use the [`mime`](https://www.npmjs.com/package/mime) library for this.

api/src/routes/files/validation.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export const validateFileExtension = (fileName: string, mimeType: string): boolean => {
+	const lastDotIndex = fileName.lastIndexOf('.');
+	const lastSlashIndex = Math.max(fileName.lastIndexOf('/'), fileName.lastIndexOf('\\'));
+
+	// No extension or extension is at the end (e.g., "file.")
+	if (lastDotIndex <= lastSlashIndex || lastDotIndex === fileName.length - 1) {
+		return false;
+	}
```
madcampos, line 42: This can be simplified to:
```javascript
testString.split('.').pop();
```

It will give you the file extension. If it is an empty string, it means the name ends with a `.`, if it is anything other than letters and numbers (Regex: `/[a-z0-9]+/iu`), it is not an extension.

api/src/routes/files/data.ts
manvinderjit: Each file is treated independent. So, the new file will have a separate uuid and will be stored accordingly. 

1. For files that have a specific purpose, like `avatar`, we can get the latest one by `insertedAt`. Alternatively, We have an `avatar` field in the `profile` table. We can set its value to the `uuid` of the image that is selected or latest. 
And we can also show a user their avatar history, if such a feature is planned. 
2. For files that have no specific purpose, we have each file existing separately.
3. For removing old files, we can use a worker to clean the files that are greater than x days and remove them from database.

api/src/routes/files/data.ts
manvinderjit: Do we need the filename? The id is a uuid being generated.

api/src/routes/files/index.ts
manvinderjit: Added this as a double check in a rare case if zod falters or someone changes the zod schema. I can remove if it is undesired.

api/src/routes/files/validation.ts
manvinderjit: Do we really want to add a whole library for this? The map is just to check the MimeType and the file extension are the same. It is not being used for anything else. i could have used a switch statement but thought this will be easier.

api/src/routes/files/validation.ts
manvinderjit: Thank you for sharing, Good to know. I thought they were separate.

api/src/routes/files/index.ts
manvinderjit: Actually the public/private check should be just before try/catch but your suggestion stands.

api/src/routes/files/data.ts
madcampos: I don't know, it may not be useful for user avatars, but we have to think:  is there a scneario where it will be useful?

api/src/routes/files/data.ts
madcampos: Avatar history is not planned. I don't think any place does that.

A blanket cleanup of files is not an ideal solution, imagine a user uploads their avatar when creating an account and never changes it. Two years have passed. Should this file be removed or kept?

An alternative would be to upload multiple files and then remove the previous one, but how would you identify what file is "the previous one"?

api/src/routes/files/index.ts
madcampos: Yeah, better remove it.

api/src/routes/files/validation.ts
madcampos: For a small subset of extensions and mimes, it is okay to keep a manual list. As long as they are correctly represented, please read those two links:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types
- https://www.iana.org/assignments/media-types/media-types.xhtml

Also note that a mime type can correspond to [multiple extensions](https://en.wikipedia.org/wiki/JPEG#:~:text=.jpg%2C%20.jpeg%2C%20.jpe%0A.jif%2C%20.jfif%2C%20.jfi).

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+			},
+			[StatusCodes.INTERNAL_SERVER_ERROR]: {
+				description: 'Server Error response',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
+		}
+	}),
+	async (context) => {
+		try {
+			const filename = context.req.valid('param').filename?.trim();
```
kbventures: Trim should definitely be in the validation file, and the filename existence as well. 

Something like:

```
const filenameSchema = z.string().trim().min(1, "Filename cannot be empty");
```

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+
+			if (!filename) {
```
kbventures: Does this still need to exist if we action the previous comment?

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+	async (context) => {
+		try {
+			const sessionData = getSession(context);
+			const profile = await getProfileById(context.env.Database, sessionData.id);
+
+			if (!profile || profile.id !== sessionData.id) {
+				return context.json({ message: 'Profile not found' } satisfies StatusResponse, StatusCodes.NOT_FOUND);
+			}
+
+			const filename = context.req.valid('param').filename?.trim();
```
kbventures: These should be in the Zod validation file as stated above.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+
+			if (!filename) {
+				return context.json({ message: 'Invalid or no filename provided!' }, StatusCodes.BAD_REQUEST);
+			}
+
+			// Fetch file data to determine if it exists
+			const fileInfo = await fetchFileInfo(context.env.Database, filename, 'protected', true);
+
+			if (fileInfo) {
+				// TODO: Add a check to verify if the user is allowed to access the file.
```
kbventures: This sounds like something that shouldn't be optional, and Git issue should be created and shown below the comment.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+
+				const r2Bucket = context.env.ExportedFiles;
+				const file = await r2Bucket.get(filename);
+
+				// If file is found, return it
```
kbventures: This is self-explanatory. Comment is unnecessary.

api/src/routes/files/data.ts
```diff
@@ -0,0 +1,41 @@
...
+	`)
+		.bind(fileNameAsId, SCHEMA_VERSION, profileId, fileType, now, fileMimeType, accessLevel)
```
kbventures: now?

This needs to be renamed.

api/src/routes/files/data.ts
```diff
@@ -0,0 +1,41 @@
+import { DBTables, SCHEMA_VERSION } from '../../utils/db.ts';
+import type { AllowedFileMimeType, FileAccessLevelType, UploadedFileType } from './validation.ts';
+
+export const fetchFileInfo = async (database: D1Database, fileNameAsId: string, fileAccessLevel: FileAccessLevelType = 'public', isAuthenticated = false) => {
```
kbventures: Question:

It seems like we're mixing authentication concerns at the data layer. 

Why is this not handled by our middleware?

I realize this might force us to end up with two functions,  but their concerns will then be unambiguous.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+				return context.json({ message: 'Invalid or no filename provided!' }, StatusCodes.BAD_REQUEST);
+			}
+			// Fetch file data to determine if it is public
+			const fileInfo = await fetchFileInfo(context.env.Database, filename);
+
+			if (fileInfo && fileInfo.accessLevel === 'public') {
+				const r2Bucket = context.env.ExportedFiles;
+				const file = await r2Bucket.get(filename);
+
+				// If file is found, return it
```
kbventures: Most of this code is self-explanatory.  No needs for notes for this.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+					return context.body(file.body, {
+						headers: {
+							'Content-Type': contentType,
+							'Content-Disposition': `inline; filename="${filename}"`
+						}
+					});
+				}
+			}
+
+			// If the file is not found
```
kbventures: Same as note above.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+				return context.json({ message: 'Profile not found' } satisfies StatusResponse, StatusCodes.NOT_FOUND);
+			}
+
+			const filename = context.req.valid('param').filename?.trim();
+
+			if (!filename) {
+				return context.json({ message: 'Invalid or no filename provided!' }, StatusCodes.BAD_REQUEST);
+			}
+
+			// Fetch file data to determine if it exists
```
kbventures: No need comments here.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+
+		if (!profile || profile.id !== sessionData.id) {
+			return context.json({ message: 'Profile not found' } satisfies StatusResponse, StatusCodes.NOT_FOUND);
+		}
+
+		const formData = await context.req.formData();
+		const file = formData.get('file') as File;
+		const fileType = formData.get('type') as UploadedFileType;
+
+		// Check if file was uploaded by the user
```
kbventures: Remove this comment. Same reason as above.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+		if (!file || !(file instanceof File)) {
+			return context.json({ message: 'No or invalid file uploaded. Please provide a file!' } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
+		}
+
+		const fileMimeType = file.type as AllowedFileMimeType;
+
+		// Validate file extension
```
kbventures: Unnecessary comments.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,270 @@
...
+		const fileName = `${crypto.randomUUID()}.${extension}`;
+		const stream = file.stream();
+
+		const fileAccessLevel = isPublicFileType(fileType) ? 'public' : 'protected';
+
+		try {
+			const r2Bucket = context.env.ExportedFiles;
+			const r2UploadResult = await r2Bucket.put(fileName, stream);
+
+			// TODO: Determine if we need to create thumbnails
```
kbventures, line 146: Create Github issue and add the link below.

api/src/routes/files/validation.ts
```diff
@@ -0,0 +1,71 @@
...
+
+	const extension = fileName.slice(lastDotIndex + 1).toLowerCase();
+
+	// Return false if MIME type isn't allowed
```
kbventures: No need for comments here.

kbventures (changes requested):
Great work.

api/src/routes/files/validation.ts
kbventures: @manvinderjit, the lite version is extremely small and is widely used, so it would save you from doing research. 

image

api/src/routes/files/index.ts
kbventures: @manvinderjit, we're not introducing code to plan against Zod failing. As for the second scenario, we just need to make sure people don't remove Zod schemas.

api/src/routes/files/index.ts
manvinderjit: Done
https://github.com/torontojs/communityhub/issues/270

api/src/routes/files/data.ts
manvinderjit: Resolve: New file overwrites old file URL in database. Existing upload is deleted from R2 after new one is uploaded.

api/src/routes/files/data.ts
manvinderjit: Removed access level entirely because the philosophy of the project is design to be public. The avatar URL is now stored in the `Profile` table.

api/src/routes/files/validation.ts
manvinderjit: Updated the code.
Right now we only have two MIME types. We can add a library in future if we expand them.

api/src/routes/files/index.ts
manvinderjit: Removed filename altogether because now the api is refactored to fetch files on the R2 prefix and not just the filename.

api/src/routes/files/index.ts
manvinderjit: Removed, same as above.

api/src/routes/files/index.ts
manvinderjit: Removed.

api/src/routes/files/index.ts
manvinderjit: Removed access levels all together for the moment.

api/src/routes/files/data.ts
manvinderjit: Outdated code. No longer required.

api/src/routes/files/data.ts
manvinderjit: Refactored code. No longer required.

api/src/routes/files/validation.ts
manvinderjit: As we are dealing with R2 prefixes that will have delimiter `/`, the suggested code will incorrectly treat  a scenario like `"/path.to/file"` as "to/file" and having extension "file", because split('.') doesn't determine if the dot was in the prefix or the file name.
While this may rarely occur, but having this edge case covered is recommended.

api/src/routes/files/index.ts
```diff
@@ -0,0 +1,174 @@
...
+	FileUploadSchema,
+	type UploadedFileType,
+	validateFileExtension
+} from './validation.ts';
+
+export const fileRoutes = new OpenAPIHono<EnvironmentBindings>({
+	defaultHook: statusResponseFormatter
+});
+
+// NOTE: Add more specific GET routes before the following route only
```
kbventures, line 19: Love to see this new cool feature. 

Some comments. 

A public service for a resource which cost money?  That sounds like  a bad idea for whomevers credit card goes into the Cloudflare account.  The service needs to be scoped
 to retreving Avatars ONLY and with authentication.

kbventures: We're building a image uplaod service specificaly not a file upload service.  I think the naming should reflect that and has mentioned above scoped for that specifially.

kbventures: As per our conversation, we need to code a check to ensure we don't go voer the "free" tier limit.  There's no way to to have alarms or limits setup at the admin console level.   Checking everything we utilize(resource) is not optimal but we can find a more optimal way after the MVP is in production.

kbventures (changes requested):
Please include at least one hurl interagration tests and create a github issue to create unit tests for this feature(image upload).  Our unit tests setup is almost ready.

api/src/routes/files/index.ts
bcha92: +1 on the limiter. If we are using a tier-limited service, we would need to manually setup a logger that tracks this and send an alert once it reaches a certain threshold.

manvinderjit:
Status Update: On hold due to infrastructure decision delegation for Cloudflare R2.

PR #225, opened 2025-08-08
fix: resolve useProfileRedirect race condition

### LINKED ISSUE #224: [BUG]: resolve useProfileRedirect hook api race condition in strict mode
### Description

The useProfileRedirect hook has api race condition in strict mode leading to false redirects to Sign In page. 

Update the hook to resolve this:
1. Add an abort controller
2. Return multiple status fields

manvinderjit:
## Summary

Resolve api race condition in the hook when running it in strict mode

## Tests

NA

## Additional information

Closes #224

### DIFF
--- a/src/hooks/useProfileRedirect.ts
+++ b/src/hooks/useProfileRedirect.ts
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 
 type ProfileStatus = 'activated' | 'created' | 'deleted' | 'error' | 'profile-completed' | 'social-handle-provided' | 'tos-accepted';
 
@@ -8,6 +8,12 @@ interface ProfileStatusResponse {
 	access: string;
 	avatar: string | null;
 	status: ProfileStatus;
+}
+
+interface ProfileStatusResult {
+	statusOkay: boolean;
+	shouldRedirect: boolean;
+	profileStatus: ProfileStatus | null;
 }
 
 const REDIRECT_PATHS = {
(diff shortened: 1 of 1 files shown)


src/hooks/useProfileRedirect.ts
```diff
@@ -37,43 +46,79 @@ const getRedirectPathForStatus = (status: ProfileStatus): string => {
 	}
 };
 
-export const getProfileStatus = async (): Promise<ProfileStatus | null> => {
+export const getProfileStatus = async (signal?: AbortSignal): Promise<ProfileStatusResult> => {
+	let statusOkay = false;
+	let shouldRedirect = true;
+	let profileStatus;
```
bcha92: It might be a good idea to declare an initial value of `null` for `profileStatus` as this may cause a type check error as `undefined` if left undeclared.

src/hooks/useProfileRedirect.ts
```diff
@@ -37,43 +46,79 @@ const getRedirectPathForStatus = (status: ProfileStatus): string => {
...
 		if (response.status !== 200) {
-			return null;
+			statusOkay = false;
+			shouldRedirect = true;
+			profileStatus = null;
+		} else {
+			const data: ProfileStatusResponse = await response.json();
+			statusOkay = true;
+			shouldRedirect = true;
+			profileStatus = isValidStatus(data.status) ? data.status : null;
```
bcha92: suggested change here since there's essentially no change in the initial `let` declarations if status code is not 200:
```ts
		if (response.status === 200) {
			const data: ProfileStatusResponse = await response.json();
			statusOkay = true;
			profileStatus = isValidStatus(data.status) ? data.status : null;
		}

        // you may need to add a new error throw condition to account for err.name === 'AbortError'
        throw new Error("Add your invalid response here");
```

src/hooks/useProfileRedirect.ts
```diff
@@ -37,43 +46,79 @@ const getRedirectPathForStatus = (status: ProfileStatus): string => {
...
+		if (err.name === 'AbortError') {
+			statusOkay = false;
+			shouldRedirect = false;
+			profileStatus = null;
+		} else {
+			console.error('err', err);
+			statusOkay = false;
+			shouldRedirect = false;
+			profileStatus = null;
+		}
```
bcha92: should omit the declarations if no change from initial values. The if/else output is exactly the same. Should we just do this instead if you're going to create another error handling issue for this later?
```ts
catch (err) {
    console.error('err', err); // <---- you should put in the `throw new Error()` value from the try block for logging in here
    shouldRedirect = false;
}
```

bcha92 (changes requested):
Just a few variable declaration cleanup needed that are not necessary if the value is unchanged.

src/hooks/useProfileRedirect.ts
```diff
@@ -37,43 +46,79 @@
 	}
 };
 
-export const getProfileStatus = async (): Promise<ProfileStatus | null> => {
+export const getProfileStatus = async (signal?: AbortSignal): Promise<ProfileStatusResult> => {
+	let statusOkay = false;
+	let shouldRedirect = true;
```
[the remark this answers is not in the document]
bcha92: ^pretty much this is the warning for the review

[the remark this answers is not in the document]
manvinderjit: @bcha92 I just refactored the whole code. I realized I didn't need to make it this complex. 

Should have done better the first time. Apologies for the inconvenience.

bcha92 (approved):

madcampos (approved):

kbventures (approved):

PR #236, opened 2025-08-13
Implement authgate

### LINKED ISSUE #228: Sign In Page
### Description

Integrate hooks in Sign In page

kbventures:
## Summary

AuthGate HOC component checking for authentication and authorization.  

NOTE: Remove strict mode for testing; otherwise, it may not function as intended.


src/hooks/useHeartBeat.ts
```diff
@@ -16,14 +16,14 @@ const getHeartBeat = async (): Promise<boolean> => {
...
-export const useAuthCheck = () => {
+export const useHeartbeat = () => {
 	const [isAuth, setIsAuth] = useState<boolean>(false);
 
 	useEffect(() => {
 		const check = async () => {
 			const auth = await getHeartBeat();
 			if (auth) { window.location.href = '/pages/home'; }
-			setIsAuth(auth);
+			setIsAuth(true);
```
bcha92: how did we miss this?

api/src/routes/auth/index.ts
```diff
@@ -237,6 +237,167 @@ authRoutes.openapi(
...
+		},
+		responses: {
+			[StatusCodes.OKAY]: {
+				description: 'Password recovery request processed successfully. If the email exists in our system, a recovery link has been sent to the provided email address.',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.BAD_REQUEST]: {
+				description: 'Invalid request data. The email format is invalid or required fields are missing.',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
```
bcha92, line 260: Since we already have a 2.x.x OKAY status response here regardless of whether an email is found or not, I think we should get rid of this 4.x.x BAD_REQUEST response and replace with something else that might be useful such as a 5.x.x INTERNAL_SERVER_ERROR as a general catch block that should be used in the event server is unresponsive?

But if this is just an unnecessary addition to the complexity we have in this logic, we can simply get rid of this?

api/src/routes/auth/validation.ts
```diff
@@ -75,3 +77,23 @@ export type SignUpData = z.infer<typeof SignUpSchema>;
...
+
+export const ForgotPasswordSchema = z.object({
+	email: ProfileSchema.shape.email
+});
+
+export type ForgotPasswordData = z.infer<typeof ForgotPasswordSchema>;
+
+export const ResetPasswordSchema = z.object({
+	password: z.string({ error: 'Password is required' })
+		.min(MIN_PASSWORD_LENGTH),
```
bcha92, line 89: remind me if MIN_PASSWORD_LENGTH wasn't the only parameter we set up here as a minimum password requirement?
I recall on the backend, we also added CC characters to enforce only ASCII characters should be used here (thanks @madcampos !).

for the MIN_PASSWORD_LENGTH variable, we already have a similar variable being used in the sign up endpoint. Since this var is on the backend, it's entirely possible to create a shared MIN_PASSWORD_LENGTH variable that can be used here and on the create password endpoint.

bcha92 (dismissed):
Will approve with minor (non-critical) possible fixes and improvements commented

api/src/routes/auth/validation.ts
madcampos: Yes, using a shared variable, and also a shared zod schema would be ideal.

We can add the CC character checks to the zod schema as well.

src/components/AuthGate/AuthGate.tsx
mahsakarami-dev, line 1: I think there’s a lot of repeated logic across this file.
we could consider creating a HOC that takes a hook as an argument and exports the different gates.
it would help reduce duplication.

src/components/ResetPassword/ResetPassword.tsx
s1m0n1au, line 129: In most cases, React recommend using controlled components to implement forms by default, making the state the single source of truth for the form inputs.

React Docs:
[Controlled and uncontrolled components](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components)
[Uncontrolled Components](https://legacy.reactjs.org/docs/uncontrolled-components.html)
[Controlled Components](https://legacy.reactjs.org/docs/forms.html#controlled-components)

manvinderjit: @s1m0n1au While using controlled components is the default way of creating forms, it is also situational. 
Here are a few reasons why the project uses uncontrolled components:

1. The hub is being developed with the aim of keeping it simple and lightweight. With uncontrolled components, we limit the state maintained and also re-rendering due to state updates. We only need the values and their validations on form submission. We do use controlled components at some places, as required.

2. Using an uncontrolled component allows us to leverage the built-in form validation methods if needed, which is ideal here because we want the product to be simple. 
https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/Form_validation#using_built-in_form_validation

3. The original tech stack proposed for the project was Vanilla JavaScript and CSS. But @kbventures advocated for React because of its popularity in the hiring market. Hence, a compromise was reached. 

As said before, there are a lot of quirks in this project, such as using Vanilla React instead of Next or Remix, having Multiple Pages instead of React Router, and minimizing the reliance on external libraries.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -1,75 +1,92 @@
...
-					<div className='confirm-account-page-btn'>
-						<Button isPrimary size='medium'>
-							Confirm my e-mail
-						</Button>
+	const handleOnClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+		event.preventDefault();
+		if (token.current) {
+			void authenticateAccount(token.current);
+		}
+	};
```
manvinderjit, line 61: **issue:** This will only run on mouse clicks. It will not run if a user presses `enter` or triggers the button via `keyboard`. This is undesired behaviour from UI/UX experience and accessibility viewpoints.

Wrapping the confirmation inside a `<form />` tag and then using the `onSubmit` handler will ensure consistent behaviour.

src/components/ConfirmAccount/ConfirmAccount.tsx
```diff
@@ -1,75 +1,92 @@
...
+				</footer>
 			</div>
-			<footer>
-				<p className='text-helper'>TorontoJS - 2025</p>
-			</footer>
-		</div>
-	);
+		);
+	}
+	return <h1>Is Loading...</h1>;
```
manvinderjit, line 88: **issue**: as Marco previously highlighted this, it is best to use an early return.

Refer: https://github.com/torontojs/communityhub/pull/238#discussion_r2282585892

src/components/AuthGate/AuthGate.tsx
```diff
@@ -0,0 +1,9 @@
+export const AuthGate = ({ children, hook }: { children: React.ReactNode, hook(): boolean | null }) => {
```
manvinderjit, line 1: nitpick: using `null` as a return for hook may cause unforseen issues. It did for my useProfileRedirect hook.

Refer: https://github.com/torontojs/communityhub/pull/238#discussion_r2298566137

src/hooks/useProfileRedirect.ts
```diff
@@ -108,48 +108,34 @@ export const getRedirectPath = async (signal?: AbortSignal, currentPath?: string
...
-					setIsRedirecting(false);
-				}
-			} catch (error) {
-				if (error.name === 'AbortError') {
-					// Do nothing for Abort Error
-					return;
-				}
-				console.error('Redirect Error!', error);
-				window.location.href = REDIRECT_PATHS.signIn;
+			const redirectPath = await getRedirectPath(signal);
```
manvinderjit: **issue**: the `currentPath` parameter, although _optional_, must be passed to getRedirectPath function specifically for `Check Steps` and `Review Conduct Code` pages. 

Both pages require the _profileStatus_ to be `activated`. 
If we don't provide the currentPath, the Review Conduct Code page will redirect back to Check Steps, causing an infinite loop. 

```suggestion
			const redirectPath = await getRedirectPath(signal, currentPath);
```

causes the issue https://github.com/torontojs/communityhub/pull/236#discussion_r2437984796

src/hooks/useProfileRedirect.ts
```diff
@@ -108,48 +108,34 @@ export const getRedirectPath = async (signal?: AbortSignal, currentPath?: string
...
+
+			if (!redirectPath) {
+				setRedirectionComplete(true);
+				return;
+			}
+
+			const currentPath = new URL(window.location.href).pathname;
```
manvinderjit: issue: this must be declared above line 119 as per the previous comment.

src/components/CheckSteps/CheckSteps.tsx
```diff
@@ -1,40 +1,31 @@
...
+					<p className='text-h6'>Check the TorontoJS's conduct code</p>
+				</li>
+				<li className='card'>
+					<CompleteProfileIcon />
+					<p className='text-h6'>Complete your profile</p>
+				</li>
+			</ul>
+			<ButtonLink id='continue-button' isPrimary isLarge href='/pages/review-conduct-code/'>
+				Let's continue
+			</ButtonLink>
```
manvinderjit, line 26: **issue**: Click the button doesn't redirect to `Review Conduct Code` page.

It is because both the Check Steps and Review Conduct Code pages are designed to have the same profile status. Please refer to the comments below in the useProfileRedirect section.

Refer to https://github.com/torontojs/communityhub/pull/236#discussion_r2437982392

src/hooks/useHeartBeat.ts
```diff
@@ -1,21 +1,178 @@
...
+// Public Page useHeartBeat
 export const useHeartBeat = () => {
+	const [isAuth, setIsAuth] = useState<boolean | null>(null);
+
+	useEffect(() => {
+		const check = async () => {
+			const auth = await getHeartBeat();
+			if (auth) { window.location.href = '/pages/home'; }
+			else {
+				setIsAuth(true);
```
manvinderjit, line 29: question: Shouldn't the value be set as false here?

src/hooks/useHeartBeat.ts
```diff
@@ -1,21 +1,178 @@
...
+		return false;
+	} catch (error) {
+		console.error(error);
+		return false;
 	}
 };
 
+// Public Page useHeartBeat
 export const useHeartBeat = () => {
+	const [isAuth, setIsAuth] = useState<boolean | null>(null);
```
manvinderjit, line 22: suggestion: instead of using `null`, would it be better to use a state like 'pending' to avoid null related issues.

Same for other occurrences.

src/hooks/useHeartBeat.ts
```diff
@@ -1,21 +1,178 @@
...
+};
+
+// Protected page hook
+export const useHeartBeatProtected = () => {
+	const [isAuth, setIsAuth] = useState<boolean | null>(null);
+
+	useEffect(() => {
+		const check = async () => {
+			const auth = await getHeartBeatProtected();
+			if (!auth) { window.location.href = '/pages/home'; }
```
manvinderjit: **typo**: incorrect window target location
```suggestion
			if (!auth) { window.location.href = '/pages/sign-in'; }
```

src/hooks/useHeartBeat.ts
```diff
@@ -1,21 +1,178 @@
...
+};
+
+// Protected organizers and admins page hook
+export const useHeartBeatProtectedOrganizer = () => {
+	const [isAuth, setIsAuth] = useState<boolean | null>(null);
+
+	useEffect(() => {
+		const check = async () => {
+			const auth = await getHeartBeatProtectedOrganizer();
+			if (!auth) { window.location.href = '/pages/home'; }
```
manvinderjit: **typo**: same as above
```suggestion
			if (!auth) { window.location.href = '/pages/sign-in'; }
```

src/hooks/useHeartBeat.ts
```diff
@@ -1,21 +1,178 @@
...
+
+// Protected admin page hook
+export const useHeartBeatProtectedAdmin = () => {
+	const [isAuth, setIsAuth] = useState<boolean | null>(null);
+
 	useEffect(() => {
-		void getHeartBeat();
+		const check = async () => {
+			const auth = await getHeartBeatProtectedAdmin();
+			if (!auth) { window.location.href = '/pages/home'; }
```
manvinderjit:
```suggestion
			if (!auth) { window.location.href = '/pages/sign-in'; }
```

src/hooks/useProfileRedirect.ts
```diff
@@ -108,48 +108,34 @@ export const getRedirectPath = async (signal?: AbortSignal, currentPath?: string
...
+
+			if (currentPath !== redirectPath) {
+				window.location.href = redirectPath;
+			} else {
+				setRedirectionComplete(true);
```
manvinderjit, line 138: **issue**: having a catch block to handle errors is required because the `getRedirectPath` throws an error if the api call is aborted. Please check code on line 100.

src/index.html
```diff
@@ -32,10 +32,15 @@ <h1>VMS pages</h1>
 			<li><a href="/pages/print-documents/?document=image-release">Print: Image Release Form</a></li>
 			<li><a href="/pages/print-documents/?document=volunteer-agreement">Print: Volunteer Agreement</a></li>
 			<hr />
+			<li><a href="/pages/protected-page-all">Protected Page</a></li>
+			<li><a href="/pages/protected-page-organizers">Protected Organizers & Admins Page</a></li>
+			<li><a href="/pages/protected-page-admins">Protected Admins Page</a></li>
+			<hr />
```
manvinderjit: **question**: Do we have these pages for demonstration purposes? I am guessing they shouldn't be in production.

manvinderjit (changes requested):
@kbventures Great work on the wrapper hook. 
A few minor typos and a couple of issues to be addressed before merging it.

src/index.html
kbventures: They are currently for desmontration purposes and testings.  They will not making it to the production version.

src/components/ConfirmAccount/ConfirmAccount.tsx
kbventures: @manvinderjit moving this to issue since it's related to my PR but great catch none the less! 

https://github.com/torontojs/communityhub/issues/295

src/components/ConfirmAccount/ConfirmAccount.tsx
kbventures: This code stems from another user.  The page needs to be fully refactored. Out of scope for auth gate.

src/components/AuthGate/AuthGate.tsx
kbventures: It's the only way I was able to get it to work.

src/components/CheckSteps/CheckSteps.tsx
kbventures: @manvinderjit I made some big changes to this section which solved problem.

src/hooks/useProfileRedirect.ts
kbventures: Some big refactor just came.  Please review again.

PR #238, opened 2025-08-16
feat: integrate frontend + add Review Code Conduct api call 

### LINKED ISSUE #223: [issue]: agreeing to code of conduct must update user profile status
### Description

On the `Review Our Conduct Code` page, when a user agrees to code of conduct, it should update their profile status to `tos-accepted`

### LINKED ISSUE #231: Check Steps
### Description

Integrate hooks in Check Steps page

### LINKED ISSUE #232: Code Conduct Page
### Description

Integrate hooks in Code of Conduct page

Blocked by #223

manvinderjit:
add api call to Review Conduct Code page for signing documents.

## Summary

Integrate the frontend pages for the profile completion flow.

## Tests

Requires setting up e2e testing

## Additional information

Closes #223

### DIFF
--- a/src/components/CheckSteps/CheckSteps.tsx
+++ b/src/components/CheckSteps/CheckSteps.tsx
@@ -7,9 +7,9 @@ import StepBar from '../StepBar/StepBar.tsx';
 import './CheckSteps.css';
 
 const CheckSteps = () => {
-	const { redirectionComplete } = useProfileRedirect();
+	const { isRedirecting } = useProfileRedirect();
 
-	if (!redirectionComplete) {
+	if (isRedirecting) {
 		return <div>loading...</div>;
 	}
 
--- a/src/components/CheckSteps/CheckSteps.tsx
+++ b/src/components/CheckSteps/CheckSteps.tsx
@@ -9,32 +9,32 @@ import './CheckSteps.css';
 const CheckSteps = () => {
 	const { redirectionComplete } = useProfileRedirect();
 
-	if (redirectionComplete) {
-		return (
-			<div id='check-steps-wrapper'>
-				<StepBar currentStep={1} steps={[{ label: 'Account confirmed' }, { label: 'Check the conduct code' }, { label: 'Complete your profile' }]} />
-				<div id='check-steps'>
-					<Header isLogoLarge={false}>Welcome to TorontoJS Community Hub!</Header>
-					<p className='intro'>Your account is active, and you're almost ready to get in. Let's complete a few more steps:</p>
-					<ul className='cards'>
-						<li className='card'>
-							<CheckConductCodeIcon />
-							<p className='text-h6'>Check the TorontoJS's conduct code</p>
-						</li>
-						<li className='card'>
-							<CompleteProfileIcon />
-							<p className='text-h6'>Complete your profile</p>
-						</li>
-					</ul>
-					<ButtonLink id='continue-button' isPrimary isLarge href='/pages/review-conduct-code/'>
-						Let's continue
-					</ButtonLink>
-				</div>
-			</div>
-		);
+	if (!redirectionComplete) {
+		return <div>loading...</div>;
 	}
 
-	return <div>loading...</div>;
+	return (
+		<div id='check-steps-wrapper'>
+			<StepBar currentStep={1} steps={[{ label: 'Account confirmed' }, { label: 'Check the conduct code' }, { label: 'Complete your profile' }]} />
+			<div id='check-steps'>
+				<Header isLogoLarge={false}>Welcome to TorontoJS Community Hub!</Header>
+				<p className='intro'>Your account is active, and you're almost ready to get in. Let's complete a few more steps:</p>
+				<ul className='cards'>
+					<li className='card'>
+						<CheckConductCodeIcon />
+						<p className='text-h6'>Check the TorontoJS's conduct code</p>
+					</li>
+					<li className='card'>
+						<CompleteProfileIcon />
+						<p className='text-h6'>Complete your profile</p>
+					</li>
+				</ul>
+				<ButtonLink id='continue-button' isPrimary isLarge href='/pages/review-conduct-code/'>
+					Let's continue
+				</ButtonLink>
+			</div>
+		</div>
+	);
 };
 
 export default CheckSteps;
--- a/src/components/CheckSteps/CheckSteps.tsx
+++ b/src/components/CheckSteps/CheckSteps.tsx
@@ -1,3 +1,4 @@
+import { useProfileRedirect } from '../../hooks/useProfileRedirect.ts';
 import ButtonLink from '../ButtonLink/ButtonLink.tsx';
 import Header from '../Header/Header.tsx';
 import { CheckConductCodeIcon } from '../Icons/CheckConductCodeIcon.tsx';
@@ -5,27 +6,35 @@ import { CompleteProfileIcon } from '../Icons/CompleteProfileIcon.tsx';
 import StepBar from '../StepBar/StepBar.tsx';
 import './CheckSteps.css';
 
-const CheckSteps = () => (
-	<div id='check-steps-wrapper'>
-		<StepBar currentStep={1} steps={[{ label: 'Account confirmed' }, { label: 'Check the conduct code' }, { label: 'Complete your profile' }]} />
-		<div id='check-steps'>
-			<Header isLogoLarge={false}>Welcome to TorontoJS Community Hub!</Header>
-			<p className='intro'>Your account is active, and you're almost ready to get in. Let's complete a few more steps:</p>
-			<ul className='cards'>
-				<li className='card'>
-					<CheckConductCodeIcon />
-					<p className='text-h6'>Check the TorontoJS's conduct code</p>
-				</li>
-				<li className='card'>
-					<CompleteProfileIcon />
-					<p className='text-h6'>Complete your profile</p>
-				</li>
-			</ul>
-			<ButtonLink id='continue-button' isPrimary isLarge href='/pages/review-conduct-code/'>
-				Let's continue
-			</ButtonLink>
-		</div>
-	</div>
-);
+const CheckSteps = () => {
+	const { redirectionComplete } = useProfileRedirect();
+
+	if (redirectionComplete) {
+		return (
+			<div id='check-steps-wrapper'>
+				<StepBar currentStep={1} steps={[{ label: 'Account confirmed' }, { label: 'Check the conduct code' }, { label: 'Complete your profile' }]} />
+				<div id='check-steps'>
+					<Header isLogoLarge={false}>Welcome to TorontoJS Community Hub!</Header>
+					<p className='intro'>Your account is active, and you're almost ready to get in. Let's complete a few more steps:</p>
+					<ul className='cards'>
+						<li className='card'>
+							<CheckConductCodeIcon />
+							<p className='text-h6'>Check the TorontoJS's conduct code</p>
+						</li>
+						<li className='card'>
+							<CompleteProfileIcon />
+							<p className='text-h6'>Complete your profile</p>
+						</li>
+					</ul>
+					<ButtonLink id='continue-button' isPrimary isLarge href='/pages/review-conduct-code/'>
+						Let's continue
+					</ButtonLink>
+				</div>
+			</div>
+		);
+	}
+
+	return <div>loading...</div>;
+};
 
 export default CheckSteps;
--- a/src/components/CompleteProfile/CompleteProfile.tsx
+++ b/src/components/CompleteProfile/CompleteProfile.tsx
@@ -101,7 +101,7 @@ const updateProfile = async (data: UpdateProfileParams, profileId: string) => {
 
 // eslint-disable-next-line max-lines-per-function
 const CompleteProfile = () => {
-	const { redirectionComplete } = useProfileRedirect();
+	const { isRedirecting } = useProfileRedirect();
 
 	const fileInputRef = useRef<HTMLInputElement>(null);
 	const uploadPhotoButtonRef = useRef<HTMLButtonElement>(null);
@@ -325,7 +325,7 @@ const CompleteProfile = () => {
 		}
 	}, [birthdayValue]);
 
-	if (!redirectionComplete || isLoading) {
+	if (isRedirecting || isLoading) {
 		return <p>Loading...</p>;
 	}
 
--- a/src/components/CompleteProfile/CompleteProfile.tsx
+++ b/src/components/CompleteProfile/CompleteProfile.tsx
@@ -317,353 +317,353 @@ const CompleteProfile = () => {
 		}
 	}, [birthdayValue]);
 
-	if (redirectionComplete && !isLoading) {
-		return (
-			<>
-				<StepBar
-					currentStep={3}
-					steps={[
-						{ label: 'Account confirmed' },
-						{ label: 'Check the conduct code' },
-						{ label: 'Complete your profile' }
-					]}
-				/>
-				<form onSubmit={handleSubmit} encType='multipart/form-data' id='complete-profile-form'>
-					<h2>Complete your profile</h2>
-
-					<div id='fields-wrapper'>
-						<details open>
-							<summary className='text-h6'>
-								<span>In a nutshell:</span>
-							</summary>
-
-							<div className='details-content-wrapper'>
-								<div className='details-nutshell-grid'>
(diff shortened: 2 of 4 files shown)


src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -316,353 +316,353 @@
...
+						<details open>
+							<summary className='text-h6'>
+								<span>Avatar:</span>
+							</summary>
+
+							<div className='details-content-wrapper'>
+								<div className='details-content-file-upload'>
 									{photoFile && (
+										<picture>
+											<img src={photoFile} />
```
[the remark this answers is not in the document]
manvinderjit: will fix when integrating the form with file upload api.

Created an issue here for this: #239

src/components/CheckSteps/CheckSteps.tsx
```diff
@@ -1,31 +1,40 @@
...
+					</ul>
+					<ButtonLink id='continue-button' isPrimary isLarge href='/pages/review-conduct-code/'>
+						Let's continue
+					</ButtonLink>
+				</div>
+			</div>
+		);
+	}
+
+	return <div>loading...</div>;
```
madcampos: **suggestion:** Instead of doing a big if statement and then returning the fallback. Return the fallback early, it makes the code easier to understand, easier to read and is less work for the interpreter to parse and analyze everything before returning.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -316,353 +317,353 @@ const CompleteProfile = () => {
...
-											<option value='24'>24</option>
-											<option value='25'>25</option>
-											<option value='26'>26</option>
-											<option value='27'>27</option>
-											<option value='28'>28</option>
-											<option value='29'>29</option>
-											<option value='30'>30</option>
-											<option value='31'>31</option>
-										</select>
+	if (redirectionComplete && !isLoading) {
```
madcampos: Same thing here: put the loading first as an early return.

src/hooks/useProfileRedirect.ts
```diff
@@ -27,14 +27,20 @@ const isValidStatus = (status: string): status is ProfileStatus => VALID_STATUSE
 const REDIRECT_PATHS = {
 	signIn: '/pages/sign-in/',
 	checkSteps: '/pages/check-steps/',
+	reviewConductCode: '/pages/review-conduct-code/',
 	completeProfile: '/pages/complete-profile/',
 	home: '/pages/home/'
 };
 
-const getRedirectPathForStatus = (status: ProfileStatus): string => {
+const normalizePath = (path: string) => path.replace(/\/+$/u, '');
```
madcampos: **question:** What is this regex trying to do? Please add a comment exlaining it.

src/hooks/useProfileRedirect.ts
```diff
@@ -87,29 +102,29 @@ export const useProfileRedirect = () => {
...
-		}
-
 		// Create abort controller to fetch abort
 		const controller = new AbortController();
 		const { signal } = controller;
 
 		const redirect = async () => {
 			try {
-				const redirectPath = await getRedirectPath(signal);
+				// Normalize URL path
```
kbventures: As we already have a normalize function. It makes this comment confusing.  This does not normalize the URL.  We are only extracting the path.

src/hooks/useProfileRedirect.ts
```diff
@@ -87,29 +102,29 @@ export const useProfileRedirect = () => {
...
+				const currentPath = new URL(window.location.href).pathname;
+
+				const redirectPath = await getRedirectPath(signal, currentPath);
+
```
kbventures, line 124: Why are we checking for a signal abort here?  The signal is aborted in the callback clean for useEffect, which happens after.

src/hooks/useProfileRedirect.ts
```diff
@@ -87,29 +102,29 @@ export const useProfileRedirect = () => {
...
 					return;
 				}
 
-				// Normalize URL path
-				const currentPath = new URL(window.location.href).pathname;
-
-				// Avoid redirect if already on the correct path
-				if (currentPath !== redirectPath) {
+				// Redirect only if current url path is different from the redirectpath
+				if (currentPath && currentPath !== redirectPath) {
```
kbventures, line 127: currentPath twice?

src/hooks/useProfileRedirect.ts
kbventures: It's noted already,  "Normalizes a path by removing any trailing slashes (/).",  but it is n ot clear that is precisely referring to what the Regex is doing.

src/hooks/useProfileRedirect.ts
```diff
@@ -68,7 +83,7 @@ export const getRedirectPath = async (signal?: AbortSignal): Promise<RedirectPat
 		const data: ProfileStatusResponse = await response.json();
 
 		// Redirects based on profile status or to sign-in if not profile status is returned
-		return (data?.status && isValidStatus(data.status)) ? getRedirectPathForStatus(data.status) : REDIRECT_PATHS.signIn;
+		return (data?.status && isValidStatus(data.status)) ? getRedirectPathForStatus(data.status, currentPath) : REDIRECT_PATHS.signIn;
```
kbventures, line 99: Question: Why do we have an explicit null at line 92?

src/hooks/useProfileRedirect.ts
```diff
@@ -87,29 +102,29 @@ export const useProfileRedirect = () => {
...
 
 		const redirect = async () => {
 			try {
-				const redirectPath = await getRedirectPath(signal);
+				// Normalize URL path
+				const currentPath = new URL(window.location.href).pathname;
+
+				const redirectPath = await getRedirectPath(signal, currentPath);
+
+				if (signal.aborted || hasRedirected.current) {
```
kbventures: Same comment as above.  The signal is only aborted in the cleanup handler for useEffect.

kbventures (changes requested):
Great work.  I left a few comments and questions.

Please introduce new API tests for these functionalities.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -316,8 +325,8 @@ const CompleteProfile = () => {
 		}
 	}, [birthdayValue]);
 
-	if (isLoading) {
-		return <p>Loading...</p>;
+	if (!redirectionComplete || isLoading) {
+		return <div>loading...</div>;
```
bcha92: **Nitpick**: Is using a `<div>` for a simple text here appropriate? Can we add a `<p>` text element here.

If this is just for a placeholder for something else (i.e. Design team ideas), then it's fine as is.

src/components/ReviewConductCode/ReviewConductCode.tsx
```diff
@@ -7,9 +7,60 @@ import HeaderTitle from '../HeaderTitle/HeaderTitle.tsx';
...
+import { useProfileRedirect } from '../../hooks/useProfileRedirect.ts';
+
+const docTypes = ['code-of-conduct', 'image-release-form', 'volunteer-agreement'];
 
 const ReviewConductCode = () => {
+	const { redirectionComplete } = useProfileRedirect();
+
 	const [agreementChecked, setAgreementChecked] = useState(false);
+	const [isSubmitting, setIsSubmitting] = useState(false);
+	// TODO: Display success and error messages
```
bcha92, line 19: **Nitpick**: Is this still a work in progress or something we can kick this down the road (create new issue)?

src/components/ReviewConductCode/ReviewConductCode.tsx
```diff
@@ -7,9 +7,60 @@ import HeaderTitle from '../HeaderTitle/HeaderTitle.tsx';
...
+				);
+
+				const results = await Promise.all(responses.map(async (res) => res.json()));
+				const allSuccessful = responses.every((res) => res.ok);
+
+				if (allSuccessful) {
+					console.log('All documents signed:', results);
+					window.location.href = '/pages/complete-profile';
+				} else {
+					setSubmitError('Some documents failed to sign. Please try again!');
```
bcha92, line 50: **Non-Blocker:** Should organizers be notified if volunteers fail to sign, or "decline" to sign the Conduct Code of review?

src/hooks/useProfileRedirect.ts
bcha92: ^same as Marco above: comment here should be `// Extract URL path`

bcha92 (commented):
Looks good, but given we now have some clarity (from our meeting today) on how to move forward with the `signal.abort`, we can wait for the changes before it looks good to go.

src/hooks/useProfileRedirect.ts
madcampos: As a suggestion: set the regex to a constant and comment that constant.

Please note though that [global regexes update the `lastIndex` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/global#:~:text=Each%20call%20to%20exec()%20will%20update%20its%20lastIndex%20property%2C%20so%20that%20the%20next%20call%20to%20exec()%20will%20start%20at%20the%20next%20character.) and start from that, so you need to clone the regex if you are declaring it in a global scope but using it multiple times in a local scope.

E.g.:
```typescript
const GLOBAL_REGEX = /a/gu; // <- Regex is global!

function wrongMatch() {
  console.log(GLOBAL_REGEX.exec('aa')); // <- Will match correctly.
  console.log(GLOBAL_REGEX.exec('ab')); // <- Will not match correctly due to lastIndex being updated.
}

function correctMatch() {
  console.log(new RegExp(GLOBAL_REGEX).exec('aa')); // <- Will match correctly.
  console.log(new RegExp(GLOBAL_REGEX).exec('ab')); // <- Will also match correctly because the regex was cloned.
}
```

This can get confusing and annoying, so as a rule of thumb, always clone a regex.

src/hooks/useProfileRedirect.ts
```diff
@@ -51,7 +66,7 @@ const getRedirectPathForStatus = (status: ProfileStatus): string => {
 	}
 };
 
-export const getRedirectPath = async (signal?: AbortSignal): Promise<RedirectPathResult> => {
+export const getRedirectPath = async (signal?: AbortSignal, currentPath?: string): Promise<RedirectPathResult> => {
```
madcampos, line 82: **nitpick**: The order of the parameters should be reversed for better UX.

Imagine the most common scenario for using this function, which parameter will be most probable to be passed, the path or the abort signal?

src/hooks/useProfileRedirect.ts
madcampos: On that note, prefer:
- Not returning anything, as functions by default return `undefined`;
- An empty return to stop function execution (`return;`), it will also implicitly return `undefined`;
- Explicitly return `udnefined` if that is what you are looking for.

Avoid using `null` as it is a weirder value, if you can, always use `undefined`.

src/hooks/useProfileRedirect.ts
madcampos: I think the abort signal is for canceling the promise.

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal

kbventures: @manvinderjit  The abort signal can be triggered in several ways, but in all cases, it must be explicitly called in your code. AbortSignal doesn't abort automatically.

If the only abort signal in your component comes from the useEffect cleanup function, then you should remove the abort signal checks.

src/components/ReviewConductCode/ReviewConductCode.tsx
kbventures: @caducarvalho is working on this, and yes, having a Github issue link
 here would be great.

src/components/ReviewConductCode/ReviewConductCode.tsx
kbventures: @bcha92 

It's not optional.

image

src/hooks/useProfileRedirect.ts
manvinderjit: @kbventures We abort the signal using this:
`controller.abort();`

`signal.aborted `checks if that particular signal was aborted. It only returns true or false. It doesn't abort the signal => it is not a function.

signal.abort is used like an `isComponentMounted` state

src/hooks/useProfileRedirect.ts
manvinderjit: https://github.com/torontojs/communityhub/pull/238#discussion_r2308810239

src/components/ReviewConductCode/ReviewConductCode.tsx
bcha92: @kbventures my apologies for the lack of clarity. I actually meant would there be a fallback mechanism (or time limit) if the volunteer did not take any action (basically "do nothing") to consent to the Code of Conduct within a certain time period after registration?

Or does this fall under the wider scope of "registered but not confirmed" users, similar conditions if a user did not verify their email address after registration?

manvinderjit: @bcha92 Operationally, the user will be redirected to this page every time the sign in, if they don't accept. 

There was a discussion on imposing a limit and deactivating the accounts if they don't accept CoC, after 90 or 180 days, but nothing is concrete yet. @kbventures Might have more on this.

src/hooks/useProfileRedirect.ts
manvinderjit: @madcampos The currentPath is only passed to a sub-function `getRedirectPathForStatus` where it is used to determine between `checkSteps` and `Review Conduct Code` as both require the same status 'activated'. Otherwise, Review Conduct Code kept on redirecting to CheckSteps, even though it required the same status. A quick fix made more sense here.

Signal is going to be used for each fetch api request in the function itself, whereas currentPath is just passed down for an odd case.

src/hooks/useProfileRedirect.ts
manvinderjit: Technically, it extracts the pathname, removing the domain and special chars like `.` & `..` , `%x` chars. 

But I will simplify it for clarity.

src/hooks/useProfileRedirect.ts
manvinderjit: I have updated the same to throw an error now.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: We do need a loading component from the design team, but I will still fix this. Good catch.

src/components/ReviewConductCode/ReviewConductCode.tsx
manvinderjit: Created. https://github.com/torontojs/communityhub/issues/252

src/hooks/useProfileRedirect.ts
manvinderjit: @kbventures We are not checking for abort signal in line 114. We are passing the signal to the function so that it can be associated with the fetch api. 

image

On a side note, I have reverted my hook to an older version. It had a bug, the community helped me figure it out.

src/hooks/useProfileRedirect.ts
manvinderjit: I have reverted my hook to an older version. It had a bug, the community helped me figure it out.

src/hooks/useProfileRedirect.ts
manvinderjit: @madcampos Why not just declare a non-global regex? Is there any advantage of using the 'g' flag?

kbventures (dismissed):
Please fix this before finishing this pr. 

https://github.com/torontojs/communityhub/issues/256

manvinderjit:
> Please fix this before finishing this pr.
> 
> #256

We postponed this for now, as per the last meeting.

src/hooks/useProfileRedirect.ts
manvinderjit: Updated based on suggestions!

src/components/ReviewConductCode/ReviewConductCode.tsx
manvinderjit: Resolving this for now. We can bring this up in a future developer meeting.

src/hooks/useProfileRedirect.ts
manvinderjit: The second parameter is optional, so the lint rule won't allow it to be placed first.

kbventures (approved):

Issue #65, opened 2025-05-08
"Your account is ready" Page


madcampos:

madcampos:
@will-frankland how is this issue going? Do you need any help with that?

kbventures:
<img width="487" height="347" alt="Image" src="https://github.com/user-attachments/assets/5b39bf99-b6a8-41d8-a09f-87c1b7ce8d94" />

kbventures:
For the time being, successfully changing your password will forward you to the sign-in page.

PR #249, opened 2025-08-27
Account Ready Page (issue #65)

roshan-budhathoki:
Creation of your account ready page. 
The link for the page is placed on the index URL. When we click on the account ready link, it will redirect to "/pages/account-ready", which will show the whole UI of the page.

### DIFF
--- a/src/components/AccountReady/AccountReady.css
+++ b/src/components/AccountReady/AccountReady.css
@@ -5,12 +5,11 @@
 	flex-direction: column;
 	max-width: 390px;
 	width: 100%;
+	gap: var(--spacing-sm) 0;
 	padding: var(--spacing-sm);
 
 	p {
 		font-size: var(--text-sm);
-		margin: 0 var(--spacing-sm);
-		padding-bottom: var(--spacing-sm);
 		text-align: center;
 	}
 
--- a/src/components/AccountReady/AccountReady.css
+++ b/src/components/AccountReady/AccountReady.css
@@ -16,7 +16,7 @@
 	hr {
 		border: none;
 		border-top: var(--border);
-		margin: 1px 0;
+		margin: 0.063rem 0;
 	}
 
 	a {
--- a/src/components/AccountReady/AccountReady.css
+++ b/src/components/AccountReady/AccountReady.css
@@ -1,0 +1,28 @@
+#account-ready-form {
+	border: var(--border);
+	border-radius: var(--rounded-md);
+	display: flex;
+	flex-direction: column;
+	max-width: 390px;
+	width: 100%;
+	padding: var(--spacing-sm);
+
+	p {
+		font-size: var(--text-sm);
+		margin: 0 var(--spacing-sm);
+		padding-bottom: var(--spacing-sm);
+		text-align: center;
+	}
+
+	hr {
+		border: none;
+		border-top: var(--border);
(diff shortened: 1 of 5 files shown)


src/components/AccountReady/AccountReady.css
```diff
@@ -0,0 +1,28 @@
+#account-ready-form {
```
kbventures, line 1: Please utilize flex "gap" to achieve consistent spacing on your page.  

Have a look at our other design implementation. 

image

src/components/AccountReady/AccountReady.css
```diff
@@ -0,0 +1,27 @@
...
+	border: var(--border);
+	border-radius: var(--rounded-md);
+	display: flex;
+	flex-direction: column;
+	max-width: 390px;
```
manvinderjit: Please use `rem` instead of `px` for units. 

`rem is often better than px because it provides superior accessibility, enabling users to adjust font sizes in their browser settings`

Here is a calculator that does the same https://nekocalc.com/px-to-rem-converter

src/pages/account-ready/account-ready.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const root = document.getElementById('root') as HTMLDivElement;
+
+createRoot(root).render(
+	(
+		<StrictMode>
+			<div className='app'>
+				<div className='page'>
+					<Header>Your account is ready</Header>
+					<AccountReadyForm />
```
manvinderjit: **suggestion**: the dominant content of the body should ideally be surrounded by a main section. Use either in the page or the component itself.
```
<Header />
<main>
    ...
</main>
<Footer />
```
**Suggested Reading**:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/main

manvinderjit (changes requested):
Good start @roshan-budhathoki and fantastic turnaround time. Please look into the suggested changes.

src/components/AccountReady/AccountReady.css
```diff
@@ -0,0 +1,27 @@
...
+
+	p {
+		font-size: var(--text-sm);
+		text-align: center;
+	}
+
+	hr {
+		border: none;
+		border-top: var(--border);
+		margin: 1px 0;
```
bcha92: also here, with the exception of the `0`, please do not use pixels

bcha92 (changes requested):
more or less things that are already covered by @manvinderjit regarding use of flexbox instead of fixed pixels and proper html component hierarchies.

src/components/AccountReady/AccountReady.css
kbventures: @roshan-budhathoki 

https://github.com/torontojs/communityhub/issues/251

src/pages/account-ready/account-ready.tsx
kbventures: @manvinderjit I'm not sure if this is applied throughout the project(Create and issue?) and whether main tag should live within the actual component in this case AccountReadyForm.

bcha92 (approved):
Looks good

kbventures (approved):

manvinderjit (approved):

manvinderjit:
@roshan-budhathoki Great work.

Issue #261, opened 2025-09-10
[SignUp - Refactor With Reusbale Component] replace text input with the design system's text input component


manvinderjit:
### Description

For the SignUp page, replace the following text input and label with the the [TextInputComponent](https://github.com/torontojs/communityhub/blob/dev/src/components/TextInputComponent/TextInputComponent.tsx) of the design system.

<img width="494" height="126" alt="Image" src="https://github.com/user-attachments/assets/e20e881a-e4fa-4bbb-8dc3-1778149b58b7" />

1. The TextInputComponent provides both label and input. You will be replacing both of them.
2. Don't forget to remove redundant CSS from the respective css file.

Please refer to the usage guide of the component [here](https://github.com/torontojs/communityhub/blob/dev/src/pages/text-input-usage/sample-text-inputs.tsx)

manvinderjit:
<img width="494" height="126" alt="Image" src="https://github.com/user-attachments/assets/0b8ad42d-9845-4c04-
a539-2ec306915b80" />

PR #263, opened 2025-09-11
Text input component added and redundant css removed #261

roshan-budhathoki:
1. Addition Of Text Input Component in Sign Up Form
2. Removing redundant button CSS

It implements the reusable input component in the Sign Up form.


src/components/SignUp/SignUpForm.tsx
```diff
@@ -95,20 +96,20 @@ const SignUpForm = (): React.JSX.Element => {
...
+							<span>Please enter a valid name that is at least 1 character long.</span>
+						</div>
+					)}
 					ref={nameInputRef}
+					minLength={1}
 				/>
-				<div id='name-input-helper-text'>
-					<span>Please enter a valid name that is at least 1 character long.</span>
-				</div>
 			</div>
```
manvinderjit, line 116: @roshan-budhathoki Good work on the task itself. 

I noticed there is a spacing issue between the password strength meter and the message. Please see if you can find the root cause and fix it. Thanks

image

Design
image

src/components/SignUp/SignUpForm.tsx
```diff
@@ -95,20 +96,20 @@ const SignUpForm = (): React.JSX.Element => {
...
+					id='name-input'
 					placeholder='Your name'
-					minLength={1}
 					required
-					aria-describedby='name-input-helper-text'
+					helperMessageSlot={(
+						<div id='name-input-helper-text'>
+							<span>Please enter a valid name that is at least 1 character long.</span>
+						</div>
+					)}
```
manvinderjit, line 111: For the helper message, we have a [HelperMessageComponent](https://github.com/torontojs/communityhub/blob/dev/src/pages/usage-helper-message/sample-helper-message.tsx) that can be passed to the text component like so
```suggestion
					helperMessageSlot={<HelperMessageComponent variant='error' labelText='Please enter a valid name that is at least 1 character long.' />}
```

manvinderjit: For the css, we can target the helper component using this approach

In the css file, we can modify the lines 110, 111 as such
From 
```
#name-input-helper-text { display: none; }
#name-input:not(:focus):user-invalid ~ #name-input-helper-text { display: block; }
```

To:
```
#name-input ~ .helper-container { display: none; }
#name-input:user-invalid:not(:focus) ~ .helper-container { display: flex; }
```

manvinderjit: What this does is, instead of directly taking the helper container id, we target the `id` of the `text input ` and selects any sibling elements with the class `helper-container` with respect to the text input.

manvinderjit: Alternatively, we can modify the HelperMessageComponent to include an `id` param and then use it to target the element directly.  In this case we won't need to modify the existing css.

Which would be a more precise approach. I would recommend this one over the other.

manvinderjit: @roshan-budhathoki This was not a part of the task itself. I will let you decided if you want to handle it here or create a separate issue for it.

roshan-budhathoki: I will handle it here @manvinderjit. It would be easier for me.

roshan-budhathoki: > Alternatively, we can modify the HelperMessageComponent to include an `id` param and then use it to target the element directly. In this case we won't need to modify the existing css.

"Helper Message Component" is used on multiple places. So, I would not touch it at the moment but will keep your thought in consideration.

manvinderjit: > > Alternatively, we can modify the HelperMessageComponent to include an `id` param and then use it to target the element directly. In this case we won't need to modify the existing css.
> 
> "Helper Message Component" is used on multiple places. So, I would not touch it at the moment but will keep your thought in consideration.

@roshan-budhathoki  Please create a GitHub issue for this.

src/components/SignUp/SignUpForm.tsx
```diff
@@ -95,20 +97,21 @@ const SignUpForm = (): React.JSX.Element => {
...
+							labelText='Please enter a valid name that is at least 1 character long.'
+						/>
+					)}
 					ref={nameInputRef}
+					minLength={1}
 				/>
-				<div id='name-input-helper-text'>
-					<span>Please enter a valid name that is at least 1 character long.</span>
-				</div>
 			</div>
```
kbventures, line 116: I know you're not responsible for this error,but I think this is a good opportunity to fix it. 

image

src/components/SignUp/SignUpForm.tsx
```diff
@@ -2,6 +2,8 @@ import './SignUpForm.css';
 import { useRef, useState } from 'react';
```
kbventures, line 2: Please implement the TextInputComponent for all three inputs for the page, as well as their respective helper components.

image

src/components/SignUp/SignUpForm.css
```diff
@@ -107,13 +107,8 @@
 	}
 }
 
-#name-input-helper-text { display: none; }
-#name-input:not(:focus):user-invalid ~ #name-input-helper-text { display: block; }
+#name-input ~ .helper-container { display: none; }
+#name-input:user-invalid:not(:focus) ~ .helper-container { display: flex; }
```
kbventures, line 111: Question:

Why are we switching not(:focus):user-invalid with user-invalid:not(:focus) ?

kbventures (changes requested):
Thank you for your contribution.  I left some nodes that need to be actioned.

src/components/SignUp/SignUpForm.css
manvinderjit: @kbventures For using the HelperMessageComponent instead of directly passing an element.

src/components/SignUp/SignUpForm.tsx
manvinderjit: @kbventures The second and the third are of the type email and password, not text.

We will need separate components for these unless we want to club all three.

src/components/SignUp/SignUpForm.tsx
roshan-budhathoki: Sure, if I can work on the Email and Password component. Then it will be resolved. I am waiting for confirmation from @manvinderjit.

manvinderjit:
@roshan-budhathoki nitpick: please use conventional commit format for the commit messages. The voice used should be active voice and it should be imperative and not use past tense.

The best way to write a commit message is this format

_when you apply this commit, it will "fix: resolve the spacing issue between password bar and error msg"_

src/components/SignUp/SignUpForm.tsx
kbventures: @roshan-budhathoki, that's a good point.  Take @manvinderjit 's lead on this.  I might be causing chaos here.   Ultimately, we want this to be fixed before the beta launch.  Please discuss the solution together.

manvinderjit: @roshan-budhathoki Skip the email and password for now. We need to determine if we need to create separate components for these or modify the text component so that it can accommodate the other two as well.

bcha92 (approved):
I think this is good to go

kbventures:
Re assigned @s1m0n1au

Saif-XI-Coderz:
@kbventures PR branch updated and ready to be approved.

PR #269, opened 2025-09-25
Design: add label component with design usage

roshan-budhathoki:
A Label component is created for the design system and implemented in its usage page.


src/pages/label-usage/sample-label.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const root = document.getElementById('root') as HTMLDivElement;
+
+createRoot(root).render(
+	(
+		<StrictMode>
+			<div className='App parent'>
+				<h2>Label Usage</h2>
+				<Label>Default</Label>
+				<Label isPrimary>Primary</Label>
```
bcha92: Given the Label property is shown in the below image as a nested element within the `<input></input>` form label element, I think we need to add an interactive potion in this page.. OR

modify the existing `/pages/text-input-usage/` design page to include this feature to match the Figma design as intended.

Mobile:
Screenshot 2025-10-06 222730

Desktop:
Screenshot 2025-10-06 222654

src/components/Label/Label.tsx
```diff
@@ -0,0 +1,18 @@
...
+interface Props extends ParamHTMLAttributes<HTMLParagraphElement> {
+	isPrimary?: boolean;
+}
+
+const Label = forwardRef<HTMLParagraphElement, Props>(({
+	isPrimary = false,
+	children,
+	...rest
+}, ref) => (
+	<p ref={ref} className='label' {...rest} data-primary={isPrimary}>
```
bcha92: If the `Label` reusable component is for inner text usage within the input element for Profile page like "Skills", I'm not sure if nesting a `<p>` within a `<input>` element may be appropriate, but I could be wrong about this... I would normally use something like `<span>` to modify existing text instead but I think this part may need further discussion and consultation as I'm not quite sure how to approach something like this...

Also, we may need to consider if accentuating/emphasizing each "Label" component for accessibility considerations is required in this case (i.e. additional aria labels for each listed skill)

any thoughts @manvinderjit @kbventures ?

bcha92 (changes requested):
Great start! Under normal circumstances, I would pass this for approval, but I think we need a few changes to make sure your reusable component also has an example use case that is similar to the approved designs by our design team (see comment with images) and if using the paragraph `<p>` element is appropriate for such case (subject to further discussion by the team).

Current submitted design (as seen on local dev env):
Screenshot 2025-10-06 222908

src/components/Label/Label.tsx
roshan-budhathoki: Thank you for the review @bcha92 . You are right, but let's wait for confirmation. I never thought about its usecase at multiple places.

manvinderjit: @roshan-budhathoki 

@bcha92 is correct here. We should use `<span />` here because it will be compatible with html elements like `<input />` and `<label />`.
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/span

The p tag is generally used for large text. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/p

manvinderjit: @roshan-budhathoki Also, note that the skills in the Complete Profile page have a close button at the end. You may need to extend the component to optionally allow this.

Check the code for Complete Profile to see how it is currently implemented.

image

roshan-budhathoki: yes, I have implemented optional button props for a button on start based on figma design. Also, the design here on skill section does not match with figma one. I am a little bit confused here.

@manvinderjit @bcha92

manvinderjit: @roshan-budhathoki It is because this component was created before the label design token was created. And Marco, our former manager, asked to add a close button. Since the Figma design has now been updated, the component here uses the older version, while the figma design has the newer version.

manvinderjit: @roshan-budhathoki I have created a thread on this in slack. 

Always reach out to Cadu or anyone else from the design team, for clarification on any design changes.

https://torontojs.slack.com/archives/C09DXP7RM1V/p1760205278525729

src/pages/label-usage/style.css
```diff
@@ -0,0 +1,18 @@
...
+	gap: var(--spacing-xs);
+}
+
+.label-collection {
+	display: flex;
+	gap: var(--spacing-lg);
+	align-items: center;
+
+	p {
+		font-weight: bold;
```
kbventures: We have bold varibles that should be utilized here enabling us to make global changes to the project.

src/components/Label/Label.tsx
```diff
@@ -0,0 +1,48 @@
...
+type LabelComponentTones = 'dark' | 'light';
+
+interface Props extends ParamHTMLAttributes<HTMLSpanElement> {
+	variant?: LabelColor;
+	tone?: LabelComponentTones;
+	hasIcon?: boolean;
+	icon?: React.JSX.Element;
+}
+
+const Label = forwardRef<HTMLSpanElement, Props>(({
```
kbventures: WHEN will be need a forwardRef currently in our front end setup here?  An example?

kbventures (changes requested):
Great work @roshan-budhathoki 

This is almost ready to be merged.  

Left one question and a simple fix to be made.

src/components/Label/Label.tsx
bcha92: +1; I think using forwardRef is over-engineering this a bit. A simple React functional component would be sufficient in this case.

src/pages/label-usage/style.css
bcha92: Looks good. Resolving this.

src/components/Label/Label.tsx
bcha92: forwardRef is fixed. Resolving this.

bcha92:
some cosmetic changes made to the page

bcha92 (dismissed):
approved

bcha92:
Closing PR as stale. Label usage is apparently already merged in to dev via another PR.

PR #271, opened 2025-09-30
262 complete profile refactor with reusable text input component

mahsakarami-dev:
Summary: 

1. Revert the TextInputComponent to support only the text type.
2. checkbox and email input fields remain direct inputs, to allow for future refactoring into reusable components.
3. url inputs for GitHub and LinkedIn fields changed to text type. The url input type doesn’t accept account names (e.g., @accountName or accountName) correctly. Switching to text allows validation for both URLs and plain account names.

### DIFF
--- a/src/components/CompleteProfile/CompleteProfile.tsx
+++ b/src/components/CompleteProfile/CompleteProfile.tsx
@@ -359,6 +359,7 @@ const CompleteProfile = () => {
 									label='Name'
 									value={profileData.name}
 									disabled={true}
+									readOnly={true}
 								/>
 								<div>
 									<label htmlFor='email' className='input-required'>
@@ -380,9 +381,7 @@ const CompleteProfile = () => {
 									name='slack'
 									label='Slack handle'
 									placeholder='Your slack handle to TorontoJS'
-									value={profileData.email}
 									onBlur={validateSlackHandle}
-									disabled={true}
 									required
 								/>
 								<TextInputComponent
@@ -541,10 +540,12 @@ const CompleteProfile = () => {
 								</div>
 							</div>
 							{!photoFile && (
-								<HelperMessageComponent
-									variant='info'
-									labelText="Please, ensure you're uploading a JPEG or PNG photo."
-								/>
+								<div className='file-upload-helper-message-container'>
+									<HelperMessageComponent
+										variant='info'
+										labelText="Please, ensure you're uploading a JPEG or PNG photo."
+									/>
+								</div>
 							)}
 						</div>
 					</details>
@@ -562,6 +563,7 @@ const CompleteProfile = () => {
 										ref={linkedinInputRef}
 										name='linkedin'
 										label='LinkedIn profile'
+										placeholder='Link to your LinkedIn profile'
 									/>
 								</div>
 								<div>
@@ -570,15 +572,17 @@ const CompleteProfile = () => {
 										ref={githubInputRef}
 										name='github'
 										label='GitHub profile'
+										placeholder='Link to your GitHub profile'
 									/>
 								</div>
-								<div>
+								<div className='details-information-grid'>
 									<label htmlFor='portfolio'>Site/portfolio</label>
 									<input
 										ref={sitePortfolioInputRef}
 										id='portfolio'
 										name='portfolio'
 										type='url'
+										placeholder='Link to your personal site'
 										className='text-input'
 									/>
 								</div>
--- a/src/components/CompleteProfile/CompleteProfile.tsx
+++ b/src/components/CompleteProfile/CompleteProfile.tsx
@@ -382,7 +382,6 @@ const CompleteProfile = () => {
 									label='Slack handle'
 									placeholder='Your slack handle to TorontoJS'
 									onBlur={validateSlackHandle}
-									required
 								/>
 								<TextInputComponent
 									id='pronouns'
--- a/src/components/CompleteProfile/CompleteProfile.tsx
+++ b/src/components/CompleteProfile/CompleteProfile.tsx
@@ -360,15 +360,20 @@ const CompleteProfile = () => {
 									value={profileData.name}
 									disabled={true}
 								/>
-								<TextInputComponent
-									id='email'
-									name='email'
-									label='E-mail'
-									type='email'
-									value={profileData.email}
-									disabled={true}
-									required
-								/>
+								<div>
+									<label htmlFor='email' className='input-required'>
+										E-mail
+									</label>
+									<input
+										className='text-input'
+										id='email'
+										name='email'
+										type='email'
+										value={profileData.email}
+										readOnly={true}
+										required
+									/>
+								</div>
 								<TextInputComponent
 									id='slack'
 									ref={slackHandleInputRef}
@@ -385,9 +390,13 @@ const CompleteProfile = () => {
 									name='pronouns'
 									label='Pronouns'
 									placeholder='Your pronouns (optional)'
-									listId='pronouns-options'
-									datalistOptions={['He/him', 'She/her', 'They/them']}
+									list='pronouns-options'
 								/>
+								<datalist id='pronouns-options'>
+									<option>He/him</option>
+									<option>She/her</option>
+									<option>They/them</option>
+								</datalist>
 								<div>
 									<label>Date of birth</label>
 									<div className='dob-wrapper'>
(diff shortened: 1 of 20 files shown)


src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -376,40 +374,31 @@ const CompleteProfile = () => {
...
+									id='pronouns'
+									name='pronouns'
+									label='Pronouns'
+									placeholder='Your pronouns (optional)'
+									list='pronouns-options'
+								/>
+								<datalist id='pronouns-options'>
+									<option>He/him</option>
+									<option>She/her</option>
+									<option>They/them</option>
```
bcha92, line 396: I think this is good for the time being but should we add an "optional" stretch goal here regarding pronouns (add your own preferred pronoun)?

EDIT: created issue here: #273 

@kbventures @manvinderjit

bcha92 (dismissed):
I don't see any other issue other than that potentially optional stretch goal. So I will approve.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -376,40 +374,31 @@ const CompleteProfile = () => {
...
-										<option>She/her</option>
-										<option>They/them</option>
-									</datalist>
-								</div>
+								<TextInputComponent
+									id='slack'
+									ref={slackHandleInputRef}
+									name='slack'
+									label='Slack handle'
+									placeholder='Your slack handle to TorontoJS'
```
robester0403, line 383: Is this meant to be the actual email or os slack handle, the name you give yourself. If it's meant to be email I think you need a different placeholder text or clarification that it can be handle or slack email.

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -376,40 +374,31 @@ const CompleteProfile = () => {
...
+									value={profileData.email}
+									onBlur={validateSlackHandle}
+									disabled={true}
```
robester0403: There's a hardcoded disable here. Is it meant to be like this?

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -376,40 +374,31 @@ const CompleteProfile = () => {
...
-									</datalist>
-								</div>
+								<TextInputComponent
+									id='slack'
+									ref={slackHandleInputRef}
+									name='slack'
+									label='Slack handle'
+									placeholder='Your slack handle to TorontoJS'
+									value={profileData.email}
+									onBlur={validateSlackHandle}
```
robester0403, line 384: Two comments here:
1. Validating onBlur is good but here is a chance that you can change this input and directly press the submit button, causing the error to never be shown to the user.
2. The validate slack handle only tests if the input is empty or not, so if you are submitting an email, you need to grab some regex to validate email (just google this) and then notify the user of the type of error (is it empty or is it not a valid email)

robester0403 (commented):
Some issues with the Text Input Component.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @bcha92 `datalist` allows users to add their custom text in addition to the available options. This functionality is missing in the `select` tag, which is not being used here.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: @robester0403 It is meant to be the slack handle in the TorontoJS slack.

The email is separate and requested when the user creates an account.

robester0403: In line 383 the value is set to `profileData.email`

manvinderjit: @mkmi7171 Slack input will have the TorontoJS slack handle. We already have an email field before. 

Also, it is an uncontrolled component. It will not have a value field.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: 1. The validation is also called in the submitHandler on line 252 for this case.

2. The input accepts a slack handle and not email. I have left feedback for this to Mahsa above.

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: Good catch. It shouldn't be.

@mkmi7171 Only the name and email fields are to be readonly. Not slack.

src/components/HelperMessageComponent/HelperMessageComponent.css
```diff
@@ -5,6 +5,7 @@
 		gap: var(--spacing-xs);
 		height: auto;
 		width: 100%;
+		justify-content: center;
```
manvinderjit: question: Justification in the component should not limit if the parent wants to justify it to left/start or right/end.

src/components/TextInputComponent/TextInputComponent.tsx
```diff
@@ -1,12 +1,14 @@
 import { forwardRef, type InputHTMLAttributes, type KeyboardEventHandler, useMemo } from 'react';
 import './TextInputComponent.css';
-
 interface Props extends InputHTMLAttributes<HTMLInputElement> {
-	label: string;
+	label?: string;
```
manvinderjit, line 4: fix: As per the WCGA standard, labels are compulsory for inputs like text, email, password, etc.
```suggestion
	label: string;
```

src/components/CompleteProfile/CompleteProfile.tsx
kbventures: @bcha92 we're not exploring optional pronouns for now.  I've responded and closed that issue.

kbventures (changes requested):
I've highlighted anomalies compared to the design :

image

Origin design: 
image

src/components/CompleteProfile/CompleteProfile.tsx
manvinderjit: **issue**: @mkmi7171 Please use the `readOnly` input type for `name` and `email`. 

image

Refer to the design usage system file accessible at localhost:3000
image

Refer to the following component in it
Screenshot from 2025-10-16 22-22-35

manvinderjit: issue: Slack handle should not be disabled. Please make it a regular text input component.

image

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -551,6 +540,12 @@ const CompleteProfile = () => {
 									)}
 								</div>
 							</div>
+							{!photoFile && (
```
manvinderjit, line 541: **issue**: Image is not being contained. Please double check the css.

Screenshot from 2025-10-16 22-21-12

manvinderjit (changes requested):

manvinderjit:
> I've highlighted anomalies compared to the design :
> image
> 
> Origin design: image

@mkmi7171 For this issue, in CompleteProfile.css please make the following changes 
image

src/components/CompleteProfile/CompleteProfile.tsx
mahsakarami-dev: I tested it, and every image I tried appears contained.
Here’s a screenshot for reference
Screenshot 2025-10-23 at 1 47 33 PM

src/components/CompleteProfile/CompleteProfile.tsx
```diff
@@ -351,17 +353,14 @@ const CompleteProfile = () => {
 						<div className='details-content-wrapper'>
 							<div className='details-nutshell-grid'>
 								{/* TODO: Create resuable input components later */}
```
bcha92, line 355: Maybe remove this comment?

src/components/TextInputComponent/TextInputComponent.tsx
```diff
@@ -60,7 +63,7 @@ const TextInputComponent = forwardRef<HTMLInputElement, Props>(({
 				ref={ref}
 				id={elementId}
 				name={name}
-				type='text'
+				type={'text'}
```
bcha92, line 66: Nice catch there!

bcha92 (commented):
Looks good. Maybe some cleanup is in order.

src/components/CompleteProfile/CompleteProfile.tsx
kbventures: @mkmi7171 have you tried images with alternative dimensions? I"m getting different results depending on the test images I am utlizing.

kbventures:
Some more knit picking here but we definitly need a bit more space here. 
image

kbventures:
Re assign @mkmi7171

bcha92:
Closing PR as stale. TextInputComponent already merged in to dev on another PR.

PR #282, opened 2025-10-08
refactor(build): discover pages entrypoints dynamically

### LINKED ISSUE #281: [Feature request]: Discover `src/pages` entrypoints dynamically for `vite.config.js`
### Description

Replace the hard-coded `rollupOptions.input` with filesystem `src/pages`  discovery.

```js
			rollupOptions: {
				input: {
					'index': fileURLToPath(new URL('./src/index.html', import.meta.url)),
					'profile': fileURLToPath(new URL('./src/pages/profile/index.html', import.meta.url)),
					'team': fileURLToPath(new URL('./src/pages/team/index.html', import.meta.url)),
					'complete-profile': fileURLToPath(new URL('./src/pages/complete-profile/index.html', import.meta.url)),
					'check-steps': fileURLToPath(new URL('./src/pages/check-steps/index.html', import.meta.url)),
					'review-conduct-code': fileURLToPath(new URL('./src/pages/review-conduct-code/index.html', import.meta.url)),
					'sign-up': fileURLToPath(new URL('./src/pages/sign-up/index.html', import.meta.url)),
					'sign-in': fileURLToPath(new URL('./src/pages/sign-in/index.html', import.meta.url)),
					'confirm-account': fileURLToPath(new URL('./src/pages/confirm-account/index.html', import.meta.url)),
					'check-your-email': fileURLToPath(new URL('./src/pages/check-your-email/index.html', import.meta.url)),
					'home': fileURLToPath(new URL('./src/pages/home/index.html', import.meta.url)),
					'button-usage': fileURLToPath(new URL('./src/pages/button-usage/index.html', import.meta.url)),
					'print-documents': fileURLToPath(new URL('./src/pages/print-documents/index.html', import.meta.url)),
					'input-usage': fileURLToPath(new URL('./src/pages/text-input-usage/index.html', import.meta.url)),
					'reset-password': fileURLToPath(new URL('./src/pages/reset-password/index.html', import.meta.url)),
					'date-usage': fileURLToPath(new URL('./src/pages/date-usage/index.html', import.meta.url)),
					'helper-message': fileURLToPath(new URL('./src/pages/usage-helper-message/index.html', import.meta.url))
				}
			}
```

### Proposed solution

Create a function to replace the hard-coded `rollupOptions.input` with filesystem `src/pages`  discovery

### Alternative solutions (if any)

_No response_

### Additional context

_No response_

s1m0n1au:
Resolves #281

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,9 +1,6 @@
 /* eslint-env node */
 /* eslint-disable camelcase */
 
-import fs from 'node:fs';
-import path from 'node:path';
-
 import { cloudflare } from '@cloudflare/vite-plugin';
 import react from '@vitejs/plugin-react';
 
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -9,6 +9,8 @@ import { cloudflare } from '@cloudflare/vite-plugin';
 import react from '@vitejs/plugin-react';
 
 import { defineConfig, type UserConfig } from 'vite';
+import { createLogger } from 'vite';
+const logger = createLogger();
 
 export default defineConfig(({ mode }) => {
 	const config: UserConfig = {
@@ -41,16 +43,35 @@ export default defineConfig(({ mode }) => {
 					const inputs: Record<string, string> = {
 						index: path.resolve('src', 'index.html')
 					};
+					logger.info(`✓ Added main entry: ${path.resolve('src', 'index.html')}`);
 
 					const pagesDir = path.resolve('src', 'pages');
-					if (!fs.existsSync(pagesDir)) { return inputs; }
+					if (!fs.existsSync(pagesDir)) {
+						logger.info('ℹ No pages directory found, using main entry only');
+						return inputs;
+					}
 
 					for (const folder of fs.readdirSync(pagesDir)) {
+						if (folder.startsWith('_')) {
+							logger.info(`⊘ Skipped ${folder}: private directory (underscore prefix)`);
+							continue;
+						}
+
 						const folderPath = path.join(pagesDir, folder);
 						const htmlEntry = path.join(folderPath, 'index.html');
 
-						if (fs.statSync(folderPath).isDirectory() && fs.existsSync(htmlEntry)) {
-							inputs[folder] = htmlEntry;
+						try {
+							const stats = fs.statSync(folderPath);
+							if (stats.isDirectory()) {
+								if (fs.existsSync(htmlEntry)) {
+									inputs[folder] = htmlEntry;
+									logger.info(`✓ Added page: ${folder}`);
+								} else {
+									logger.info(`⊘ Skipped ${folder}: no index.html found`);
+								}
+							}
+						} catch (error: any) {
+							logger.warn(`⚠ Error processing ${folder}: ${error.message}`);
 						}
 					}
 
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -21,7 +21,7 @@ export default defineConfig(({ mode }) => {
 			cloudflare({
 				configPath: '../wrangler.toml',
 				persistState: { path: '../.wrangler/state' }
-			}),
+			})
 		],
 		appType: 'mpa',
 		esbuild: { target: 'esnext' },
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -3,22 +3,12 @@
 
 import fs from 'node:fs';
 import path from 'node:path';
-import { fileURLToPath } from 'node:url';
 
 import { cloudflare } from '@cloudflare/vite-plugin';
 import react from '@vitejs/plugin-react';
 
 import { defineConfig, type UserConfig } from 'vite';
-import { createLogger } from 'vite';
-const logger = createLogger();
-
-declare global {
-	var __appInputCache: {
-		inputs: Record<string, string> | null
-	} | undefined;
-}
-
-const inputCache = globalThis.__appInputCache ??= { inputs: null };
+import { virtualMpaPlugin } from './plugins/vite-plugin-virtual-mpa';
 
 export default defineConfig(({ mode }) => {
 	const config: UserConfig = {
@@ -27,7 +17,8 @@ export default defineConfig(({ mode }) => {
 			cloudflare({
 				configPath: '../wrangler.toml',
 				persistState: { path: '../.wrangler/state' }
-			})
+			}),
+			virtualMpaPlugin()
 		],
 		appType: 'mpa',
 		esbuild: { target: 'esnext' },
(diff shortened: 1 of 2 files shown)


kbventures:
@s1m0n1au 

Thanks for contributing.  Please elaborate on your rationale for moving from explicit page build vs dynamic page build, and the pros and cons of each, which led you to think we should move in that direction? I haven't formed any opinions yet appreciate reading yours on this topic while I do some research.

PROS: Not having to add them manuall anymore would save development time.

kbventures:
@s1m0n1au does Vite have opinions/recommendations,  that are good / practices in regards to how to set up this specific file?

At first glance, it removes the annoyance of having to add the files explicitly, but at the cost of control, potentially slower build time(this could be marginal though), and if we introduce something weird in the file system, the build can fail, and it might then be harder to find the error than simply looking at the config file. 

To convince me this is a good idea I think the Vite documentation might be the place to obtain the information on whether this is a good idea or not.

s1m0n1au:
> @s1m0n1au 
> 
> 
> 
> Thanks for contributing.  Please elaborate on your rationale for moving from explicit page build vs dynamic page build, and the pros and cons of each, which led you to think we should move in that direction? I haven't formed any opinions yet appreciate reading yours on this topic while I do some research.
> 
> 
> 
> PROS: Not having to add them manuall anymore would save development time. 

Thanks for taking a look! From my vantage point the explicit `rollupOptions.input` list did the job, yet it carried a hidden maintenance tax. **Every new HTML page required a manual config change**, and if I—or anyone else—missed that step, the page simply never made it into the production bundle. That’s exactly how the `account-ready` page slipped through: `npm run dev` picked it up because dev-server crawls the filesystem, but `npm run preview` relied on the hard-coded list, so the page vanished.

“Dynamic” discovery flips that failure mode. Instead of remembering to register each page, I only create `src/pages/{slug}/index.html` and it’s automatically bundled. **The upside is reduced drift between the source tree and the build manifest, plus fewer review steps for new pages.** The trade-offs are mostly around visibility: with an explicit list I can see the full routing surface in one spot, while the dynamic variant requires trust in the directory structure (and perhaps guardrails to ignore non-pages). There’s also a very slight runtime cost to scanning the filesystem during config load, but it’s negligible for our scale.

s1m0n1au:
It’s also worth noting that the ecosystem already embraces this pattern. Several Vite plugins virtualize MPA setups by scanning directories and even generating HTML templates—think vite-plugin-virtual-mpa, vite-plugin-mpa, or vite-plugin-html-template.

kbventures:
@s1m0n1au 

"It’s also worth noting that the ecosystem already embraces this pattern. Several Vite plugins virtualize MPA setups by scanning directories and even generating HTML templates—think vite-plugin-virtual-mpa, vite-plugin-mpa, or vite-plugin-html-template" 

URL?.

s1m0n1au:
> @s1m0n1au 
> 
> 
> 
> "It’s also worth noting that the ecosystem already embraces this pattern. Several Vite plugins virtualize MPA setups by scanning directories and even generating HTML templates—think vite-plugin-virtual-mpa, vite-plugin-mpa, or vite-plugin-html-template" 
> 
> 
> 
> URL?.

https://github.com/emosheeep/vite-plugin-virtual-mpa?tab=readme-ov-file

https://github.com/IndexXuan/vite-plugin-mpa

https://github.com/IndexXuan/vite-plugin-html-template

kbventures:
We have indeed had several pages not make it into production build because they don't exist in the config file.   This is not an isolated incident.

But, the opposite will be true if we introduce working code/file structures accidentally, which will be less obvious since it's not explicitly found in the build file, and we then will expose code pages we may not intend. 

A 404 is a bad user experience, but it also guarantees we never expose something we don't want to expose.

s1m0n1au:
> We have indeed had several pages not make it into production build because they don't exist in the config file. This is not an isolated incident.
> 
> But, the opposite will be true if we introduce working code/file structures accidentally, which will be less obvious since it's not explicitly found in the build file, and we then will expose code pages we may not intend.
> 
> A 404 is a bad user experience, but it also guarantees we never expose something we don't want to expose.

Cuz I think dynamic discovery is a good way to avoid missing pages.

For private pages, we can follow the private-route pattern from Next.js: skip building any directory whose name begins with an underscore. For example, src/pages/_some-hidden-page/index.html would be ignored during the production build. 

When the pages is ready, just remove the prefix underscore.

kbventures:
@s1m0n1au 

I didn't mean explicitely wanted private page.  I meant a page we created and forgot to delete.  If the file and directory are correct, we could end up with pages that aren't supposed to "EXIST".  A test page, for example. 

For this project’s scope (open source), we can accept the small risk of exposing unwanted files. However, we need stronger error handling and clearer build output—logging which pages and directories are created would make debugging far easier.

kbventures (changes requested):
I left some changes to be made at the end of the discussion.  Great work.

s1m0n1au:
> @s1m0n1au 
> 
> 
> 
> I didn't mean explicitely wanted private page.  I meant a page we created and forgot to delete.  If the file and directory are correct, we could end up with pages that aren't supposed to "EXIST".  A test page, for example. 
> 
> 
> 
> For this project’s scope (open source), we can accept the small risk of exposing unwanted files. However, we need stronger error handling and clearer build output—logging which pages and directories are created would make debugging far easier.

I think for the test page, we can just add an underscore for the directory. For example, `src/pages/_test-page`. You can access with `localhost:3000/pages/_test-pages/` in dev mode. But it will not be built if the function recognizes `_` prefix.

kbventures:
@s1m0n1au 

Some suggestions to make debugging easier: 

```javascript
	
				input: (() => {
					const inputs: Record<string, string> = {
						index: path.resolve('src', 'index.html')
					};
					console.log(`✓ Added main entry: ${path.resolve('src', 'index.html')}`);

					const pagesDir = path.resolve('src', 'pages');
					if (!fs.existsSync(pagesDir)) {
						console.log('ℹ No pages directory found, using main entry only');
						return inputs;
					}

					for (const folder of fs.readdirSync(pagesDir)) {
						const folderPath = path.join(pagesDir, folder);
						const htmlEntry = path.join(folderPath, 'index.html');

						try {
							const stats = fs.statSync(folderPath);
							if (stats.isDirectory()) {
								if (fs.existsSync(htmlEntry)) {
									inputs[folder] = htmlEntry;
									console.log(`✓ Added page: ${folder}`);
								} else {
									console.log(`⊘ Skipped ${folder}: no index.html found`);
								}
							}
						} catch (error) {
							console.warn(`⚠ Error processing ${folder}:`, error.message);
						}
					}
					return inputs;
				})()
			}
		},

```

This is overly verbose but since it's in the build it can only help debugging quickly if it fails.

vite.config.ts
```diff
@@ -1,12 +1,24 @@
...
+import { createLogger } from 'vite';
+const logger = createLogger();
+
+declare global {
+	var __appInputCache: {
+		inputs: Record<string, string> | null
+	} | undefined;
+}
+
+const inputCache = globalThis.__appInputCache ??= { inputs: null };
```
manvinderjit: **question:** How will the cache behave if a new page is added at runtime? Will it pick the new page automatically or require a restart?

vite.config.ts
```diff
@@ -35,25 +47,50 @@ export default defineConfig(({ mode }) => {
...
+						logger.info('ℹ No pages directory found, using main entry only');
+						inputCache.inputs = inputs;
+						return inputs;
+					}
+
+					for (const folder of fs.readdirSync(pagesDir)) {
+						if (folder.startsWith('_')) {
+							logger.info(`⊘ Skipped ${folder}: private directory (underscore prefix)`);
+							continue;
+						}
```
manvinderjit: **question:** Is it possible to skip files based on the environment type, e.g. dev or prod.

manvinderjit (commented):
**praise**: Great work @s1m0n1au taking the initiative on this. It does help improve the build process. Just left a couple of questions as Ken has already covered most of it.

The synchronous file scanning here will be limited to local dev environment and GitHub Runners. It should not impact the actual production environment because we only deploy the built artifact over there.

But for any production level changes do consider that the Cloudflare Workers come with severe restrictions and have a strict `10ms` CPU runtime limit.

vite.config.ts
s1m0n1au: This vite config is only for build. By default, the `vite dev` will handle everything about hot reload(We can config the hot reload too).

vite.config.ts
s1m0n1au: For now, this vite config is only for build. 

The `vite preview` require a restart if  a new page is added.

kbventures (approved):
Great work!

vite.config.ts
kbventures: @s1m0n1au is this reallyl necessary?

```
declare global {
	var __appInputCache: {
		inputs: Record<string, string> | null
	} | undefined;
}

const inputCache = globalThis.__appInputCache ??= { inputs: null };
```

kbventures (changes requested):
Check my notes

vite.config.ts
s1m0n1au: > @s1m0n1au is this reallyl necessary?
> 
> ```
> declare global {
> 	var __appInputCache: {
> 		inputs: Record<string, string> | null
> 	} | undefined;
> }
> 
> const inputCache = globalThis.__appInputCache ??= { inputs: null };
> ```

This prevents duplicate logging of the result.

Reference:
https://samuelplumppu.se/blog/detect-vite-plugin-restarts-to-avoid-rerunning-expensive-tasks

kbventures: @s1m0n1au as discussed during meeting with @0916dhkim please move this to it's own plugin.

kbventures (changes requested):
Ready to be merged.  Just need to move suggested code to own plugin.

bcha92 (approved):
nice

kbventures (approved):

PR #284, opened 2025-10-14
test: unit tests for the back-end setup

s1m0n1au:
## Summary
- Set up a unit test framework/library for the back recommended by Cloudflare

## References
- https://developers.cloudflare.com/workers/testing/vitest-integration/write-your-first-test/
- https://developers.cloudflare.com/workers/testing/vitest-integration/test-apis/
- https://developers.cloudflare.com/workers/testing/vitest-integration/recipes/


vitest.config.ts
```diff
@@ -1,11 +1,28 @@
...
-		poolOptions: {
-			workers: {
-				wrangler: { configPath: './wrangler.toml' }
-			}
+export default defineWorkersConfig(async () => {
+	const migrationsPath = path.join(process.cwd(), 'db/migrations');
+	const migrations = await readD1Migrations(migrationsPath);
+
+	const rawSeedSql = fs.readFileSync(path.join(process.cwd(), 'db/seeds/seed-data.sql'), 'utf8');
+	const seedSql = rawSeedSql.split('\n').filter((line) => !line.trim().startsWith('--')).join('');
```
madcampos: **nitpick:** This can be moved to a file on the seeds folder so that it does not clutter the configuration.

package.json
```diff
@@ -53,7 +44,8 @@
 	},
 	"devDependencies": {
 		"@cloudflare/vite-plugin": "1.9.6",
-		"@cloudflare/vitest-pool-workers": "0.8.55",
+		"@cloudflare/vitest-pool-workers": "0.8.49",
+		"@cloudflare/workers-types": "^4.20251011.0",
```
madcampos: **question:** Isn't this module basically deprecated and superseeded by the `wrangler types` command?

package.json
```diff
@@ -7,34 +7,25 @@
...
-
 		"typecheck:api": "tsc -p ./api --noEmit",
 		"typecheck:client": "tsc -p ./src --noEmit",
 		"typecheck": "npm run typecheck:api && npm run typecheck:client",
 		"lint:js": "eslint",
 		"lint": "npm run typecheck && npm run lint:js",
-
 		"test:coverage": "vitest run --coverage",
 		"test:unit": "vitest run",
 		"test:api-file": "hurl --continue-on-error --jobs 4 --max-time 300 --test --variables-file .dev.vars --variables-file .api-test.vars",
```
madcampos, line 40: **nitpick:** Please keep the line breaks for readability.

api/tsconfig.json
```diff
@@ -1,11 +1,18 @@
...
-	"include": ["./**/*"],
+	"include": [
+		"./**/*"
+	],
 	"extends": "../tsconfig.base.json",
 	"compilerOptions": {
-		"lib": ["ESNext"],
+		"lib": [
+			"ESNext"
+		],
```
madcampos: **nitpick:** Please keep the formatting as it was before. The project formatter, `dprint`, should not change this as it was already formatted with the tool.

If you are using `prettier`, `biome`, or any other tool that is not `dprint`, please disable or uninstall those as they will mess up with the formatting already set by the project.

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
-	password: 'hashed-password',
-	access: ACCESS_LEVEL.ORGANIZER,
-	name: 'John Doe',
-	avatar: 'avatar-url'
+import { env } from 'cloudflare:test';
+import { applyD1Migrations } from 'cloudflare:test';
+import { beforeAll, describe, expect, it } from 'vitest';
+import { activateProfile, checkActiveEmail, checkExistingEmail, getHeartbeatInfo, getLoginInfo } from './data.ts';
+
+const USER = {
```
madcampos: **issue:** Use a more descriptive variable name. `USER` in this context doesn't give any clue to what this variable will be used for. Is it a single user used for **ALL** tests? Are there diferent users with diferent data?

Also, keep the data related to a test closer to the test using that data. It improves locality and makes tests more readable and easier to understand.

Tests are the type of code that _should_ be written in a verbose way to be very descriptive and avoid bugs by omitting or obscuring code. Make it simple, make it straight forward, and make it clear to someone reading the test what it checks for.

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
+	name: 'King Arthur',
+	access: 'organizer'
 };
 
-describe('Auth Data Functions', () => {
-	let env: MockEnvBindings;
-
-	beforeEach(() => {
-		env = new MockEnvBindings();
+describe('Auth data functions', () => {
```
madcampos: **suggestion:** while refactoring tests, add more descriptive descriptions to the tests.

Imagine someone is just running the tests without reading the files. The output should be descriptive to understand what is going on and what is being tested.

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
+	beforeAll(async () => {
+		await applyD1Migrations(env.Database, env.TEST_MIGRATIONS);
+		await env.Database.exec(env.SEED_SQL);
```
madcampos: **question:** Does this apply to every single test? If so, shouldn't it be called in the outer scope instead of here?

Also, what executing this SQL does? Does it overwrite existing rows? Does it seeds again a new database? Does it `DROP` the existing rows and then re-`INSERT` them?

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
-				password: sampleUser.password,
-				access: sampleUser.access
-			};
-
-			env.setDatabase({
-				first: <T>() => mockUser as T
-			});
-
-			const result = await getLoginInfo(env.Database, sampleUser.email);
+		it('returns credentials for an active, non-deleted profile', async () => {
```
madcampos: **suggestion:** Again: describe what are you trying to test, not the implementation of the test.

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
+			const result = await getLoginInfo(env.Database, 'king.arthur@camelot.uk');
 
 			expect(result).toEqual({
-				id: sampleUser.id,
-				password: sampleUser.password,
-				access: sampleUser.access
+				id: USER.id,
+				password: USER.password,
+				access: USER.access
```
madcampos: **issue:** Keep the user information _in the test_, and not in a variable referenced elsewhere.

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
 			});
 		});
 
-		test('returns null if no matching user found', async () => {
+		it('returns null if no matching user found', async () => {
```
madcampos: **question:** Why the change from `test` to `it`?

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
+				id: USER.id,
+				access: USER.access,
+				name: USER.name,
+				avatar: null,
+				status: 'profile-completed'
 			});
 		});
 
-		test('returns null if user is not found', async () => {
+		it('returns null if user is not found', async () => {
```
madcampos: Same here as the others: locality of the test data and change from `test` to `it`

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,94 @@
...
-		test.todo('returns false when update fails');
+		it('returns true when update succeeds', async () => {
+			const result = await activateProfile(env.Database, USER.email);
+
+			expect(result).toBe(true);
+		});
+
+		it.todo('returns false when update fail');
 	});
 });
```
madcampos, line 90: Same here: why the change from `test` to `it`?

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
+import { applyD1Migrations, env } from 'cloudflare:test';
+import { beforeAll, describe, expect, it, vi } from 'vitest';
+
+import type * as EmailModule from '../../email/index.ts';
```
madcampos: **nitpick:** Please use explicit imports. When you import or export into a property bag it is not immediatly evident what you are importing, what you are using and how it is used.

It makes harder to clean-up unused code and adds indirection to the the code.

If you have a name clash use this construct instead:
```typescript
import { nameClash as notNameClash } from './module.ts';
```

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+vi.mock('../../email/index.ts', async () => {
+	const actual = await vi.importActual<typeof EmailModule>(
+		'../../../src/email/index.ts'
+	);
+	return {
+		...actual,
+		sendAccountConfirmationEmail: vi.fn().mockResolvedValue(undefined)
+	};
+});
```
madcampos: **issue:** Don't.

Mocked modules create issues because they fake what the module does, they often deviate from the implementation causing issues where the test pass but the actual implementation has errors.

If you need control over the result of a function call while still keeping it "black boxed", use dependency injection to provide a callback to be called after the execution with the resulting value.

Furthermore, for testing emails, by default, no email will be sent and the email contents will be output to the console. You can control the behavior of email sending through environment variables.

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+const loadApp = async () => (await import('../../index.ts')).app;
```
madcampos: **suggestion:** Prefer a regular import over dynamic import.

Also: use a regular function declaration instead of an arrow function.

Furthermore: if the app code itself (and it's internal state) doesn't change between tests, keep a global refernece instead of recrating the app every time, this is just wasteful and not needed.

api/src/routes/auth/index.db.test.ts
madcampos, line 1: **question:** Why is the file named like this? Shouldn't it be just `index.test.ts`?

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+describe('Authentication routes', () => {
+	beforeAll(async () => {
+		await applyD1Migrations(env.Database, env.TEST_MIGRATIONS);
+		await env.Database.exec(env.SEED_SQL);
+	});
```
madcampos: Same question as the other test file: if it is applied to all tests, shouldn't this be called in the outer scope?

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+	access: 'volunteer' as const,
+	status: 'active'
+} as const;
+
+const API_ENDPOINTS = {
+	SIGN_UP: '/api/auth/sign-up',
+	SIGN_IN: '/api/auth/sign-in',
+	SIGN_OUT: '/api/auth/sign-out',
+	ACTIVATE: '/api/auth/activate'
+} as const;
```
madcampos: **suggestion:** Here is a place where extracting information to a variable does not help and can actually hurt DX.

You can hard code the endpoints to the tests, as changing the endpoint breaks the tests and that is expected behavior.

When the person runs the test the information will be there already improving locality of the code instead of hidden away in a reference that needs to be navigated to, where it could very much be right by the test itself.

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+				message: 'Created a new profile and sent an email for confirmation'
+			});
+
+			const tokens = await env.ActivationTokens.list();
+			expect(tokens.keys.length).toBe(1);
+
+			const [firstActivationTokenKey] = tokens.keys;
+			if (!firstActivationTokenKey) {
+				throw new Error('Activation token was not created');
+			}
```
madcampos: **issue:** This error throwing here seems odd... Tests should assert something instead.

Also, the KV stores should be flushed between every test execution to avoid mising up data between tests.

Every test should be as self contained as possible. Meaning: it should execute all the steps required for an action from an inital state.

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+			expect(stored).toEqual(expect.objectContaining({
+				email: 'new@new.new'
+			}));
+
+			const { sendAccountConfirmationEmail } = await import('../../email/index.ts');
+			expect(sendAccountConfirmationEmail).toHaveBeenCalledWith(
+				expect.anything(),
+				expect.objectContaining({
+					email: 'new@new.new',
+					// apiKey: env.RESEND_API_KEY,
```
madcampos: **nitpick:** Remove unused code to tidy up the codebase. If you need to go back to it, the git history is here to help you.

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+					senderEmail: env.SENDER_EMAIL
+				})
+			);
+		});
+
+		describe('POST /api/auth/sign-in', () => {
+			it('returns 201 and creates a session when credentials are valid', async () => {
```
madcampos: **nitpick:** Better test description

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+				const app = await loadApp();
+				const response = await app.request(API_ENDPOINTS.SIGN_IN, {
+					method: 'POST',
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify(USER)
+				}, env);
+
+				expect(response.status).toBe(StatusCodes.CREATED);
+				await expect(response.json()).resolves.toEqual({ message: 'Sign in successful' });
```
madcampos: **issue:** Do not validate messages as they may change. The status code should be enough of a test if the operation succeeded or not.

Messages should be informative to the consumer of the API but not depended upon as they are ephemeral and subject to change. All good API designs should provide a way for code to infer the result of an operation, and sometimes a message to help users understand what happened. In this case the API provides HTTP status codes to inform the result of an operation.

api/src/routes/auth/index.db.test.ts
```diff
@@ -0,0 +1,142 @@
...
+						body: JSON.stringify({
+							...USER,
+							email: 'wrong@email.com'
+						})
+					},
+					env
+				);
+
+				expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
+				await expect(response.json()).resolves.toEqual({ message: 'Either your email/password combination is invalid, or your account is not active' });
```
madcampos: Same here: https://github.com/torontojs/communityhub/pull/284/files#r2429868980

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+import { env } from 'cloudflare:test';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type * as EmailModule from '../../email/index.ts';
+import type * as AuthUtilsModule from '../../utils/auth.ts';
+import type * as PasswordHashingModule from '../../utils/password-hashing.ts';
+import type * as PasswordStrengthModule from '../../utils/passwordStrengthCheck.ts';
+import { StatusCodes } from '../../utils/responses.ts';
+import type * as ProfileModule from '../profile/data.ts';
+import type * as AuthDataModule from './data.ts';
```
madcampos: Same issue as the other imports: prefer explicit imports.

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+	access: 'volunteer' as const,
+	status: 'active'
+} as const;
+
+const API_ENDPOINTS = {
+	SIGN_UP: '/api/auth/sign-up',
+	SIGN_IN: '/api/auth/sign-in',
+	SIGN_OUT: '/api/auth/sign-out',
+	ACTIVATE: '/api/auth/activate'
+} as const;
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429841323

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+import type * as ProfileModule from '../profile/data.ts';
+import type * as AuthDataModule from './data.ts';
+
+const USER = {
+	name: 'King Arthur',
+	email: 'king.arthur@camelot.uk',
+	password: 'H0lyGr@il42!',
+	access: 'volunteer' as const,
+	status: 'active'
+} as const;
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429771544

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+vi.mock('../../utils/auth.ts', async () => {
+	const actual = await vi.importActual<typeof AuthUtilsModule>(
+		'../../../src/utils/auth.ts'
+	);
+	return {
+		...actual,
+		revalidateSession: vi.fn().mockResolvedValue(null),
+		createSession: vi.fn().mockResolvedValue(undefined)
+	};
+});
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429827129

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
 
-	test.todo('failure: activation should not be successful if token expired');
+const loadApp = async () => (await import('../../index.ts')).app;
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429829979

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+	beforeEach(async () => {
+		vi.clearAllMocks();
 
-		const json: StatusResponse = await response.json();
+		const activationKeys = await env.ActivationTokens.list();
+		await Promise.all(activationKeys.keys.map(async ({ name }) => env.ActivationTokens.delete(name)));
 
-		expect(json.message).toBe('Account activated successfully');
+		const sessionKeys = await env.SessionTokens.list();
+		await Promise.all(sessionKeys.keys.map(async ({ name }) => env.SessionTokens.delete(name)));
```
madcampos: This should be done on the other tests that reference KV

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
-			new MockEnvBindings({
-				activations: {
-					// @ts-expect-error
-					get: () => 'test@email.com'
-				}
-			})
-		);
+describe('Authentication routes', () => {
+	beforeEach(async () => {
+		vi.clearAllMocks();
```
madcampos: Again: don't use mocks.

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
-		expect(json.message).toBe('Either your email/password combination is invalid, or your account is not active');
+	describe('POST /api/auth/sign-up', () => {
+		it('creates profile and activation token when payload is valid', async () => {
+			const app = await loadApp();
+			const response = await app.request(
+				API_ENDPOINTS.SIGN_UP,
+				{
+					method: 'POST',
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify(USER)
```
madcampos: **question:** Does `body` expect a string?

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+				},
+				env
+			);
+
+			expect(response.status).toBe(StatusCodes.OKAY);
+			await expect(response.json()).resolves.toEqual({
+				message: 'Created a new profile and sent an email for confirmation'
+			});
```
madcampos: Same here: https://github.com/torontojs/communityhub/pull/284/files#r2429868980

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+
+			const { hashPassword } = await import('../../utils/password-hashing.ts');
+			expect(hashPassword).toHaveBeenCalledWith('H0lyGr@il42!');
+
+			const { insertProfile } = await import('../../routes/profile/data.ts');
+			expect(insertProfile).toHaveBeenCalledWith(env.Database, {
+				email: USER.email,
+				name: USER.name,
+				password: 'hashed-password'
+			});
```
madcampos: **issue:** Whenever possible, don't use dynamic imports.

They make code harder to debug and harder to test as they make the code less deterministic.

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+
+			const tokens = await env.ActivationTokens.list();
+			expect(tokens.keys.length).toBe(1);
+
+			const [firstActivationTokenKey] = tokens.keys;
+			if (!firstActivationTokenKey) {
+				throw new Error('Activation token was not created');
+			}
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429853036

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+
+			const stored = await env.ActivationTokens.get(firstActivationTokenKey.name, 'json');
+			expect(stored).toMatchObject({
+				email: 'king.arthur@camelot.uk',
+				id: '3c5123c0-8548-4a02-a83c-32e9ce67eae8'
+			});
+
+			const { sendAccountConfirmationEmail } = await import('../../email/index.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+					senderEmail: env.SENDER_EMAIL,
+					token: expect.any(String)
+				})
+			);
+		});
+
+		it('returns 200 but skips creation when email already exists', async () => {
+			const app = await loadApp();
+
+			const { checkExistingEmail } = await import('../../routes/auth/data.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+			vi.mocked(checkExistingEmail).mockResolvedValueOnce(true);
```
madcampos: **issue**: mocks

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify(USER)
+				},
+				env
+			);
+
+			expect(response.status).toBe(StatusCodes.OKAY);
+
+			const { insertProfile } = await import('../../routes/profile/data.ts');
+			const { sendAccountConfirmationEmail } = await import('../../email/index.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+
+			expect(insertProfile).not.toHaveBeenCalled();
+			expect(sendAccountConfirmationEmail).not.toHaveBeenCalled();
+		});
+
+		it('returns 422 when password fails strength check', async () => {
+			const app = await loadApp();
+
+			const { passwordStrengthCheck } = await import('../../utils/passwordStrengthCheck.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+			expect(tokens.keys).toHaveLength(0);
+		});
+
+		it('return 401 when incorrect token', async () => {
+			const app = await loadApp();
+
+			const keys = await env.ActivationTokens.list();
+			await Promise.all(keys.keys.map(async ({ name }) => env.ActivationTokens.delete(name)));
+
+			const invalidToken = '3c5123c0-8548-4a02-a83c-32e9ce67eae8';
```
madcampos: **issue:** Please use an UUID with all zeros as the reference for non existent UUID. It is a valid UUID in the sense that it follows the format, but should not be a valid value in the application. As a bonus it is a very explicit value.

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify(USER)
+				},
+				env
+			);
+
+			expect(response.status).toBe(StatusCodes.CREATED);
+			await expect(response.json()).resolves.toEqual({ message: 'Sign in successful' });
+
+			const { createSession } = await import('../../../src/utils/auth.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+			expect(createSession).toHaveBeenCalledWith(
+				expect.objectContaining({
+					id: '3c5123c0-8548-4a02-a83c-32e9ce67eae8',
```
madcampos: **issue:** Do not depend on a value from another test. The tests can (and should) be run out of order and in parallel.

Each test should be independent and self contained.

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+					email: USER.email
+				})
+			);
+		});
+
+		it('returns 401 login with incorrect email', async () => {
+			const app = await loadApp();
+
+			const { getLoginInfo } = await import('../../routes/auth/data.ts');
+			vi.mocked(getLoginInfo).mockResolvedValueOnce(null);
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+				{
+					method: 'POST',
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify(USER)
+				},
+				env
+			);
+
+			expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
+			await expect(response.json()).resolves.toEqual({ message: 'Either your email/password combination is invalid, or your account is not active' });
```
madcampos: Same here: https://github.com/torontojs/communityhub/pull/284/files#r2429868980

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+		});
+
+		it('returns 401 login with incorrect password', async () => {
+			const app = await loadApp();
+
+			const { validatePassword } = await import('../../utils/password-hashing.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+				{
+					method: 'POST',
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify(USER)
+				},
+				env
+			);
+
+			expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
+			await expect(response.json()).resolves.toEqual({ message: 'Either your email/password combination is invalid, or your account is not active' });
```
madcampos: Same here: https://github.com/torontojs/communityhub/pull/284/files#r2429868980

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+		it('returns 204 when the user signs out successfully', async () => {
+			const token = '3c5123c0-8548-4a02-a83c-32e9ce67eae8';
+			const sessionPayload = {
+				id: token,
+				...USER,
+				token,
+				originalExpiry: new Date().toISOString()
+			};
+
+			const { revalidateSession } = await import('../../utils/auth.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+
+			await env.SessionTokens.put(token, JSON.stringify(sessionPayload));
+			vi.mocked(revalidateSession).mockResolvedValue(sessionPayload);
```
madcampos: **isue**: mocks

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+
+			expect(response.status).toBe(StatusCodes.NO_CONTENT);
+
+			const remaining = await env.SessionTokens.list();
+			expect(remaining.keys.some(({ name }) => name === token)).toBe(false);
+		});
+
+		it('returns 401 when the session is missing or invalid', async () => {
+			const app = await loadApp();
+			const { revalidateSession } = await import('../../utils/auth.ts');
```
madcampos: Same as here: https://github.com/torontojs/communityhub/pull/284/files#r2429899595

api/src/routes/auth/index.test.ts
```diff
@@ -1,99 +1,398 @@
...
+
+			vi.mocked(revalidateSession).mockResolvedValue(undefined);
```
madcampos: **issue**: mocks

madcampos (changes requested):
Please review the code, there are some serious anti-patterns going on.

A lot of the test code is following the same patterns, so I may have missed some, so take it that everywhere you see something that should be changed, change it.

api/src/routes/auth/index.test.ts
s1m0n1au: Yes. The body should be JSON string Payload.

package.json
```diff
@@ -72,7 +72,7 @@
 		"husky": "9.1.7",
 		"typescript": "5.8.3",
 		"vite": "^7.0.4",
-		"vitest": "3.2.4",
+		"vitest": "~3.2.0",
```
madcampos: **issue:** Please prefer `^` or fixed version numbers.

package.json
```diff
@@ -53,7 +53,7 @@
 	},
 	"devDependencies": {
 		"@cloudflare/vite-plugin": "1.9.6",
-		"@cloudflare/vitest-pool-workers": "0.8.55",
+		"@cloudflare/vitest-pool-workers": "0.8.49",
```
madcampos, line 56: **issue:** Unless it is strictly necessary, don't downgrade packages.

wrangler.toml
madcampos, line 1: **issue:** The variables are set in [`.dev.vars.example`](https://github.com/torontojs/communityhub/blob/dev/.dev.vars.example) and should not be updated in the `wrangler.toml` file.

On your local environment you should copy the `.dev.vars.example` file to `.dev.vars` and adjust the values as needed.

See: [Setup Guide](https://github.com/torontojs/communityhub/blob/dev/docs/contributing-guide/setup.md?plain=1#L14-L17)

madcampos (changes requested):
I haven't finished reviewing the tests themselves, but here are a couple of notes.

package.json
bcha92: **issue:** agreed. Downgrading packages are not recommended as it would introduce potentially a host of bugs and vulnerabilities that otherwise have been patched in later version. There are exceptions of course, but for minor (non-breaking) downgrades like these, it is generally not recommended.

Conversely, upgrading packages also warrants caution because we could be introducing breaking changes (i.e. `1.0.0` to `2.0.0`) that may render the current logic to fail or malfunction when executed. But rule of thumb is (but not always) to upgrade to the version just prior to the breaking changes (i.e. `1.0.4` to `1.1.6`)

bcha92: I see there's an annoying console warning log for an error when cleaning up temp directories while attempting to use `npm run test` on v0.8.55 that doesn't appear on v0.8.49. It's not a blocker per se for testing, but it's annoying having to scroll up a bit to see if the tests passed or failed.

After some google searching, I think this `console.warn` is only seen by Windows users and there is no clear fix. Then again this SDK is a bit experimental to say at the least with more weird stuff worse than this, so I think we can just revert the change here back to the original (latest) version and put in a non-urgent ticket for this down the road.

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,90 @@
...
-	getLoginInfo
-} from './data.ts';
-
-interface User {
-	id: string;
-	email: string;
-	password: string;
-	access: AccessLevel;
-	name: string;
-	avatar?: string;
```
bcha92, line 18: **comment:** I think this is a good opportunity to move this interface into a schema where we can re-inforce this strict typing in both the testing environment and in the logic.
**issue:** I don't see the interface being re-introduced to re-inforce strict typing, would that cause some issues down the road when testing for values not consistent with the user attributes?

bcha92 (changes requested):

package.json
madcampos: @bcha92 is there a report already for this warning? If there is, please link the issue so we can also keep track of that.

If not, would be a good idea to create a bug report. Please try to be as specific as possible with the report and provide as much info as you can.

api/src/routes/auth/data.test.ts
madcampos: +1: This interface should be inferred from an schema already, and if there is a need for slightly altering the interface, that's what the `Omit`, `Pick`, and `Partial` [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) are for.

madcampos: One addition to that: as the parameters for the API are already covered by zod (or should be), adding tests to that is just bloat, as the library should already be tested.

We should only add tests to things that fall outside of library code, meaning, things we add _after_ the zod validation runs and is successful.

package.json
bcha92: @madcampos only in-console (tried running with the `"@cloudflare/vitest-pool-workers": "0.8.55"` version reverted)
Screenshot 2025-10-27 010200

madcampos: @bcha92 raise an issue on the cloudflare repo, this seems like a library issue.
Also, if you have the resources, would be good to try on linux or mac to see if it is a Windows only issue.

kbventures (approved):

package.json
bcha92: hey @madcampos issue on the cloudflare repo is already up regarding failure to remove temp directory. Temporary solution is to downgrade the `"@cloudflare/vitest-pool-workers"` package to version `"0.8.49"`

https://github.com/cloudflare/workers-sdk/issues/10511

I will close this comment if that is okay.

bcha92 (approved):
Looks good (just update base branch)

api/src/routes/auth/data.test.ts
```diff
@@ -1,139 +1,90 @@
...
+import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
+import { activateProfile, checkActiveEmail, checkExistingEmail, getHeartbeatInfo, getLoginInfo } from './data.ts';
+
+beforeAll(async () => {
+	await applyD1Migrations(env.Database, env.TEST_MIGRATIONS);
+});
 
+beforeEach(async () => {
+	await env.Database.exec(env.SEED_SQL);
+});
```
madcampos, line 11: **question**: won't this cause errors?

Like, if you are seeding the database again before each test, you will try to insert the same ID over and over again, causing a constraint validation to fail.

Is there any "cleanup" process to remove the seeded data?

madcampos (approved):
Approved with a non blocking question :)

api/src/routes/auth/data.test.ts
s1m0n1au: According to the docs, when `isolatedStorage: true` default have the  "cleanup" process to remove the seeded data.
image
image

Reference:
https://developers.cloudflare.com/workers/testing/vitest-integration/configuration/ 
https://developers.cloudflare.com/workers/testing/vitest-integration/isolation-and-concurrency/

PR #294, opened 2025-11-03
Task Component Design Token

roshan-budhathoki:
Creating task component for pending and completed task. 

Here, "pending" is considered as default parameter.


src/components/Task/Task.tsx
```diff
@@ -0,0 +1,39 @@
+import { forwardRef, type ParamHTMLAttributes } from 'react';
```
kbventures: I think I brought this up in another PR of yours.   Same comment. We shouldn't utilize forwardRef untless there's deliberate reason for its use not just beause it might be convinient later.

src/components/Task/Task.css
```diff
@@ -0,0 +1,7 @@
+.taskParent {
```
kbventures: Would this be an opportunity to utilize our @layer component?

kbventures (changes requested):
Great work.  See notes.

src/components/Task/Task.css
```diff
@@ -0,0 +1,11 @@
+@layer components {
+	.taskParent {
+		display: flex;
+		gap: var(--spacing-sm) 0;
+		align-items: center;
+		font-weight: bold;
```
bcha92, line 6: please use preset value `font-weight: var(--weight-bold);` instead of just `bold`

bcha92 (changes requested):
I haven't fully reviewed it yet (need to find time to manually test this on local env first) but it's the first thing I found.

bcha92:
Closing PR as stale.

PR #302, opened 2025-11-10
Chore/deployment

### LINKED ISSUE #285: [Chore]: Deploy website to test domain
### Description

Modify the integration pipeline to deploy the website to a test environment on cloudflare and connect it to a domain.

manvinderjit:
## Summary

Create a GitHub Actions pipeline to automatically deploy the project to dev and production environments.

## Tests

Tested manually in a private fork of the repository

## Additional information

Need to do the following for this to work:

1. Update API tokens and GitHub secrets
2. Create a Cloudflare API user with appropriate permissions.


.github/workflows/cd-deployments.yml
madcampos, line 1: **question**: I think right now the deployment is done "in reverse", from cloudflare observing the changes on github.

This would change it to go from github to cloudflare right?
Do we need to disable the connection on cloudflare's dashboard as well?

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      deployments: write
+
+    env:
+      NODE_VERSION: 22
```
madcampos: **question:** Can we set this to a moving target like "lts"?

Or rather, skip hardcoding the version here entirely and relay on the `.nvmrc` file?

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+      DEV_COMM_HUB_PASSWORD_RESET_TOKENS_ID: ${{ secrets.DEV_COMM_HUB_PASSWORD_RESET_TOKENS_ID }}
+      DEV_COMM_HUB_SENDER_EMAIL: ${{ secrets.DEV_COMM_HUB_SENDER_EMAIL }}
+
+      # PROD
+      PROD_COMM_HUB_D1_DB_NAME: ${{ secrets.PROD_COMM_HUB_D1_DB_NAME }}
+      PROD_COMM_HUB_D1_DB_ID: ${{ secrets.PROD_COMM_HUB_D1_DB_ID }}
+      PROD_COMM_HUB_SESSION_TOKENS_ID: ${{ secrets.PROD_COMM_HUB_SESSION_TOKENS_ID }}
+      PROD_COMM_HUB_ACTIVATION_TOKENS_ID: ${{ secrets.PROD_COMM_HUB_ACTIVATION_TOKENS_ID }}
+      PROD_COMM_HUB_PASSWORD_RESET_TOKENS_ID: ${{ secrets.PROD_COMM_HUB_PASSWORD_RESET_TOKENS_ID }}
+      PROD_COMM_HUB_SENDER_EMAIL: ${{ secrets.PROD_COMM_HUB_SENDER_EMAIL }}
```
madcampos: **suggestion:** Would it be possible to use the variables set on cloudflare's dashboard instead or do we absolutely need to set them here?

That way they are not present on `wrangler.toml` and so are not visible to everyone and also not on github secrets which have known issues with leaking secrets

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
```
madcampos: Same question here: can we infer it from the nvmrc file?

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+          cache: "npm"
+
+      - name: Install dependencies and utilities
+        run: |
+          npm ci
+          sudo apt-get update
+          # Install envsubst and jq to substitute secrets in wrangler.toml
```
madcampos: **suggestion:** wrnagler also supports a `wrangler.json` file as an alternative to the TOML syntax. Maybe using that instead will make things easier, no?

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+        run: |
+          echo "Attempting to delete Dev database"
+          npx wrangler d1 delete ${{ env.DEV_COMM_HUB_D1_DB_NAME }} -y || true > /dev/null
+
+      - name: Create New Dev D1 Instance (Dev Only)
+        if: ${{ env.TARGET_ENV == 'dev' }}
+        run: |
+          npx wrangler d1 create ${{ env.DEV_COMM_HUB_D1_DB_NAME }} --env dev > /dev/null
+          DB_ID=$(npx wrangler d1 info ${{ env.DEV_COMM_HUB_D1_DB_NAME }} --json | jq -r '.uuid')
+          echo "DEV_COMM_HUB_D1_DB_ID=$DB_ID" >> $GITHUB_ENV
```
madcampos: **question:** Why delete and recreate the dev db every time?

Would it make sense instead to invoke an SQL script that drops the tables on the database, or run a rollback on all migrations?

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Inject Bindings/Vars into Wrangler TOML using envsubst (Dev Only)
+        if: ${{ env.TARGET_ENV == 'dev' }}
+        run: |
+          # The list of variables to substitute, matching the placeholders in wrangler.toml
+          export ENV_VARS_TO_SUB="\$DEV_COMM_HUB_D1_DB_ID \$DEV_COMM_HUB_D1_DB_NAME \$DEV_COMM_HUB_SESSION_TOKENS_ID \$DEV_COMM_HUB_ACTIVATION_TOKENS_ID \$DEV_COMM_HUB_PASSWORD_RESET_TOKENS_ID \$DEV_COMM_HUB_SENDER_EMAIL"
+
+          # Use envsubst to replace DEV placeholders in wrangler.toml, overwriting the file
+          envsubst "$ENV_VARS_TO_SUB" < wrangler.toml > wrangler.toml.tmp
+          mv wrangler.toml.tmp wrangler.toml
```
madcampos: **suggestion:** Two things:
1. If the variables are okay to be shared, like a random meaningless id for something, I think it is fine leaving it in the wrangler config.
2. As this is dev, why not add a `.dev.vars` file with all of the values instead of fiddling with the wrangler config?

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Apply D1 Migrations to DEV D1 Database (Dev Only)
```
madcampos: **thought:** Migrations should be applied in all environments. They also need to be thoroughly tested to be compatible with the previous data and not break anything.

Like, if a new version needs a new format for the database, **not** applying the migration will break the application. So it makes sense to apply the migrations regardless of environment.

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+        env:
+          # Expose Prod D1 ID as a shell variable for envsubst
+          PROD_COMM_HUB_D1_DB_ID: ${{ secrets.PROD_COMM_HUB_D1_DB_ID }}
+        run: |
+          # The list of variables to substitute, matching the placeholders in wrangler.toml
+          export ENV_VARS_TO_SUB="\$PROD_COMM_HUB_D1_DB_ID \$PROD_COMM_HUB_D1_DB_NAME \$PROD_COMM_HUB_SESSION_TOKENS_ID \$PROD_COMM_HUB_ACTIVATION_TOKENS_ID \$PROD_COMM_HUB_PASSWORD_RESET_TOKENS_ID \$PROD_COMM_HUB_SENDER_EMAIL"
+
+          # Use envsubst to replace PROD placeholders in wrangler.toml, overwriting the file
+          envsubst "$ENV_VARS_TO_SUB" < wrangler.toml > wrangler.toml.tmp
+          mv wrangler.toml.tmp wrangler.toml
```
madcampos: **suggestion:** These variables should be kept on cloudflare's dashboard and read from there instead of messing around with the config file

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Apply D1 Migrations to PROD D1 Database (Prod Only)
+        env:
+          PROD_COMM_HUB_D1_DB_ID: ${{ env.PROD_COMM_HUB_D1_DB_ID }}
+        if: ${{ env.TARGET_ENV == 'production' }}
+        run: npx wrangler d1 migrations apply Database --env production --remote
```
madcampos: This is the same command as the dev one, expect for the environment. use the `TARGET_ENV` variable instead and combine everything into a single action.

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,129 @@
...
+          apiToken: ${{ env.CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ env.CLOUDFLARE_ACCOUNT_ID }}
+          environment: ${{ env.TARGET_ENV }}
+          workingDirectory: "./"
+
+          # Pass secrets directly for the Worker's environment
+          secrets: |
+            RESEND_API_KEY
+        env:
+          RESEND_API_KEY: ${{ env.TARGET_ENV == 'production' && secrets.PROD_COMM_HUB_RESEND_API_KEY || secrets.DEV_COMM_HUB_RESEND_API_KEY}}
```
madcampos: **suggestion:** Please read their docs and check if the secret can be kept in cloudflare's dashboard instead of being passed arround like this.

db/seeds/seed-data.sql
madcampos, line 1: **praise:** 10/10 love that the passowrds were kept in theme haha

[Brian](https://www.youtube.com/watch?v=jHPOzQzk9Qo&pp=ygU0bGlmZSBvZiBicmlhbiBhbHdheXMgbG9vayBvbiB0aGUgYnJpZ2h0IHNpZGUgb2YgbGlmZQ%3D%3D) would be pleased.

wrangler.toml
madcampos, line 1: **question:** Why not keep the ids here?

If that is a security concern, I think the best option is to provide them in two places instead:
1. `.dev.vars` for local development
2. Cloudflare's dashboard for dev and prod environments

We should also set the `keep_vars` option to `true` so they are not deleted when wrangler run and the values will be merged with what is on the dashboard.

.github/workflows/cd-deployments.yml
manvinderjit: @kbventures wanted to delete and recreate the database each time. 

Also, deleting and recreating the database is faster, compared to using a SQL script that drops the tables on the database or rolling back migrations. It will reduce the reads and writes on the database so that we don't go above the free tier.

.github/workflows/cd-deployments.yml
manvinderjit: I already have created an issue around that but the goal is to get it working first and once we have it figured out, we can migrate to jsonc file. 
The jsonc file gives additional configuration options but we still can't use GitHub secrets in it.

.github/workflows/cd-deployments.yml
manvinderjit: 1. While it is generally considered acceptable to have the wrangler file public, it may be a better idea to obscure the values if possible.
2. All the ids and configuration values must be present in the wrangler.toml or wrangler.jsonc file. The thing is, there is no way to reference dev.vars or GitHub secrets in the file. So I am editing the file on the fly to substitute the values. 

The other option is to have the wrangler file committed to the repo, making it public. I would prefer to not do that just in case.

.github/workflows/cd-deployments.yml
manvinderjit: Migrations are being applied to both environments. I kept the steps separate to minimize the chances of dev migration accidentally being deployed to prod accidentally.

.github/workflows/cd-deployments.yml
manvinderjit: These variables can NOT be kept on cloudflare's dashboard because they are used for the wrangler file. The file requires the bindings to be present. While defining bindings in the dashboard is possible, but because we are also using CI/CD, the recommendations is to use the wrangler file as source of truth

image

.github/workflows/cd-deployments.yml
manvinderjit: Migrations are being applied to both environments. I kept the steps separate to minimize the chances of dev migration accidentally being deployed to prod accidentally.

.github/workflows/cd-deployments.yml
manvinderjit: This command deploys the key in Cloudflare's Dashboard itself and not in wrangler file. If you want, we can remove it from the pipeline and put it directly in cloudflare's dashboard.

wrangler.toml
manvinderjit: 1. Refer https://github.com/torontojs/communityhub/pull/302#discussion_r2550440619

2. While we can merge the values, cloudflare recommends against it. https://github.com/torontojs/communityhub/pull/302#discussion_r2550457397

.github/workflows/cd-deployments.yml
manvinderjit: These variables are for substitution into the wrangler toml file for bindings. Cloudflare recommends having them in the wrangler file instead of using keep_vars=true and using variables from both places.

.github/workflows/cd-deployments.yml
manvinderjit: Sure.

.github/workflows/cd-deployments.yml
madcampos: I don't particularly like string interpolation like this to be honest... It sounds hacky... :/

.github/workflows/cd-deployments.yml
madcampos: So, instead of editing the `wrangler.toml` file, why not output the variables from github secrets to a `.dev.vars` file? That way there is no interpolation on the wrangler config, and the secrets are still segregated to a single place.

My concerns are basically with string interpolation and how that looks hacky, and the separation of regular variables from secrets.

.github/workflows/cd-deployments.yml
madcampos: If you are configuring everything based on the environment, I don't see a problem on keeping the same code and using a variable to set the environment.

In fact it is better and less error prone than duplication.

.github/workflows/cd-deployments.yml
madcampos: We should use secrets instead of variables: https://developers.cloudflare.com/workers/configuration/secrets/#compare-secrets-and-environment-variables

Furthermore:
> To set different secrets for each Cloudflare environment, create files named `.dev.vars.<environment-name>` or `.env.<environment-name>`.

[Reference](https://developers.cloudflare.com/workers/configuration/secrets/#:~:text=To%20set%20different%20secrets%20for%20each%20Cloudflare%20environment%2C%20create%20files%20named%20.dev.vars.%3Cenvironment%2Dname%3E%20or%20.env.%3Cenvironment%2Dname%3E.)

madcampos: I've closed the other comments related to this to keep everything in one thread.

kbventures (dismissed):

bcha92 (dismissed):
Looks good. Approving this. But since I can't test this without a cloudflare account, have you addressed Marco's comments?

.github/workflows/cd-deployments.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Reset Dev Database Tables
+        if: ${{ env.TARGET_ENV == 'dev' }}
+        run: |
+          # List tables from most-dependent (children) to least-dependent (parents)
+          TABLES=("documents" "access" "role" "team" "profile_skills" "profile_links" "profile" "event_log" "migrations")
+
+          for table in "${TABLES[@]}"; do
+            echo "Dropping table: $table"
+
+            npx wrangler d1 execute Database --env dev --remote --command "DROP TABLE IF EXISTS \"$table\";"
```
Shajansheriff: We're executing wrangler d1 execute once per table in a loop. Since each call is a remote request, this adds unnecessary latency to the CI pipeline.

Suggestion: move the DROP statements into a single SQL file and execute them in one wrangler d1 execute --file call. This will reduce network overhead and speed up the workflow.

.github/workflows/cd-deployments.yml
manvinderjit: Yes. 

I have disabled the deployments on CloudFlare for the following:

1. Volunteer (Pages-based deployment, doesn't use workers)
2. VMS-Dev (Old branch, before frontend and backend consolidation)
3. VMS (Old branch, before frontend and backend consolidation)

image

@madcampos @kbventures Do we want to delete the older workers and pages or keep them for reference?

.github/workflows/cd-deployments.yml
manvinderjit: The workflow now users the ```.nvmrc``` file!

.github/workflows/cd-deployments.yml
manvinderjit: Migrations are now applied based on the environment using the same block of code.

.github/workflows/cd-deployments.yml
manvinderjit: Reverted ```wrangler.toml``` to previous version wherein it has all the variables. Makes the process simpler and is a Cloudflare recommended standard approach.

.github/workflows/cd-deployments.yml
manvinderjit: Resend API key is now in Cloudflare's Dashboard. We will have to manually update it though for future changes.

wrangler.toml
manvinderjit: ID's  are now in wrangler.toml

.github/workflows/cd-deployments.yml
manvinderjit: @Shajansheriff Thank you for the input. 

The pipeline now uses an SQL file to DROP tables.

kbventures (dismissed):

bcha92 (dismissed):

bcha92 (approved):

bcha92 (approved):

PR #303, opened 2025-11-10
docs: update `unit-tests.md`

s1m0n1au:


docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
 ```shell
 npm test
 ```
```
madcampos, line 30: **suggestion**: there should be an explanation of the command to run an what is the expected output instead of jumping straight to the command.

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+## Principles
+
+1. **No mocking**
+   - Use the real Workers test runtime (D1, KV, R2, timers). Storage is isolated per test.
+2. **Do not test Zod/validators**
+   - Validation libraries are well-tested. Avoid re-testing them.
+3. **Types from schemas**
+   - Derive types with `z.infer<typeof Schema>` (or Valibot/TypeBox equivalents). Use `Pick`/`Omit`/`Partial` when necessary.
+4. **Focus on post-validation logic**
+   - Validation problems are already caught by Hurl/API tests. Unit tests assert business logic and side effects **after** validation.
```
madcampos: **nitpick**: This should be formatted like a seties of headings with the descriptions underneath.

Like so:
```markdown
### Principle
Description here
```

madcampos: **suggestion**: It would also benefit from explanations of the _why_  for each principle.

E.g.:

This is correct, but incomplete.
> Validation libraries are well-tested. Avoid re-testing them.

A better explanation would be:
> Testing any library code is redundant and adds noise to the tests. The expectation is that library authors test their codes so we don't need to.
>
> Tests should only be concerned with our own code and logic, not with how a library works or it's behaviours.

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+
+## Writing classic unit tests
+
+A "pure" function example (no side effects):
```
madcampos: **nitpick:** Instead of adding a brief comment about side effects, it would be better to explain the scenario and why the test is doing what it does.

docs/contributing-guide/testing/01-unit-tests.md
madcampos: **suggestion**: Another principle to add is that tests should be repeatable. Meaning when you run the same test multiple times, it should yield the same results. If a test does not have the same result in subsequent runs, then it is not possible to compare the before and after a code change to verify if a failure was introduced by the new code.

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+	it('handles big and special values', () => {
+		expect(clamp(Number.MAX_SAFE_INTEGER + 1, 0, Number.MAX_SAFE_INTEGER))
+			.toBe(Number.MAX_SAFE_INTEGER);
+		expect(clamp(Infinity, 0, 1)).toBe(1);
+		expect(() => clamp(NaN as any, 0, 1)).toThrow(/number/);
+	});
+});
+```
+
+## Tests that use real Workers runtime
```
madcampos: **suggestion**: in this section, instead of saying what APIs are available from Cloudflare, it would be better to provide some patterns that recurr on the code and why we do it in a certain way.

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+
+```ts
+// test/setup.ts
+import { env } from 'cloudflare:test';
+
+// Example: ensure D1 schema exists before tests
+export default async function() {
+	await env.Database.prepare('CREATE TABLE IF NOT EXISTS audit (sid TEXT, userId TEXT)').run();
+}
+```
```
madcampos: Why is it done this way, any caveats that need to be stated for people not familiar with the codebase?

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+
+		// KV side-effect present
+		const kv = await env.SESSIONS.get(`sid:${sid}`, 'json');
+		expect(kv).toEqual({ userId: 'u1' });
+
+		// D1 side-effect present
+		const row = await env.Database.prepare('SELECT COUNT(*) as c FROM audit WHERE sid = ? AND userId = ?')
+			.bind(sid, 'u1').first<{ c: number }>();
+		expect(row?.c).toBe(1);
+	});
```
madcampos, line 181: **suggestion**: This should be an explanation of a pattern, like "when you want to create a new cookie, this is the pattern that we use..."

Also, it should include an explanation of why, and if the code is repeated multiple times, it can be extracted to an utility function, and that function should have tests of itself.
Again, this is to avoid noise, so the utility function is only tested once, and then any consumer of it doesn't need to care about testing it's behaviour again.

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+});
+```
+
+### DB helper functions
+
+```ts
+// src/db.ts
+import type { Bindings } from './app';
+
+export async function countAudit(env: Bindings['DB']) {
```
madcampos: **issue:** I think the binding we are using is `Database` instead of `DB`

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+	const res = await app.request('/api/files/greeting.txt', { method: 'PUT', body: payload }, env);
+	expect(res.status).toBe(201);
+
+	// Assert side-effect in R2
+	const obj = await env.FILES.get('greeting.txt');
+	expect(await obj?.text()).toBe('hello');
+});
+```
+
+### DB helper functions
```
madcampos: Again here, instead of dumping code that without context is meaningless, explain why the code is here, the pattern it represents, how the pattern is used in code, and when the person should use that pattern

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+	const c0 = await countAudit(env.Database);
+	// Seed one audit row via API
+	await (await fetch('http://dummy', { method: 'POST', body: JSON.stringify({ userId: 'seed' }) })).body?.cancel(); // placeholder
+	await env.Database.prepare('INSERT INTO audit (sid, userId) VALUES (?, ?)').bind('s1', 'u1').run();
+	const c1 = await countAudit(env.Database);
+	expect(c1).toBe(c0 + 1);
+});
+```
+
+> Note: Each test runs with **isolated storage** (enabled by default). Writes are undone after the test, so ordering is reliable.
```
madcampos: **issue:** This should be more proeminent information and not a footnote. It completely changes how a user should think about the tests.

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+
+---
```
madcampos: **nitpick:** Don't add random horizontal lines.

```suggestion
```

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+		const kv = await env.SESSIONS.get(`sid:${sid}`, 'json');
+		expect(kv).toEqual({ userId: 'u42' });
+
+		const row = await env.Database.prepare('SELECT COUNT(*) as c FROM audit WHERE sid=?').bind(sid).first<{ c: number }>();
+		expect(row?.c).toBe(1);
+	});
+});
+```
+
+> We do **not** unit test validators themselves; Hurl or integration tests already ensure payload mismatches return 4xx.
```
madcampos: **issue:** This should also not be a footnote, and instead be highlighted as it completely changes what kind of things are being tested.

docs/contributing-guide/testing/01-unit-tests.md
```diff
@@ -26,3 +26,208 @@ For unit testing we use [`vitest`](https://vitest.dev/guide/), to run the unit t
...
+> Note: Each test runs with **isolated storage** (enabled by default). Writes are undone after the test, so ordering is reliable.
+
+---
+
+## Workers side effect tests with Hono OpenAPI
+
+These tests verify that:
+
+1. The OpenAPI document endpoint responds and contains expected metadata.
+2. Endpoints **perform side effects** against KV/D1/R2 and return the expected HTTP contract defined in OpenAPI.
```
madcampos: **suggestion:** This is a good start of explaining why use this pattern, but can be expanded.

madcampos (changes requested):
After reading the text a second time, it sounds too much like AI slop.

The text is hollow and dry, in a bad way. Yes, it explains what things are and what they do, but it does so in a way a dictionary would do it, flat, empty, and without intentionality.

The documentation for a project should convey the reasoning for the decisions, the tradeoffs, the _whys_!

Please rewrite it yourself, with your own words, or if you use any LLM to write something review and edit it. Add your voice.

In the end of the day, any person can use an LLM and generate very decent text the same way, regardless of knowledge or life journey. Your voice is unique though, use it.

PR #319, opened 2026-01-14
feat: add pagination details using query to getAllTeamMembers using t…

### LINKED ISSUE #307: [Feature request]:  Pagination (Team Page)
### Description

Pagination for Team Page

### Proposed solution

_No response_

### Alternative solutions (if any)

_No response_

### Additional context

_No response_

bcha92:
## Summary

Added pagination in the backend to getAllTeamMembers (with teamId)

## Tests

No test files created

## Additional information

added pagination in the backend SQL query using optional parameters LIMIT and PAGE to minimize frontend overhead on fetch


bcha92:
@kbventures currently the pagination logic has not been implemented as a reusable function (it is FIXED to the getAllTeamMembers() API call). Will find time to work on creating a util function that will behave in a similar way.

api/src/routes/team-members/data.ts
```diff
@@ -101,7 +115,8 @@ export async function getAllMembers(database: D1Database, teamId: string) {
 			AND role.deletedAt IS NULL
 			AND access.activatedAt IS NOT NULL
 			AND access.deletedAt IS NULL
-	`).bind(teamId).run<TeamMemberInfo>();
+		${limit ? `LIMIT ${limit} OFFSET ${offset}` : ''}
```
madcampos, line 118: **issue:** This leaves room for SQL injection as the limit and offset values will most probably come from somewhere else, potentially untrusted input.

Set both `limit` and `offset` as variables passed on the `bind` method instead of on the `prepare` method. And set defaults for them, so if the value is missing it will fallback to the defaults.

api/src/routes/team-members/index.ts
```diff
@@ -51,22 +51,32 @@ teamMemberRoutes.openapi(
 			return context.json({ message: 'Team not found' } satisfies StatusResponse, StatusCodes.NOT_FOUND);
 		}
 
-		const members = await getAllMembers(context.env.Database, id);
+		// Pagination Setup TODO: Create a utility function for the below?
+		const { limit, page } = context.req.query();
+		const integerOrUndefined = (query?: string): number | undefined => query === '0' || Number(query) ? Math.floor(Number(query)) : undefined;
```
madcampos: **issue:** this should be part of the schema validation, `zod` already provide the `number().optional()` construct, use that.

api/src/routes/team-members/index.ts
```diff
@@ -51,22 +51,32 @@ teamMemberRoutes.openapi(
...
+
+		const count = await countAllMembers(context.env.Database, id);
+		const totalMembersCount = (count[0]?.['count']) as number;
+		const limitCount = integerOrUndefined(limit);
+		const selectedPage = integerOrUndefined(page);
+		const lastPageCount = !limitCount ? 1 : Math.ceil(totalMembersCount / limitCount);
+		const offset = !limitCount ? 0 : ((selectedPage ? selectedPage - 1 : 1) * limitCount);
+
+		const members = await getAllMembers(context.env.Database, id, limitCount, offset);
```
madcampos, line 62: **suggestion:** this may be better represented as a single function call that returns a composed result (i.e. and object) with both results.

Use the `batch` method of the database to create multiple unrelated queries.

Here are examples: https://github.com/search?q=repo%3Atorontojs%2Fcommunityhub%20database.batch&type=code

api/src/routes/team-members/index.ts
```diff
@@ -51,22 +51,32 @@ teamMemberRoutes.openapi(
...
 		return context.json(
-			// TODO: implement proper pagination
 			{
 				data: members,
-				start: 0,
-				end: members.length - 1,
-				total: members.length,
+				start: offset,
+				end: !limitCount || offset + limitCount > totalMembersCount ? totalMembersCount - 1 : offset + limitCount - 1,
+				total: totalMembersCount,
```
madcampos, line 78: **suggestion:** Same as before: return this calculation from the database, I think you can infer it from the results you already heve.

This is the inspiration I used when designing the initial interface, it is a good read to understand better the parameters: https://docs.astro.build/en/reference/routing-reference/#the-pagination-page-prop

api/src/routes/team-members/index.ts
```diff
@@ -51,22 +51,32 @@ teamMemberRoutes.openapi(
...
-				lastPage: 1,
+				currentPage: selectedPage ?? 1,
+				lastPage: lastPageCount,
 				_links: {
-					self: { href: context.req.url },
-					first: { href: context.req.url },
-					last: { href: context.req.url }
+					self: { href: context.req.url.split('?')[0] + (limit ? `?limit=${limit}&page=${selectedPage}` : '') },
+					first: { href: context.req.url.split('?')[0] + (limit ? `?limit=${limit}&page=1` : '') },
+					last: { href: context.req.url.split('?')[0] + (limit ? `?limit=${limit}&page=${lastPageCount}` : '') }
```
madcampos: Use the [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) class.

E.g.:
```typescript
const firstPage = new URL(context.req.url);

firstPage.searchParams.set('limit', limit);
firstPage.searchParams.set('page', '1');
```

Also, the `self` url, is the exact url provided, no changes ot it, so you can just echo it back.

api/src/routes/team-members/data.ts
bcha92: I've addressed this by preparing a default method in the event an unsuitable value is presented, and also added `limit` and `offset` in the bind method.

api/src/routes/team-members/index.ts
bcha92: still looking into this

api/src/routes/team-members/index.ts
bcha92: still looking into this.

api/src/routes/team-members/index.ts
bcha92: Moved to new github issue #324 
closing this

api/src/routes/team-members/index.ts
bcha92: Moved to #324

madcampos (changes requested):
Tests are failing, please double check if they need updating or it is just that the branch deviated from `dev`.

If it is the second one, try rebasing the branch or merging `dev` into it.

bcha92:
reverted PRs to find root cause of CI testing failure

madcampos (dismissed):

kbventures (approved):
Great work!

PR #331, opened 2026-06-06
308 teams page teams page front end implementtion

### LINKED ISSUE #308: [Teams Page] Teams Page Front End Implementtion
### Description

Image

kbventures:
## Summary

Teams Page, Team Page, Profile Page(not current user), Sign Out Button and Gravatar Input/Edit

https://github.com/torontojs/communityhub/issues/308


bcha92 (dismissed):
Looks good but will make minor notes to address once the MVP is released:

- On incorrect email/password entry, sign in page only shows console log object, no alert or visual display showing incorrect login attempt.
- For user sign up, if password does not meet security requirements and user attempted to create account anyway, we should have a modal or "alert" indicating insufficiency with the set password requirement. 
- The pages for "account ready", "set password complete" (first time users), and "reset password success" look almost exactly the same. Maybe we can consider turning the page itself as a reusable component since there is only one link going back to sign in page.

Will mark for approval as these are minor issues.

AD9-1:
image

I am getting this console error, this might not be an obstruction for whole app to run, but this might cause a build error in future

image
I am getting the above error while trying to edit the name/pronouns.

I am with the login_id: root@torontojs.com

api/src/index.ts
```diff
@@ -28,7 +29,27 @@ apiRoutes.onError((err, context) => {
 	return context.json({ message: 'An error has occured' }, StatusCodes.INTERNAL_SERVER_ERROR);
 });
 
-// CORS middleware22
+// Browser hardening only; these headers are not required for API functionality.
+apiRoutes.use(
```
Saif-XI-Coderz: I'm thinking that since this is already a major PR, would it be better off focusing on teams features in this one and delegate to smaller atomic PRs for other concerns? There's a chance of the code review being overwhelmed.

api/src/utils/validation.ts
```diff
@@ -6,3 +6,8 @@ export const IdParamSchema = z.object({
 });
 
 export type IdParam = z.infer<typeof IdParamSchema>;
+
+export const PaginationQuerySchema = z.object({
+	limit: z.coerce.number().int().positive().optional(),
```
Saif-XI-Coderz: I see the 'page' getting a default but 'limit' does not have one. Is it intentional or would we want a default for limit too?

src/components/TeamDetail/TeamDetail.tsx
```diff
@@ -0,0 +1,607 @@
...
+import AuthenticatedLayout from '../AuthenticatedLayout/AuthenticatedLayout.tsx';
+import Button from '../Button/Button.tsx';
+import EmptyIcon from '../EmptyIcon/EmptyIcon.tsx';
+import TrashIcon from '../Icons/Utilities/TrashIcon.tsx';
+
+interface Props {
+	teamId: string;
+}
+
+interface Team {
```
AD9-1, line 17: Question: Is there any global file to store all the types or interfaces?

api/src/index.ts
Saif-XI-Coderz: We talked about in the weekly contribution call. Consider this resolved.

api/src/routes/team-members/data.ts
```diff
@@ -115,10 +141,14 @@ export async function getAllMembers(database: D1Database, teamId: string, limit?
 			AND role.deletedAt IS NULL
 			AND access.activatedAt IS NOT NULL
 			AND access.deletedAt IS NULL
-		${limit ? `LIMIT ${limit} OFFSET ${offset}` : ''}
-		`).bind(teamId).run<TeamMemberInfo>();
+		${limit ? 'LIMIT ? OFFSET ?' : ''}
+		`).bind(teamId, ...(limit ? [limit, offset] : [])).run<TeamMemberInfo>();
```
Saif-XI-Coderz: Parameterized bindings replacing string interpolation. Nice!

api/src/routes/team-members/data.ts
```diff
@@ -62,25 +64,37 @@ export async function addTeamMembers(database: D1Database, teamId: string, data:
...
+				keys.push('description = ?');
+				values.push(description);
+			}
+
+			return database.prepare(`
 				UPDATE ${DBTables.ROLE}
 				SET
-					name = ?,
-					description = ?
+					${keys.join(', ')}
```
Saif-XI-Coderz, line 84: Question: Can keys be empty here? If it is empty, what happens?

api/src/routes/profile/validation.ts
```diff
@@ -60,6 +61,13 @@ export const ProfileTeamSchema = z.object({
 
 export type ProfileTeam = z.infer<typeof ProfileTeamSchema>;
 
+const AvatarSchema = z
+	.url()
+	.trim()
+	.min(1, 'Avatar must be at least one character long.')
+	.max(SHORT_TEXT_SIZE_IN_CHAR, `Avatar must be at most ${SHORT_TEXT_SIZE_IN_CHAR} characters long.`)
+	.refine((url) => isGravatarAvatarUrl(url) || isGravatarProfileUrl(url), { message: 'Avatar must be a Gravatar URL.' });
```
Saif-XI-Coderz, line 69: When we tested together, we noticed that the profile page is not saving updates and throwing some errors. This must be where that error is originating from based on matching error message. Could be worthwhile to dig here when fixing that issue.

api/src/routes/profile/data.ts
```diff
@@ -131,14 +134,10 @@ export async function updateProfileById(
...
-			FROM (
-				SELECT id, activatedAt, deletedAt
-				FROM ${DBTables.ACCESS}
-				WHERE
-					access.id = id
-			) AS access
+			FROM ${DBTables.ACCESS} AS access
 			WHERE
 				profile.id = ?
+				AND access.id = profile.id
```
Saif-XI-Coderz, line 140: I like this. Good catch.

src/components/TeamDetail/TeamDetail.tsx
kbventures: We do  not have a global types(shared resuable types across the entire code base and we have no intention on having one) .  We can utilize shared types for respective front or back end when needed.

api/src/utils/gravatar.ts
```diff
@@ -0,0 +1,59 @@
...
+	try {
+		const response = await fetch(`https://api.gravatar.com/v3/profiles/${encodeURIComponent(slug.replace(/\.card$/iu, ''))}`);
+		if (!response.ok) { return null; }
+
+		const profile: { avatar_url?: string } = await response.json();
+		return profile.avatar_url && isGravatarAvatarUrl(profile.avatar_url) ? profile.avatar_url : null;
+	} catch {
+		return null;
+	}
+}
```
Saif-XI-Coderz, line 102: Suggestion: These suite of functions could benefit from a few ts docs. They are related and similar, and requires a little work to understand how they are different and what do they do.

src/context/AuthContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { safeAvatarUrl } from '../utils/safeAvatarUrl.ts';
+
+interface AuthState {
+	accessLevel: AccessLevel | null;
+	avatar: string;
+	isLoading: boolean;
+	profileStatus: ProfileStatus | null;
+}
+
+const DEFAULT_AVATAR = '/default-avatar.png';
```
Saif-XI-Coderz: Suggestion: We could set the default avatar directly at the db level for each user, ensuring we never start null. I'm guessing this could be via migration/seeding? This then becomes a data constraint. It would help us avoid validation gymnastics in the code. I believe this will make other parts of the code that deals with gravatars simpler as well.

api/src/utils/gravatar.ts
```diff
@@ -0,0 +1,59 @@
...
+			!extraPath
+		);
+	} catch {
+		return false;
+	}
+}
+
+export async function resolveGravatarAvatarUrl(url: string): Promise<string | null> {
+	if (isGravatarAvatarUrl(url)) { return url; }
+	if (!isGravatarProfileUrl(url)) { return url; }
```
Saif-XI-Coderz: Question: Is the negation a typo? If not, I'm reading these two lines like this: "if it's not an avatar url and it's not an profile url, resolve it as is". Maybe it works, but somehow it does not read right.

api/src/utils/validation.test.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, expect, it } from 'vitest';
+import { IdParamSchema, PaginationQuerySchema } from './validation.ts';
+
+const VALID_UUID = '3c5123c0-8548-4a02-a83c-32e9ce67eae8';
+
+describe('PaginationQuerySchema', () => {
```
Saif-XI-Coderz, line 6: Comprehensive suite of test (Y).

src/components/AuthenticatedLayout/AuthenticatedLayout.tsx
```diff
@@ -0,0 +1,130 @@
...
+	},
+	{
+		id: 'profile',
+		href: '/pages/protected-profile/',
+		icon: '/person-icon-white.png',
+		label: 'My Profile'
+	},
+	{
+		id: 'teams',
+		href: '/pages/team',
```
Saif-XI-Coderz: Currently landing on a "Team not found" page with the current url.

```suggestion
		href: '/pages/teams',
```

api/src/utils/responses.ts
```diff
@@ -335,3 +334,32 @@ export function generatePaginatedResponseSchema<T extends unknown[]>(data: ZodTy
...
+	currentPage.searchParams.set('limit', limit?.toString() ?? '');
+	currentPage.searchParams.set('page', page.toString());
+	const lastPage = new URL(url);
+	lastPage.searchParams.set('limit', limit?.toString() ?? '');
+	lastPage.searchParams.set('page', lastPageCount.toString());
+
+	return {
+		offset,
+		start: offset,
+		end: !limit || offset + limit > total ? total - 1 : offset + limit - 1,
```
Saif-XI-Coderz: Question: It seems like against an empty table (total=0), end will be -1. If ever this happens, will this break things?

src/components/Teams/Teams.tsx
```diff
@@ -1,152 +1,309 @@
...
-				setTeamMembersData(await responseTeamMemebers.json());
-			} catch (error) {
-				setTeamMembersDataError(error);
-				console.error('Error fetching team members data:', error);
-			} finally {
-				setIsLoadedTeamMembersData(true);
+		void fetchTeamsData({ page: teamsCurrentPage, pageSize: teamsPageSize });
+	}, [fetchTeamsData, teamsCurrentPage, teamsPageSize]);
+
+	const handleAddTeamSubmit = async (event: React.FormEvent<HTMLFormElement>): Promise<boolean> => {
```
AD9-1, line 74: Suggestion: handleAddTeamSubmit() and handleEditTeamSubmit() have nearly identical logic, differing only in HTTP method (POST vs PATCH) and URL. We could refactor them into a single handleAddEditTeam function and use a ternary/conditional to determine the request type and endpoint based on whether a team is being edited.

api/src/routes/auth/index.ts
```diff
@@ -384,17 +397,13 @@ authRoutes.openapi(
 	async (context) => {
 		const { token } = context.req.valid('json');
 
-		if (!token) {
-			return context.json({ message: 'Invalid or missing token' } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
-		}
-
 		const success = await context.env.PasswordResetToken.get(token);
```
manvinderjit, line 400: **bug**: The reset password feature is broken because of the following key change from ```resetToken - email``` 

to 
```email - resetToken ``` 

It was brought about by the following code on line **297**

```
await context.env.PasswordResetToken.put(email, resetToken, { expirationTtl: TEN_MINUTES_IN_SECONDS });
```


Image

The password reset token fetched from database will always be empty because get(token) still uses the token as the key, but the key is now email, so it will never find anything.

The fetch should use **email** as the key and not **token**.

File: src/components/AuthenticatedLayout/AuthenticatedLayout.tsx (hunk 12 above)
Comment (manvinderjit) [inline src/components/AuthenticatedLayout/AuthenticatedLayout.tsx]: **typo:** incorrect URL for the teams page.
```suggestion
		href: '/pages/teams',
```

src/context/AuthContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+	avatar: string;
+	isLoading: boolean;
+	profileStatus: ProfileStatus | null;
+}
+
+const AuthContext = createContext<AuthState>({
+	accessLevel: null,
+	avatar: DEFAULT_AVATAR,
+	isLoading: true,
+	profileStatus: null
```
manvinderjit, line 19: **suggestion:** instead of using ```profileStatus: null```, it might be better to have a dedicated ```isAuthenticated``` boolean value with status as true or false.

**related to:** https://github.com/torontojs/communityhub/pull/331/changes#r3495654530

src/hooks/useHeartBeat.ts
```diff
@@ -1,178 +1,72 @@
...
-
-// Protected page check for volunteers, organizers and admin
-export const getHeartBeatProtectedHomePage = async (): Promise<boolean> => {
-	try {
-		const response = await fetch('/api/auth/heartbeat', {
-			method: 'GET',
-			credentials: 'include'
-		});
+		if (isLoading) { return; }
+		if (profileStatus !== null) { window.location.href = '/pages/home'; }
```
manvinderjit: **question**: the check ```if (profileStatus !== null) { window.location.href = '/pages/home'; }``` relies on the assumption that the api will always return a status. 

What if the api doesn't return a status or returns a falsy status and the user is still authenticated?

Might be better to add an ```isAuthenticated``` boolean as indicated in the following comment

**related to**:
https://github.com/torontojs/communityhub/pull/331/changes#r3495651641

api/src/routes/team-members/data.ts
manvinderjit: +1. This will throw an error when neither of the fields are defined. 

Add a check in zod validation schema to ensure the query runs only if there is at least one field provided for modification.

api/src/utils/validation.ts
manvinderjit: +1: This should limit the maximum number of fetched rows.

src/components/AuthenticatedLayout/AuthenticatedLayout.css
```diff
@@ -0,0 +1,299 @@
...
+.authenticated-layout .inner-header .small-avatar,
+.authenticated-layout .inner-header .notification-bell {
+	height: 2rem;
+	width: 2rem;
+}
+
+.authenticated-layout .sidebar-left {
+	border-right: var(--border);
+	box-sizing: border-box;
+	height: 41.25rem;
```
manvinderjit: **css:** The vertical side bar border doesn't extend to the full page length when the page is bigger than viewport. 

Image

manvinderjit (commented):
@kbventures Great work and thank you for all your effort.

I have done a quick overall review of the changes and run the project locally. I have only pointed major blockers.

As the code uses AI, there is potential for refactoring to improve usability but considering we need to got to MVP ASAP, I am ignoring those for now.

src/components/TeamMemberCard/TeamMemberCard.tsx
```diff
@@ -1,29 +1,29 @@
 import './TeamMemberCard.css';
-import type { TeamMemberProfile } from '../Teams/Teams.tsx';
+import type { TeamMemberProfile } from '../../types/index.ts';
+import { getInitials } from '../../utils/getInitials.ts';
 
 const TeamMemberCard = (props: TeamMemberProfile) => (
 	<article className='team-member-card'>
 		<a
```
AD9-1, line 8: Suggestion: If this is an internal route, consider using the project’s routing component (e.g. Link) instead of a plain <a> to avoid full page reloads and improve SPA navigation experience.

src/components/AuthenticatedLayout/AuthenticatedLayout.tsx
Saif-XI-Coderz: Resolving. Same suggestion covered by @manvinderjit.

PR #332, opened 2026-06-23
298 fix: minor dependency updates to address security vulnerabilities

bcha92:
## Summary
Minor dependency updates to address immediate security vulnerabilities in the VMS project

## Tests

None; The only check made is to ensure peer dependency and version update does not conflict with the current code base

## Additional information

None


kbventures (approved):
Great work @bcha92

PR #343, opened 2026-07-28
Created the open_role and open_role_members in 0008_create-open-role.sql

AD9-1:
## Summary

This PR introduces the database schema required to support the **Open Roles** feature in the TorontoJS Community application.

#### Added `open_role` table

The `open_role` table stores public opportunities posted by a team.

Fields:
- `id`
- `open_role_name`
- `open_role_description`
- `isBasedOnGTA`
- `experience_level`
- `tech_stacks`
- `position_required`
- `team_id` (FK → `team.id`)
- `insertedAt`
- `deletedAt`

#### Added `open_role_members` table

The `open_role_members` table stores members who have applied for an open role.

Fields:
- `id`
- `open_role_id` (FK → `open_role.id`)
- `profile_id` (FK → `profile.id`)
- `open_role_name` (FK → `open_role.open_role_name`)
- `insertedAt`
- `deletedAt`

### Database Relationships

```
team
  │
  │ 1
  ▼
open_role
  │
  │ 1
  ▼
open_role_members
  ▲
  │
profile
```
This schema establishes the foundation for supporting public community opportunities and member applications.

## Tests

---I have not done any tests

---

## Additional Information
---No additional info

### DIFF
--- a/.vite-dev.log
+++ b/.vite-dev.log
@@ -1,0 +1,44 @@
+
+> vms@0.0.1 dev
+> vite dev
+
+✓ Added main entry: O:\codes\communityhub\src\index.html
+✓ Added page: account-ready
+✓ Added page: button-usage
+✓ Added page: check-steps
+✓ Added page: check-your-email
+✓ Added page: complete-profile
+✓ Added page: confirm-account
+✓ Added page: date-usage
+✓ Added page: dev-links
+✓ Added page: forgot-password
+✓ Added page: home
+✓ Added page: notification-box-usage
+✓ Added page: notifications
+✓ Added page: print-documents
+✓ Added page: profile
+✓ Added page: profiles
+✓ Added page: protected-page-admins
+✓ Added page: protected-page-all
+✓ Added page: protected-page-organizers
+✓ Added page: protected-profile
+✓ Added page: reset-password
+✓ Added page: reset-password-success
+✓ Added page: review-conduct-code
+✓ Added page: sign-in
+✓ Added page: sign-up
+✓ Added page: team
+✓ Added page: teams
+✓ Added page: text-input-usage
+✓ Added page: usage-helper-message
+✓ Added page: volunteer
+Using vars defined in .dev.vars
+
+  VITE v7.0.5  ready in 1539 ms
+
+  ➜  Local:   http://localhost:3000/
+  ➜  Network: http://192.168.26.144:3000/
+  ➜  Debug:   http://localhost:3000/__debug
+  ➜  press h + enter to show help
+6:26:35 pm [vite] Pre-transform error: Failed to load url /sign-up.tsx (resolved id: /sign-up.tsx). Does the file exist?
+6:26:48 pm [vite] Pre-transform error: Failed to load url /sign-in.tsx (resolved id: /sign-in.tsx). Does the file exist?
--- a/db/migrations/0008_create-open-role.sql
+++ b/db/migrations/0008_create-open-role.sql
@@ -8,28 +8,28 @@ CREATE TABLE IF NOT EXISTS open_role(
 	-- The UUID, stored as text
 	id TEXT NOT NULL UNIQUE COLLATE BINARY,
 	-- The name of the open role
-	open_role_name TEXT NOT NULL,
+	openRoleName TEXT NOT NULL,
 	-- A text blurb the person can provide about themselves
-	open_role_description TEXT,
+	openRoleDescription TEXT,
 	-- A flag indicating if the user is based on the Grater Toronto Area (GTA)
 	-- This and the following flags are a proxy for information if people can attend online and in person events.
 	-- It is enough to give us information if the person is around Toronto without needing to ask the actual location.
 	isBasedOnGTA INTEGER NOT NULL DEFAULT 1 CHECK(isBasedOnGTA IN (0, 1)),
 	-- The experience level of the open role (senior, mid, entry)
-	experience_level TEXT NOT NULL CHECK(experience_level IN ('senior', 'mid', 'entry')),
+	experienceLevel TEXT NOT NULL CHECK(experienceLevel IN ('senior', 'mid', 'entry')),
 	-- The tech stacks required for the open role
-	tech_stacks TEXT NOT NULL,
+	techStacks TEXT NOT NULL,
 	-- The position of the open role (1,2,5...)
-	position_required INTEGER NOT NULL,
+	positionRequired INTEGER NOT NULL,
 	-- The team id of the open role (FK of Team(id))
-	team_id TEXT NOT NULL,
+	teamId TEXT NOT NULL,
 	-- The date this open role was added to the database, saved as an ISO timestamp
 	insertedAt DATETIME NOT NULL,
 	-- The date this open role was deleted from the database, saved as an ISO timestamp
 	deletedAt DATETIME DEFAULT NULL,
 
 	PRIMARY KEY (id),
-	FOREIGN KEY (team_id) REFERENCES team(id)
+	FOREIGN KEY (teamId) REFERENCES team(id)
 );
 
 Drop TABLE IF EXISTS open_role_members;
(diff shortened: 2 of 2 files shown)


db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
+-- Migration number: 0002 	 2025-01-30T00:58:21.685Z
+
+DROP TABLE IF EXISTS open-role;
```
bcha92: API and unit tests are failing because the hyphen (`-`) in `open-role` is a special syntax in SQL and will not be recognized as part of SQL's nomenclature (naming convention). Suggesting alternative names here to bypass this:
1. `open_role` <-- use underscore (`_`) for the simplest fix
2. `"open-role"` <-- use double quotes if the hyphen needs to be intact another reason.

Same goes for `open-role` in Lines 7, 50, 52, and comment in 39 on this sql file.

bcha92 (changes requested):

db/migrations/0008_create-open-role.sql
AD9-1: did

bcha92 (dismissed):

.vite-dev.log
kbventures, line 1: Why are we introduce vite log in this PR?

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
...
+-- A person's profile inside the database
+-- It should contain no sensitive information
+CREATE TABLE IF NOT EXISTS open_role(
+	-- The UUID, stored as text
+	id TEXT NOT NULL UNIQUE COLLATE BINARY,
+	-- The name of the open role
+	openRoleName TEXT NOT NULL,
+	-- A text blurb the person can provide about themselves
+	openRoleDescription TEXT,
+	-- A flag indicating if the user is based on the Grater Toronto Area (GTA)
```
kbventures, line 14: "Grater" :)

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
...
+	-- This and the following flags are a proxy for information if people can attend online and in person events.
+	-- It is enough to give us information if the person is around Toronto without needing to ask the actual location.
+	isBasedOnGTA INTEGER NOT NULL DEFAULT 1 CHECK(isBasedOnGTA IN (0, 1)),
+	-- The experience level of the open role (senior, mid, entry)
+	experienceLevel TEXT NOT NULL CHECK(experienceLevel IN ('senior', 'mid', 'entry')),
```
kbventures, line 19: Why are we introduce a new field which isn't found in profile currently?  We do not currently have a "experienceLevel" found in the create profile schema.  We  are utilzing the description for that purpose.   Opting for simplicity and the option of enhancing specificity later.

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
...
+	-- The tech stacks required for the open role
+	techStacks TEXT NOT NULL,
```
kbventures, line 21: This comment is along the lines of the previous one.

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
...
+	-- The position of the open role (1,2,5...)
+	positionRequired INTEGER NOT NULL,
```
kbventures, line 23: teamId should be a FK and positionRequired is therefore redundant?

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
...
+	-- The date this open role was added to the database, saved as an ISO timestamp
+	insertedAt DATETIME NOT NULL,
+	-- The date this open role was deleted from the database, saved as an ISO timestamp
+	deletedAt DATETIME DEFAULT NULL,
+
+	PRIMARY KEY (id),
+	FOREIGN KEY (teamId) REFERENCES team(id)
+);
+
+Drop TABLE IF EXISTS open_role_members;
```
kbventures, line 35: I'm not sure if this NEEDS to exists.  Can you share the rational why we need a second table for the open roles?

kbventures (changes requested):

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
+-- Migration number: 0002 	 2025-01-30T00:58:21.685Z
+
+DROP TABLE IF EXISTS open_role;
+
+-- A person's profile inside the database
+-- It should contain no sensitive information
+CREATE TABLE IF NOT EXISTS open_role(
```
Saif-XI-Coderz, line 7: Question (probably for @manvinderjit or @bcha92 ). If we just deleted the table above, we know the table does not exist. In this case, does the "if not exists" qualifier have a meaningful effect here? I'm not too familiar with the environment and its state, where the migrations are run, it might have an impact that I'm not getting.

Another question. When running migration, do we always want to delete the table (and its data)? Or is the "if not exists" sufficient to keep the migrations working when a table already exists?

@AD9-1 is just following previous migration convention here. We are asking if the convention so far is the right one.

AD9-1: This migration is intended to be executed locally during development, rather than against an existing production database. The DROP TABLE IF EXISTS is intentional so that we can reset the open_role table and recreate it with the current schema when rerunning the migration.
For this local setup, we do want to delete and recreate the table when the migration is rerun so that the local database starts with a clean schema.

db/migrations/0008_create-open-role.sql
AD9-1: If that is the concern and refers with the profile schema, then it can be romoved

bcha92:
merging blocked pending approval from design team (Tales)

bcha92:
Remove the entire `open_role_members` table, it's not needed.

In the `open_role` schema:
1. remove `team_id`, `position_required` (because the simple existence of the open role already indicates that).
2. change `open_role_name` to `title`, and `open_role_description` to `description`.
3. add `user_id` (the person who created the role), `modified_at` (allows editing)
4. change these properties to `camel_case` format: `isBasedOnGTA`, `insertedAt`, and `deletedAt`

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
...
+	-- The UUID, stored as text
+	id TEXT NOT NULL UNIQUE COLLATE BINARY,
+	-- The name of the open role
+	openRoleName TEXT NOT NULL,
```
manvinderjit, line 11:
```suggestion:
name TEXT NOT NULL,
```

suggestion: please change the field name to ```name``` instead of ```openRoleName``` to align with the other sql migrations.

When we access a sql field, we use table_name.field_name, which in this case will be ```open_role.openRoleName``` which is redundant. Having it as ```name``` will allow it to be referred as ```open_role.name```

db/migrations/0008_create-open-role.sql
```diff
@@ -0,0 +1,53 @@
...
+	-- A text blurb the person can provide about themselves
+	openRoleDescription TEXT,
```
manvinderjit, line 13:
```suggestion:
description TEXT NOT NULL,
```

Same as https://github.com/torontojs/communityhub/pull/343/changes#r3962956334

```open_role.description``` is better than ```open_role.openRoleDescription```

db/migrations/0008_create-open-role.sql
manvinderjit: I agreed with what Ken said. Additionally, I don't think experience level will hinder if an individual is accepted into a volunteer role.

db/migrations/0008_create-open-role.sql
manvinderjit: Not all roles require tech stack. We have non technical roles as well. Also, I don't think we will be using this as a filter - @bcha92 @Saif-XI-Coderz

db/migrations/0008_create-open-role.sql
manvinderjit: I think we don't need to track users who applied for a role as long as the admin has the capability to assign a user to an open role, making the mechanism simpler.

db/migrations/0008_create-open-role.sql
manvinderjit: @Saif-XI-Coderz Migrations files are executed when they are first pushed. As long as an existing migration file is not edited, it will not be re-executed, making this statement safe for production as it is executed only the first time. 

But this depends on the condition that we never modify a migration file again, which is why any Schema changes must be in a new migration file. For reference, check this migraton file [https://github.com/torontojs/communityhub/blob/dev/db/migrations/0006_move-profile-fields-to-access.sql](url)

.vite-dev.log
manvinderjit: @AD9-1 Please remove this file. Thank you.

manvinderjit (commented):
Great work @AD9-1 . Recommended a few changes and suggestions. Keep it up.

manvinderjit:
@AD9-1 As mentioned in the documentation, please use conventional commits for commit messages:

**Link:** https://www.conventionalcommits.org/en/v1.0.0/

Here is a github cheatsheet as well. No need to use the tool, just refer to the cheatsheet: 
https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13

db/migrations/0008_create-open-role.sql
Saif-XI-Coderz: Got it. In that case can we simplify `CREATE TABLE IF NOT EXISTS` into `CREATE TABLE`?

.vite-dev.log
bcha92: add this file to the `.gitignore` file in the repository under `# Logs`

bcha92 (changes requested):
just a few notes on improvement then it should be good.

PR #349, opened 2026-08-04
fix(profile): allow editing general info without a gravatar avatar

### LINKED ISSUE #348: [BUG]: Cannot save changes in "Edit General Information" modal
### Description

When saving changes in the Edit General Information modal (name, pronouns, location flags), the save silently fails for users who haven't set a Gravatar avatar. The modal stays open with no error message.

### Steps to Reproduce

1. Go to your profile page
2. Click the edit pencil on your header section
3. Change your name (or any field)
4. Click Save

### Expected Behavior

The changes should be saved and the modal should close.

### Actual Behavior

The modal stays open and nothing is saved. No error message is shown.

### Screenshots (if applicable)

Image

### Operating System

Debian

### Browser

Brave

### Browser Version

latest

### Additional Information

_No response_

Saif-XI-Coderz:
## Summary

Fixes #348 — a bug where users without a Gravatar avatar could not save any general profile info edits (name, pronouns, GTA/local-event flags).

The `GeneralInfoFormModal` submits the avatar via a hidden input. For users without a Gravatar, the frontend substitutes the local default avatar (`/default-avatar.png`). That path was sent to `PATCH /api/profiles/{id}`, where `UpdateProfileSchema` / `AvatarSchema` requires an absolute Gravatar URL, so the request was rejected with 422 and the save silently failed.

The form now sends `null` for the avatar when it is empty or a local path, which the schema already accepts (`avatar: AvatarSchema.nullable().optional()`) and the DB column already supports.

## Tests

- Manually verify: editing general info as a user without a Gravatar avatar now succeeds

### DIFF
--- a/src/components/ProtectedProfile/ProtectedProfile.tsx
+++ b/src/components/ProtectedProfile/ProtectedProfile.tsx
@@ -207,10 +207,11 @@ export const ProtectedProfile = (): React.JSX.Element => {
 	const handleGeneralInfoSubmit = async (event: React.FormEvent<HTMLFormElement>): Promise<boolean> => {
 		event.preventDefault();
 		const formData = new FormData(event.currentTarget);
+		const rawAvatar = (formData.get('avatar') as string)?.trim();
 
 		const updatedData = {
 			name: (formData.get('name') as string)?.trim(),
-			avatar: (formData.get('avatar') as string)?.trim() || null,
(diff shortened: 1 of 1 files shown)


bcha92 (approved):

kbventures (approved):
Great work. Thanks. Approved.

